PR #2, opened 2021-04-20
Update to the latest abi-next branch.

sunfishcode:
This involved some straight-forward API updates, but also involved the transition away from `Document`.

The approach I ended up using here was to to have the rust-wasm backend put each witx module in its own Rust `mod` so that we can generate multiple modules in the same bindings.rs file. That currently means that users have to do things like `use wasm::*` though. I'm open to other ideas here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1142,6 +1142,7 @@ name = "test-wasmtime"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "bitflags",
  "wasi-cap-std-sync",
  "wasmtime",
  "wasmtime-wasi",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ version = "0.13.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "458e98ed00e4276d0ac60da888d80957a177dfa7efa8dbb3be59f1e2b0e02ae5"
 dependencies = [
- "rand 0.8.3",
+ "rand",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


alexcrichton:
egad for some reason I had notifications turned off on this repo so I only saw this now, sorry about that!

sunfishcode:
No worries :-). This PR ended up heading off into some experimental directions, and I should probably split things out to make it easier to review anyway.

alexcrichton:
Nah it's fine to keep all here, I just looked over it all and it all seems fine by me. Some questions though:

* Could you explain a bit more about the rationale for `HandleIndex`? (on both the host and guest side)
* Do you want to merge your WASI changes into my branch? (to update the WASI PR itself)

sunfishcode:
Ok, I opened https://github.com/alexcrichton/WASI/pull/1 to request merging my WASI changes into your branch. It's mostly just witx (s-expr) parsing changes, though there's also https://github.com/alexcrichton/WASI/pull/1/commits/9013e18b5d33e442387ca9b103833b00f750198b which makes it a little more strict about requiring the parsed module name to match the filename.

`HandleIndex` is one of the experimental parts :-). I was trying to export a function with an argument type that's a handle to an imported resource, and I didn't fully understand how tables worked at first, so it was convenient for me to have the indices exposed so that I could hook things up manually in my host code while figuring it out.

alexcrichton:
Ah ok makes sense! Want to separate that out for now and then I can merge the rest? I think if you rebase `Cargo.lock` should now be tracked so it can have an update of the git rev too

sunfishcode:
Ok, I removed the `HandleIndex` parts, and then added a few patches to fix some of the tests.

Several tests still fail, but they look like they just need the same kinds of changes as the "integers" test. I'm not sure if we should make those changes, or push more of the boilerplate into the generated code.

alexcrichton:
Seems reasonable to me! At least enough things are still running on CI. It's fine to clean things up later if necessary.

PR #5, opened 2021-05-20
Get `witx-bindgen-gen-{core,rust,wasmtime}` and `witx-bindgen` compiling again

fitzgen:
This doesn't get the test crates compiling or running again, as I think that requires at minimum some updates to the test witx files.


alexcrichton:
👍

PR #6, opened 2021-05-24
Get tests running again

fitzgen:
With these changes the tests will run, but fail right now. In the test runner we
do

    let doc = witx::parse(source).unwrap();

and this results in a module whose name is `"-"` which we eventually camel case
and get `""`, and then including this in our emitted Rust source results in
broken Rust code, which causes our tests to fail.

The module name definitely should not be "-" but it is not clear to me yet what
the name *should* be or which bit of code is to blame for this incorrect name.


alexcrichton:
👍

PR #11, opened 2021-06-04
Sync with wasmtime's new API, Implement new syntax

alexcrichton:
This is a big change. High level goals include:

* Update to the latest `wasmtime` API
  * This invalidates the previous design of `GuestPtr`.
  * The `GuestPtr` type is now dropped, and all host functions get raw borrows instead
  * Adjust to API changes literally, such as passing more context around
  * Adjust to ownership of store state, using trait bounds and tweaking where things like host tables for objects live
  * Removes the need for a borrow checker to be on the host data, it's now per-call
* Implement a new syntax
  * A new `witx2` crate is added with a custom recursive-descent parser to parse a new syntax
  * With the `old-witx-compat` feature, this transparently translates from old s-expressions syntax to the new syntax
  * Does not support  features of the old syntax that required `@witx`
  * The ABI logic has now all moved to `witx2` and duplicate my PR on the WASI repo for now. This may make more sense to live outside? Unclear where this wants to live but in any case we want it operating over the new AST, not the old.
  * `resource` annotations can now have methods, although this is not supported in a sugar-y style by the bindings generators just yet.
* Update to the latest canonical ABI draft
  * return area space of wasm-exported functions is now allocated in wasm, removing the need for wasmtime to malloc/free on each call
  * Bitflags are different. In the draft they're always i32 arrays, but I've reinterpreted that as they're a "native integer" unless there's more than 64 flags, then it's an i32 array. Code generators don't support the i32 array case just yet. This is different from the draft and needs to be updated upstream (which I think should happen)
  * Naming of malloc/free changed
  * Malloc is no longer exported, only realloc is
  * Destructors for handles are unconditional, they're exposed through the `canonical_abi` module namespace now.

Some capabilities have been dropped for now while work is done to figure them out. One is wasm-exported handles because I need to figure out how best to handle running destructors on the host. Another is pointer-indirected-things in buffers. I'm not sure that's compatible with how we want wasmtime's API traits to get generated, so we may never support this in wasmtime. I'm not sure if it's worth supporting in general? Unsure if we should be modifing the canonical ABI upstream as well to reflect this. 

There's probably more changes here too, I'll update as I think of them. This should be in a decent state for building on I think though. I want to add quite a few more tests, though, this all feels relatively under-tested still.


alexcrichton:
Ok I'm gonna go ahead and merge this and we can continue to iterate in-tree.

PR #13, opened 2021-06-18
Explode `CallMode` into `Direction` and `LiftLower`

fitzgen:
Often code only cares about whether it is defining an ABI for an import or
export, and doesn't care about whether arguments are lifted and results are
lowered or if arguments are lowered and results are lifted. Often code only
cares about the lifting and lowering, and not whether we are dealing with an
import or an export. These two degrees of freedom were tied together in a single
`CallMode` enum previously. Now they are exploded into two enums:

1. `Direction` for discriminating between import vs export, and
2. `LiftLower` for determining whether we lift arguments and lower results or
   lower arguments and then lift results in this glue code.

Additionally, I hope that the lift/lower nomenclature is more clear than the old
wasm vs native nomenclature, which I personally found quite confusing.

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -1,7 +1,9 @@
 use heck::*;
 use std::collections::{BTreeSet, HashMap};
 use std::mem;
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):
👍 I agree this reads a whole lot better, thanks!

PR #14, opened 2021-06-18
Fix endianness and alignment of guest values

### LINKED ISSUE #1: `GuestPtr` and endianity+alignedness checking
`GuestPtr`'s constructor should either check+assert alignedness of the data or it should be added to its `unsafe` contract. We shouldn't be checking/asserting alignedness when grabbing the underlying slice of Wasm memory that it represents.

Endianity is an interesting one: if the host is big endian, then none of the underlying slice's elements will have the same representation as the host (other than guest pointers of `u8` slices) since Wasm is always little endian. We should probably remove the ability to get a raw slice from the `GuestPtr` API so that we don't have to have methods that are only conditionally available depending on the host's target architecture. Instead we should have an iterator/random access/cell style API that allows you to get/set each element individually.

cc @alexcrichton

alexcrichton:
This commit fixes the implementation of Wasmtime imported functions to
stop assuming higher alignment requirements for guest slices.
Additionally it also fixes the wasmtime import functions on big-endian
hosts by automatically converting little-endian values to host-endian values.

I've added tests for alignment but I'm not sure how to add tests for
endianness here. I think we need a big-endian host, but for now let's
just assume it's all correct...

Closes #1

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -369,11 +369,17 @@ impl Generator for Js {
                 let field = field.name.to_shouty_snake_case();
                 self.src.js(&format!(
                     "export const {}_{} = {}{};\n",
-                    name, field, i, suffix,
+                    name,
+                    field,
+                    1u64 << i,
+                    suffix,
                 ));
                 self.src.ts(&format!(
                     "export const {}_{} = {}{};\n",
-                    name, field, i, suffix,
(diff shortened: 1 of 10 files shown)


alexcrichton:
Er oops I didn't mean to insta-merge this...

Well in any case if anyone's got more thoughts on this please let me know!

PR #24, opened 2021-06-27
Assign explicit values to TS enums

Asraelite:
Generated TypeScript enum definitions are currently difficult to use because they lack explicitly assigned values.

For example, the following TypeScript code currently does not compile without this change:

#### generated .d.ts declaration
```ts
export enum Foo {
  Bar,
  Baz,
}
```
#### .ts script
```ts
interface MyInterface {
  tag: Foo.Bar, // error: Enum type 'Foo' has members with initializers that are not literals.
  value: number,
}
```

This change would add values to the enum definitions:

```ts
export enum Foo {
  Bar = 0,
  Baz = 1,
}
```

I am not well-acquainted with TypeScript so maybe there's a better way to fix this.

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -347,10 +347,10 @@ impl Generator for Js {
         } else if variant.is_enum() {
             self.src
                 .ts(&format!("export enum {} {{\n", name.to_camel_case()));
-            for case in variant.cases.iter() {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Would it be possible to extend the `run.ts` with code that fails to compile before but succeeds afterwards?

Asraelite:
Sure. How's that?

alexcrichton:
Perfect, thanks!

PR #26, opened 2021-06-30
Merge wasmlink into witx-bindgen.

peterhuene:


peterhuene:
It looks like witx-bindgen isn't working on Windows, but I haven't looked into it yet.

I'm also getting other wasmlink test failures that I'm investigating.

alexcrichton (approved):
Looks good to me! I think there's some CI issues though?

I'll also add you to the repo so you can merge too

peterhuene:
I figured out the issue: the proc-macros that are doing a `include_str!` aren't accounting for the path separator difference on Windows, hence the "unknown escape character" issue.

Fix is to use a raw-string literal for the proc-macro output.  I'll push up a fix.

peterhuene:
Additional failures in es-lint test that are probably the same thing, fixing that now.

alexcrichton:
Oh oops I didn't realize these were failures in preexisting code just now cropping up on Windows since I didn't test there before.

Feel free to comment out windows for now and I can get to fixing these later too

PR #29, opened 2021-06-30
Fix some Windows CI issues

alexcrichton:
... or at least attempt to!

### DIFF
--- a/crates/witx2/tests/all.rs
+++ b/crates/witx2/tests/all.rs
@@ -141,6 +141,7 @@ impl Runner<'_> {
                 &test.display().to_string(),
                 &test.display().to_string().replace("\\", "/"),
             )
+            .replace("\\parse-fail\\", "/parse-fail/")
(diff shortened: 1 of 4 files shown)


crates/witx2/tests/all.rs
```diff
@@ -100,19 +100,7 @@ impl Runner<'_> {
         let result = if contents.contains("// parse-fail") {
             match result {
                 Ok(_) => bail!("expected test to not parse but it did"),
-                Err(e) => {
-                    if cfg!(windows) {
-                        // Normalize path separator and system error on Windows
-                        format!("{:?}", e)
-                            .replace("\\parse-fail\\", "/parse-fail/")
-                            .replace(
```
peterhuene, line 108: I think this replacement is still needed given the difference in system error messages.

peterhuene (approved):
Just the one comment as to the system error message.

PR #32, opened 2021-07-07
Implement support for resources in wasmlink. 

peterhuene:
This PR implements support for resources and handle types in wasmlink.

A small runtime will be injected into the linked module that will be
responsible for tracking all currently-live resources within every submodule.

Each defined resource is given a specific identifier so that the runtime will
segment the tracking slabs per-resource.

To facilitate this feature, `AdaptedModule` was refactored such that it can
generate the adapter in coordination with a `Resources` helper type.

The `ModuleShim` helper was folded into `AdaptedModule`.

`CallAdapter` was updated to support adapting return values not coming from a
retptr (e.g. a single-value return of a handle).

We now fully rely on `Interface::all_bits_valid` to determine if we need to
adapt a type.  This means we will generate adapters for bools, flags, and enum
variants that are currently a no-op as we don't yet validate the value's
range against the underlying variant type. A future change will make these
adapter functions more useful.

A simple test case was added where resources are exchanged between two modules.


crates/test-modules/modules/crates/resource-main/src/main.rs
```diff
@@ -0,0 +1,27 @@
+witx_bindgen_rust::import!("../../../tests/resource.witx");
```
alexcrichton, line 1: FWIW I switched these to using custom `*.witx` files inside each crate since the top-level `tests` folder is currently used by the code generators for "can I generate valid code for this interface". I should probably rename that folder at some point now that we've merged wasmlink though... In any case having to keep all this in sync can be sort of difficult, so I think it may be best to have this be a local witx file for just this crate?

crates/test-modules/modules/crates/resource-main/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "resource-main"
+version = "0.1.0"
+authors = ["Peter Huene <peter@huene.dev>"]
+edition = "2018"
```
alexcrichton, line 5: Mind throwing in `publish = false` for this (and other crates now that I think about it, but that can happen somewhere else of course)

crates/test-modules/modules/crates/resource/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+        x.0.clone()
+    }
+
+    fn receive_lots_of_x(&self, vals: Vec<Handle<X>>) -> Vec<String> {
+        vals.into_iter().map(|x| x.0.clone()).collect()
+    }
+
+    fn all_dropped(&self) -> bool {
+        *self.0.lock().unwrap() == 0
+    }
```
alexcrichton, line 39: Could this override the default-empty method of `drop_x` as well to ensure that dtor route is also fired?

crates/wasmlink/build.rs
```diff
@@ -0,0 +1,32 @@
...
+    let output = cmd.output().expect("command should execute");
+
+    let status = output.status;
+    if !status.success() {
+        panic!(
+            "Building wasmlink runtime failed: exit code: {}",
+            status.code().unwrap()
+        );
+    }
+}
```
alexcrichton, line 32: One thing I've thought is that we'll likely want to trim the runtime along the lines of removing dwarf info or the name section, although that may be best done at link-time rather than at build-time here since the size of the `wasmlink` executable probably doesn't matter too too much.

crates/wasmlink/runtime/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+}
+
+lazy_static::lazy_static! {
+    static ref SLABS: Mutex<Vec<(IndexSlab, ResourceSlab)>> = Mutex::new(Vec::new());
+}
+
+#[no_mangle]
+pub extern "C" fn resource_insert(id: u32, res: u32) -> u32 {
+    let mut slabs = SLABS.lock().unwrap();
```
alexcrichton, line 136: Depending on the generated code size this may want to avoid the use of `Mutex` and `lazy_static` since ideally this is a pretty lean runtime module.

Additionally I think that we'll need to be super careful here about "panic safety" in that an aborting `unwrap()` means that this module could still be resumed via some other entry point. 

While I know it's not implemented, `Mutex` is poisoned by default on panics so the `.unwrap()` here may want to be `.unwrap_or_else` to remove the poison (and/or avoid using a mutex altogether)

crates/wasmlink/runtime/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+
+#[no_mangle]
+pub extern "C" fn resource_get(id: u32, idx: u32) -> u32 {
+    let mut slabs = SLABS.lock().unwrap();
+
+    if slabs.len() <= id as usize {
+        slabs.resize_with(id as usize + 1, Default::default);
+    }
+
+    let (indexes, resources) = slabs.get(id as usize).unwrap();
```
alexcrichton, line 155: FWIW I think it's fine to have `id: usize` in the signature since it becomes `i32` at the wasm layer anyway

crates/wasmlink/runtime/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    pub fn drop(&mut self, idx: ResourceIndex) -> Option<u32> {
+        let resource = self.slab.get_mut(idx.0).unwrap();
+        assert!(resource.references > 0);
+        resource.references -= 1;
+        if resource.references != 0 {
+            return None;
+        }
+        let resource = self.slab.remove(idx.0).unwrap();
```
alexcrichton, line 125: There's a good number of `unwrap()` in this module which is used to abort the process, which while semantically fine makes it hard to distinguish an `unwrap()` like this where it's a bug in this module if this trips. Could a separate function be added which is specifically used for `unreachable`-on-bad-inputs? We could presumably give it a name that's indicative of "this is the consuming module's fault" and the implementation of the function could be to just call [`unreachable`](https://doc.rust-lang.org/core/arch/wasm32/fn.unreachable.html) (since presumably this module doesn't import WASI)

crates/wasmlink/build.rs
```diff
@@ -0,0 +1,32 @@
+use std::process::{Command, Stdio};
+
+fn main() {
+    println!("cargo:rerun-if-changed=runtime/src");
+    println!("cargo:rerun-if-changed=runtime/Cargo.toml");
+    println!("cargo:rerun-if-changed=runtime/Cargo.lock");
+
+    let mut cmd = Command::new("cargo");
+
+    cmd.args(&["build", "--target=wasm32-wasi"]);
```
alexcrichton, line 10: Oh below I thought this may not use WASI, but apparently it does! I think though that to make this module as lean-as-possible we may want it to use the `wasm32-unknown-unknown` target?

crates/wasmlink/build.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    if !cfg!(debug_assertions) {
+        cmd.arg("--release");
+        println!("cargo:rustc-env=RUNTIME_WASM_PATH=../runtime/target/wasm32-wasi/release/wasmlink_runtime.wasm");
```
alexcrichton, line 14: For release mode this may want to use `CARGO_PROFILE_RELEASE_LTO=yes` to force-on that? Along the lines of producing a lean-as-possible module

crates/wasmlink/runtime/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    let (indexes, resources) = slabs.get_mut(id as usize).unwrap();
+    let res_idx = indexes.remove(idx).unwrap();
+
+    // The return value's upper 32-bits is a flag to denote if the resource is still alive.
+    // If the upper 32-bits are 0, the lower 32-bits are expected to be the resource to drop.
+    match resources.drop(res_idx) {
+        Some(wasm) => wasm as u64,
+        None => 1u64 << 32,
```
alexcrichton, line 189: An alternative design is that instead of having callers check this flag the initialization of this module appends closures to its function table, and the function pointer is passed in as an argument here. That way this would perform, when the resource is actually dropped, a dynamic dispatch to the destructor.

Not that it's really any better one way or another, just an alternate possibility!

alexcrichton (commented):
I'd ideally prefer to not have to boot up too much context about the internals and nitty-gritty of `wasmlink` just yet, but one thing I think would be helpful to review changes like this is the `*.wat` of adapter modules? 

Would it be possible to have some tests of `*.wat` files where a `*.witx` interface is provided and the test asserts that the "adapter functions" generated have a particular shape? I'm imagining a `BLESS=1` style tests where it presents a picture-perfect output but you can automatically update the test assertion easiy if the codegen changes.

For me that would help a lot in reviewing what the generated module looks like and how the adaptation is happening.

Also since resources are so important, I think it would be good to have some tests specifically about invalid indices. Ideally all of the `.unwrap()` paths are tested to ensure that resources can't leak and resources also can't be accidentally used elsewhere.

alexcrichton:
I'm also totally fine iterating on all this in-tree if you want to land this now and have follow-up issues/PRs

crates/test-modules/modules/crates/resource-main/src/main.rs
peterhuene: That works for me, I'll copy each witx local to the exporting test module.

crates/test-modules/modules/crates/resource-main/Cargo.toml
peterhuene: I'll add that to each test module.

crates/test-modules/modules/crates/resource/src/lib.rs
peterhuene: I'll add that.

crates/wasmlink/runtime/src/lib.rs
peterhuene: Generally I consider resuming a trap from the embedder to be a generally-unsound operation since there wasn't an unwind of wasm code such that destructors were called and thus may be in an incongruent state.

Still, I'm definitely open to removing the mutex.

crates/wasmlink/build.rs
peterhuene: I'll see about targeting it to `wasm32-unknown-unknown`.  Are there any other tips that would reduce the size of the module any maybe scrape off any unnecessary baggage? My knowledge is a little lacking on how to get the Rust toolchain to spit out the leanest Wasm module possible.

crates/wasmlink/build.rs
peterhuene: I'll add that, thanks.

crates/wasmlink/runtime/src/lib.rs
peterhuene: I hadn't considered a `funcref` approach; I'll see what I can make happen.

peterhuene:
I'll see about adding in tests with wat baselines for easier review as well.

crates/wasmlink/runtime/src/lib.rs
peterhuene: I think that makes as lot of sense to at least indicate somehow that the problem was coming from the user's code even through it was the wasmlink runtime that trapped.

Presumably without WASI, the function name would need to suffice to communicate this, so perhaps we should keep the name section for the runtime after all.

crates/wasmlink/build.rs
alexcrichton: Yeah I agree about follow-up PRs for this!

crates/wasmlink/runtime/src/lib.rs
alexcrichton: That's a good point. I'm sort of blindly assuming that this module, injected by the linker, needs to be "super hardened" in that it'll in theory run in a lot of places. That being said I think we should probably leave this as-is and revisit this later.

FWIW I don't think any poisoning will happen today, IIRC poisoning is implemented through destructors so if a panic happens this module is already "reentrant", but only in the same way that any wasm is reentrant after a trap happens (weird things may happen in the wasm)

crates/wasmlink/runtime/src/lib.rs
alexcrichton: Yeah that's a good point, and I was thinking that the name section would be the way to communicate this, yeah.

I'm also sort of naively assuming that we *can't* use WASI here since the linker may produce binaries that don't want to use wasi at all, but we could possibly have a flag where if wasi is used (or detected already) then we could have an alternate version here which prints a more descriptive error. Anyway lots of future possibilities.

crates/wasmlink/build.rs
alexcrichton: The main thing with unknown-unknown is that it's all Rust so it's much easier to use full-crate-graph LTO which enables as many optimizations as possible to produce a small module (basically we give LLVM total knowledge). 

I figured though that we'd want to do this for functionality reasons I mentioned above, but in any case I think it's fine to go either way on this.

crates/wasmlink/runtime/src/lib.rs
alexcrichton: Oh sorry to be clear I wouldn't use `funcref` since Rust has no support for that. Rather the functions are inserted into the funcref table exported by this module (via the adapter glue) and then the index of the function to call is passed here (which is the representation of a function pointer in Rust anyway).

This would require modifying the source wasm, though, since IIRC LLD fills in the min/max of the funcref table as the same value, so you'd have to increase the max at least to be able to insert functions into it. I don't think any of this is really worth it to do at this time when what you've got here is already working.

peterhuene: Ah now I understand; yeah, given the complexity of that approach (i.e. having to post-process the runtime module), perhaps keeping a simple "branch if the upper 32-bits is non-zero" approach I have here will suffice for now.

crates/wasmlink/runtime/src/lib.rs
peterhuene: I've added a function named `invalid_handle_value` that will show up in the backtrace when the user passes an invalid index.

peterhuene:
Ok I think I have everything addressed other than adding wat-baseline tests, which I think I'll put up in a new PR.

alexcrichton:
Oh I realize now the use-the-witx-at-the-root-of-the-repo I changed, but never pushed up. It's in a local branch! In any case w/e lands I'll fix it up locally.

alexcrichton (approved):

PR #35, opened 2021-07-13
Add a C bindings generator

alexcrichton:
This commit adds the initial implementation of a C bindings generator. The intention for this generator is to be a second implementation of a "guest" language compiled to WebAssembly. The hope is that this is developed alongside the Rust implementation to ensure that anything implemented in one can also reasonably be implemented in the other.

Lots of plumbing to put things in the right places here. I haven't paid too too much attention to C idioms or customizability, but I presume those can come over time.


Hywan:
:+1: Good job!

PR #39, opened 2021-07-13
Add file-based tests to wasmlink.

peterhuene:
This PR adds file-based tests to wasmlink to make it easier to see the
generated code for adapted modules.

Additionally, `AdaptedModule` has been renamed to `ModuleAdapter` and exported
from the `wasmlink` crate.

The adapted module now conditionally imports and aliases a memory or the
canonical ABI functions relating to lists depending on whether the adapted
module needs them.


peterhuene:
Wiiiiiiindoooooooows! *shakes fist*

alexcrichton:
👍 nice!

peterhuene:
Err, yeah that failure is legit.  We shouldn't be using a static for the resource id as that only works for command line invocations of `wasmlink` and not in tests such as these.

I'll fix.

PR #40, opened 2021-07-13
Make fields of the `Tables` structure `pub(crate)`.

peterhuene:
This PR makes the fields of the Wasmtime `Tables` structure `pub(crate)`.

Without this, the underlying tables cannot be manipulated outside of the
generated code; for example, it might be useful to prepopulate some resources
in the tables prior to invoking any wasm code.


alexcrichton:
Seems reasonable to me!

PR #55, opened 2021-07-27
Initial commit of a wasmtime-py bindings generator

alexcrichton:
Currently supports everything necessary to run `host.witx`, but there
are known missing holes:

* Endianness isn't handled
* Probably not as efficient as it could be
* Legacy witx features aren't supported

This generator is intended for host embeddings of Python using the
`wasmtime` package on PyPI. This will generate Python with annotated
types for usage with `mypy` and other type checkers as well (and `mypy`
is a bit part of the tests for all of this). The idioms generated here
are a sort of blend of JS idioms where ownership doesn't matter and
Wasmtime idioms where types like the `Store` must be passed around. This
is intended to feel like idiomatic Python, but being someone who hasn't
written a lot of Python I'm sure there's room for improvement here and
there, so this is just a first pass.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,6 +52,10 @@ jobs:
     - name: Install NPM packages
       run: npm install
       working-directory: crates/gen-js
+    - uses: actions/setup-python@v1
+      with:
+        python-version: 3.9
+    - run: pip install mypy wasmtime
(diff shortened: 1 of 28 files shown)


alexcrichton:
While I was at this I took the liberty of rewriting the README and the language support section to be a little less repetitive and also extend it a bit with future-facing words.

PR #58, opened 2021-08-07
Implement support for multiple exported interfaces in wasmlink.

peterhuene:
Currently `wasmlink::Module` only supports reading an interface from a
custom section (unused) or from a path to a witx file.

In preparation for some upcoming work, this commit implements creating a
`wasmlink::Module` that exports more than one `witx2::Interface`.

`Module::new` now checks that given interfaces are not in conflict (e.g.
have a function/resource of the same name) and also verify that the
module actually does export those interfaces.

Also fixed some clippy lints in wasmlink.


peterhuene:
cc: @radu-matei

alexcrichton (approved):
Seems reasonable!

radu-matei (approved):

PR #60, opened 2021-08-09
Remove requirement that the module to link exports `_start`.

### LINKED ISSUE #59: Using linked module from the Wasmtime generated code results in missing `canonical_abi_free` export
The use case is the following: I am using the generated export Wasmtime bindings to execute a specific function through Wasmtime. This works fine for modules that have not been linked, but fails for modules that import any interfaces and need to be linked.

(First, the module must export a `_start` function. Adding a dummy `_start` bypasses this) 

The error I'm seeing: "failed to find function export `canonical_abi_free`".

The Wasmtime bindings explicitly try to use the free function (when using strings in the function signatures), and apparently the linked module does not re-export it.

A couple of questions:
- is there a reason to explicitly require the `_start` function before linking a module?
- is there a way to re-export the `canonical_abi_free` function (and I assume the same is valid for the `realloc` function)?

Thanks!

cc @peterhuene

peterhuene:
This PR removes the requirement that the root module being linked
exports a `_start` function.

Instead, it now simply reexports all exported functions.  This will
enable use of the linked modules from profiles where a start function
isn't necessary (i.e. the host will call some other function).

Updated the CLI to remove references to a "main" module.

Fixes #59.

### DIFF
--- a/crates/wasmlink-cli/src/lib.rs
+++ b/crates/wasmlink-cli/src/lib.rs
@@ -41,13 +41,13 @@ pub struct App {
     #[structopt(long, short = "p", value_name = "PROFILE")]
     pub profile: String,
 
-    /// The path of the output linked module; defaults to replacing the main module.
+    /// The path of the output linked module; defaults to replacing the given module.
     #[structopt(long, short = "o", value_name = "OUTPUT", parse(from_os_str))]
     pub output: Option<PathBuf>,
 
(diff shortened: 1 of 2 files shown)


peterhuene:
cc @radu-matei

alexcrichton (approved):

radu-matei (approved):

PR #61, opened 2021-08-11
Make generated export Wasmtime bindings module public

radu-matei:
This commit updates the Wasmtime export bindings module's visibility to public.
This makes it very convenient to directly use the generated module in a crate.

Signed-off-by: Radu M <root@radu.sh>


alexcrichton:
Seems reasonable to me!

PR #62, opened 2021-08-12
Add #ifdef to C headers so they can be directly reused in CPP files

radu-matei:
This commit enables the C bindings to be imported directly by a C++ source file.

(Note that the C implementation will still have to be separately compiled with the C compiler.)

Signed-off-by: Radu M <root@radu.sh>

### DIFF
--- a/crates/gen-c/src/lib.rs
+++ b/crates/gen-c/src/lib.rs
@@ -1056,6 +1056,11 @@ impl Generator for C {
     fn finish(&mut self, iface: &Interface, files: &mut Files) {
         self.src.h(&format!(
             "\
+                #ifdef __cplusplus
+                extern \"C\"
+                {{
+                #endif
+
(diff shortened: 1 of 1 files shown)


crates/gen-c/src/lib.rs
```diff
@@ -1056,6 +1056,11 @@ impl Generator for C {
     fn finish(&mut self, iface: &Interface, files: &mut Files) {
         self.src.h(&format!(
             "\
+                #ifdef __cplusplus
+                extern \"C\"
+                {{
+                #endif
+
```
sunfishcode, line 1063: Could this go inside the `#ifndef __BINDINGS_{0}_H` guard below? That makes it easier for C++ compilers to recognize the include-guard idiom and avoid opening a header multiple times.

alexcrichton:
👍

PR #71, opened 2021-09-20
Introduce a WITX bindings generator for JS via SpiderMonkey

fitzgen:
This `witx-bindgen` backend targets JS running on Wasm via the SpiderMonkey JS
engine. Unlike other `witx-bindgen` backends, this one does not generate source
code. Instead, it generates Wasm glue code directly. The goal is to avoid
requiring that users have a working C++ toolchain that targets Wasm. Users just
supply their JS file and the WITX interfaces to `witx-bindgen-spidermonkey`, and
don't need any other tools.

## Code Shape

The output is a single Wasm file that imports and exports the functions defined
in the given WITX files and additionally

* embeds or imports (configurable) a `spidermonkey.wasm` instance, and

* exports a `wizer.initialize` function that initializes SpiderMonkey and
  evaluates the top level of the JavaScript.

### Initialization

As an API contract, the `wizer.initialize` function must be invoked before any
other function. It must only be invoked once.

The initialization function performs the following tasks:

* Calls `spidermonkey.wasm`'s `_initialize` function, which runs C++ global
  contructors.

* `malloc`s space in `spidermonkey.wasm`'s linear memory and copies the
  JavaScript source code from its linear memory into the malloc'd space.

* Evaluates the JavaScript source, compiling it to bytecode and initializing
  globals and defining top-level functions in the process.

### Imports

By the time an imported WITX function is called, we have the following layers of
code on the stack, listed from older to younger frames:

* User JS code (inside `spidermonkey.wasm`'s internal JS stack)

  This is the user's JavaScript code that is running inside of
  `spidermonkey.wasm` and which wants to call an external, imported function
  that is described with WITX.

* Import glue Wasm code (on the Wasm stack)

  This is a synthesized Wasm function that understands both the canonical ABI
  and the SpiderMonkey API. It translates outgoing arguments from SpiderMonkey
  values into the canonical ABI representation, calls the actual imported Wasm
  function, and then translates the incoming results from the canonical ABI
  representation into SpiderMonkey values.

* Imported function (on the Wasm Stack)

  This is the actual Wasm function whose signature is described in WITX and uses
  the canonical ABI.

### Exports

By the time an exported JS function that implements a WITX signature is called,
we have the following frames on the stack, listed form older to younger frames:

* External caller (on the Wasm or native stack)

  This is whoever is calling our JS-implemented WITX export, using the canonical
  ABI. This might be another Wasm module or it might be some native code in the
  host.

* Export glue Wasm code (on the Wasm stack)

  This is a synthesized function that understands both the canonical ABI and the
  SpiderMonkey API. It translates incoming arguments from the canonical ABI
  representation into SpiderMonkey values, calls the JS function that implements
  this export with those values, and then translates the function's outgoing
  results from SpiderMonkey values into the canonical ABI representation.

* JavaScript function implementing the WITX signature (inside
  `spidermonkey.wasm`'s internal stack)

  This is the user-written JavaScript function that is being exported. It
  accepts and returns the JavaScript values that correspond to the interface
  types used in the WITX signature.

## `Bindgen` Implementation

Most `witx2::abi::Instruction`s have a corresponding "intrinsic" function that
is exported from `spidermonkey.wasm`, e.g. there is
`witx2::abi::Instruction::I32FromU32` and `spidermonkey.wasm` exports the
`SMW_i32_from_u32` function. These intrinsics are used when emitting Wasm
opcodes directly is too difficult. The intrinsics are defined in
`spidermonkey-wasm/bindgen.cpp`.

We keep track of whether operands are JS values or Wasm locals.

## Implementation Status

`witx-bindgen-spidermonkey` currently supports the following types:

* `u32`
* string
* list

It is expected that the other scalar types will be relatively easy to add
support for. They should look basically the same as `u32`.

It is expected that implementing structs and variants will be more difficult
than the other scalars, but no more difficult than implementing lists already
was.

Either way, this seemed like a good point to check in and share the progress
I've made more widely.


crates/gen-spidermonkey/src/lib.rs
```diff
@@ -0,0 +1,2180 @@
...
+                let locals: Vec<_> = sig
+                    .results
+                    .iter()
+                    .map(|ty| {
+                        self.new_local(match ty {
+                            WasmType::I32 => wasm_encoder::ValType::I32,
+                            WasmType::I64 => wasm_encoder::ValType::I64,
+                            WasmType::F32 => wasm_encoder::ValType::F32,
+                            WasmType::F64 => wasm_encoder::ValType::F64,
+                        })
```
alexcrichton, line 1972: This might make a good helper function (the `match` statement) since I think I saw this somewhere else as well?

crates/gen-spidermonkey/src/lib.rs
```diff
@@ -0,0 +1,2180 @@
...
+    fn finish_block(&mut self, results: &mut Vec<Self::Operand>) {
+        self.block_results.push(results.to_vec());
+    }
+
+    fn sizes(&self) -> &witx2::SizeAlign {
+        todo!()
+    }
+
+    fn is_list_canonical(&self, _iface: &witx2::Interface, _ty: &witx2::Type) -> bool {
+        false
```
alexcrichton, line 2178: Oh I see this might answer my question from earlier. Does this mean though that things like `list<u8>` have element-per-element processing rather than something that lowers down to a `memory.copy`? If so I think it might be good to have a `TODO` here to at least change that for `TypedArray`-style things.

alexcrichton (approved):
Looking quite good to me! The main bits I think are:

* I think it'd be best to unify the `witx-bindgen` CLI interface with what spidermonkey-wasm needs to avoid having something duplicate.
* Out of curiosity have you had any issues with memory leaks? There's a lot of allocations happening in the bindings and while I didnt' look super closely at the C++ I'm curious if you had ideas about how to track this and try to ensure that nothing is accidentally leaked.
* I do suspect that some things are going to still be somewhat significantly complicated.  Things like records/variants are pretty nontrivial each in their own right, but that doesn't mean this can't all land ahead of time of course.

fitzgen:
> * I think it'd be best to unify the `witx-bindgen` CLI interface with what spidermonkey-wasm needs to avoid having something duplicate.

The interface to `SpiderMonkeyWasm` is a little different than the other generators: 

* It needs to preprocess all the WITX interfaces at once, not one at a time, before generating bindings. This is necessary so we can precompute function indices.
* Also, we don't incrementally generate bindings as each interface is processed and then when you're done processing all interfaces the `Files` is up-to-date. Instead we don't emit any files until after everything is done being processed and you call `into_wasm`, because all the bindings end up in the same Wasm file.
* Also we need to take a JS file as input, where the other bindings generators only take WITX files.

So I guess the question is whether we want this in the same CLI, even if it takes its own unique code path that is separate from all the other bindings generators. I guess that makes sense to me, so there is still just one tool. But trying to shoe-horn `SpiderMonkeyWasm` into the same code path as the other generators isn't going to work out very well.

> * Out of curiosity have you had any issues with memory leaks? There's a lot of allocations happening in the bindings and while I didnt' look super closely at the C++ I'm curious if you had ideas about how to track this and try to ensure that nothing is accidentally leaked.

Some strings are intentionally leaked, because in a regular toolchain they would be static data that is assigned a static location by the linker. Instead we `malloc` them during initialization (and then eventually freeze them via Wizer).

When owned strings are passed into C++ I try to immediately wrap them in `mozilla::UniquePtr` (the mozilla C++ equivalent of `Box`) so that the ownership is immediately visible and we can't forget to drop the strings. So that's a mitigation, but I don't actually have anything *checking* for leaks here (or double frees, etc). Would be cool to port https://github.com/rustwasm/wasm-tracing-allocator to wasm32-wasi and module linking. Someday.

-------------------------------

More replies inline coming.

alexcrichton:
Would it be possible to tweak the bindgen trait to have a top-level method that takes a list of import interfaces and a list of export interfaces? That would have a default implementation that calls today's helper methods, and all existing generators would probably just panic if there's more than one interface (or generate invalid code or something like that with a TODO to get fixed later). Each generator already has custom options via the structopt flattening so I think spidermonkey can ingest the JS file this method as well.

I basically want to avoid having this large spidermonkey thing on the side that sits for a long time and is very noticeable different than all the other generators. The existing interface has change a good deal for all new generators I've added, so I think iterating towards something that encompasses everything is important.

For allocations, would it be possible to do what the rust-wasm bits do today and have a counter of allocated bytes updated on the canonical alloc/free functions? If all allocs are then routed through those functions when allocated across the boundary then tests can use some sort of intrinsic to  test that the number of allocated bytes is the same before/after the test.

fitzgen:
> Would it be possible to tweak the bindgen trait to have a top-level method that takes a list of import interfaces and a list of export interfaces? That would have a default implementation that calls today's helper methods, and all existing generators would probably just panic if there's more than one interface (or generate invalid code or something like that with a TODO to get fixed later). Each generator already has custom options via the structopt flattening so I think spidermonkey can ingest the JS file this method as well.
> 
> I basically want to avoid having this large spidermonkey thing on the side that sits for a long time and is very noticeable different than all the other generators. The existing interface has change a good deal for all new generators I've added, so I think iterating towards something that encompasses everything is important.

This makes sense. I'll poke at it.

> For allocations, would it be possible to do what the rust-wasm bits do today and have a counter of allocated bytes updated on the canonical alloc/free functions? If all allocs are then routed through those functions when allocated across the boundary then tests can use some sort of intrinsic to test that the number of allocated bytes is the same before/after the test.

Good idea. I'll add this.

crates/gen-spidermonkey/src/lib.rs
```diff
@@ -0,0 +1,2180 @@
...
+            export_glue_fns: Default::default(),
+            data_segments: DataSegments::new(1),
+            sizes: Default::default(),
+        }
+    }
+}
+
+lazy_static! {
+    /// Functions exported from `spidermonkey.wasm`
+    static ref SMW_EXPORTS: Vec<(&'static str, WasmSignature)> = vec![
```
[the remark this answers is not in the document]
fitzgen: What would the benefit of that be?

This is "static" data, but can't be an actual `static` because `WasmSignature` contains `Vec`s. So if we were to put this inside `SpiderMonkeyWasm` then it would just be dynamically creating copies of the same static data for every `SpiderMonkeyWasm` instance, which seems unnecessary?

fitzgen:
Addressed most of the comments. Still need to expand the `Generator` interface so that I can make `SpiderMonkeyWasm` fit in the CLI better and also get `test-codegen` to work with it.

fitzgen:
@alexcrichton want to take a look at the two recent commits? These included tweaking the `Generator` trait (see the commit message for more details).

Tomorrow I will get the `test-codegen` framework working with `witx-bindgen-gen-spidermonkey`.

crates/gen-spidermonkey/src/lib.rs
```diff
@@ -2,10 +2,16 @@
 
 #![deny(missing_docs)]
 
+#[cfg(feature = "structopt")]
+mod opts;
+#[cfg(feature = "structopt")]
```
alexcrichton, line 7: FWIW for the other generators I've been having a conditional derive of `structopt` where the main "builder" has configuration options on it for libraries and such. That may be applicable here too?

alexcrichton:
Seems reasonable to me! I'll ping you about https://github.com/bytecodealliance/witx-bindgen/pull/72 since it maybe relevant here.

crates/gen-spidermonkey/src/lib.rs
fitzgen: I just didn't want `SpiderMonkeyWasm` construction to be fallible and do I/O. It feels cleaner to pass in the JS source string as a parameter to the generator, rather than pass the file path and expect it to read it for you.

fitzgen:
> Seems reasonable to me! I'll ping you about #72 since it maybe relevant here.

Yeah, I saw that. Seems great to have more little tests. Haven't looked at implementation at all yet.

fitzgen:
Oh also we might want to always build `cranelift-codegen` with optimizations and no debug assertions, even in dev profiles, because otherwise compileing `SpiderMonkey.wasm` takes 3 minutes on my laptop (vs 1 second with optimizations and no debug assertions).

I'll add a commit.

Cargo.toml
```diff
@@ -27,3 +28,9 @@ witx-bindgen-gen-wasmtime-py = { path = 'crates/gen-wasmtime-py', features = ['s
 witx-bindgen-gen-js = { path = 'crates/gen-js', features = ['structopt'] }
 witx-bindgen-gen-c = { path = 'crates/gen-c', features = ['structopt'] }
 witx-bindgen-gen-markdown = { path = 'crates/gen-markdown', features = ['structopt'] }
+witx-bindgen-gen-spidermonkey = { path = 'crates/gen-spidermonkey', features = ['structopt'] }
+
+# Compiling `spidermonkey.wasm` takes way too long without this.
+[profile.dev.package.cranelift-codegen]
+debug-assertions = false
+opt = 2
```
alexcrichton, line 36: I think this may need to be `opt-level` instead of `opt`?

crates/gen-spidermonkey/spidermonkey-wasm/Makefile
```diff
@@ -0,0 +1,142 @@
+SM_REPO := https://github.com/fitzgen/gecko-dev
```
alexcrichton, line 1: Oh one thing I forgot to ask, would it make sense to build this on CI and ensure that `spidermonkey.wasm` in the repo is the same?

crates/gen-spidermonkey/src/lib.rs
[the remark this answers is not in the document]
alexcrichton: I've always felt that `lazy_static` is a non-rust-ism since it just leaks everything. I don't really care either way though.

alexcrichton:
Ok I'm gonna merge this and we can continue to iterate on specifics in-tree

crates/gen-spidermonkey/spidermonkey-wasm/Makefile
fitzgen: In an ideal world, yes. Do we want to spend that CI time though?

PR #78, opened 2021-09-24
Add initial support for `async` functions in witx

alexcrichton:
This commit is the initial steps towards supporting and implementing
`async` function as-specified in `*.witx` itself. The goal of this is to
implement the callback-based ABI of async functions and streams as
presented in WASI subgroup meetings.

This commit itself only implements async functions for the
compiled-to-Rust guest and a JS host. The next PR is to implement async
functions for a Wasmtime host, followed by an implementation of streams
afterwards.

Included here is:

* Updates to the `*.witx` syntax and parsing.
* Assertions in code generators that don't support async that async
  isn't used.
* An implementation of the callback ABI for `async` functions for
  compiled-to-wasm Rust modules. This includes a simple "executor" to
  track the state of async imports/exports. Everything is only exposed
  with `async` functions in Rust, no callbacks are part of the interface
  of the generated bindings.
* An implementation of the callback ABI for JS hosts. This is
  exposed with `async` functions and only uses callbacks under-the-hood.

This change had some significant ramifications on the generated bindings
for JS hosts. Namely both imports and exports needed to access the same
state of callbacks so they are no longer generated separately in
isolation but rather together in the same manner as the spidermonkey
bindings. I expect that most generators will migrate to this pattern in
the future. Work was done internally to deduplicate intrinsics across
bindings modules, namely the promise-related intrinsics.

The testing so far isn't super strenuous, but a major thing I made sure
to implement is that when an asynchronous wasm function is called and
later some asynchronous import throws an exception that should carry the
exception to the original wasm export's promise in JS. This is intended
to map to basically what `async` and `await` feel like in JS normally.
Otherwise though there's codegen tests for flavorful types being used
in parameters/results for async functions (and this test is ignored in
most generators at the moment) and runtime tests for some concurrency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -571,6 +571,48 @@ dependencies = [
 ]
 
 [[package]]
+name = "futures-core"
+version = "0.3.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "88d1c26957f23603395cd326b0ffe64124b818f4449552f960d815cfba83a53d"
+
+[[package]]
+name = "futures-macro"
+version = "0.3.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "18e4a4b95cea4b4ccbcf1c5675ca7c4ee4e9e75eb79944d07defde18068f79bb"
+dependencies = [
+ "autocfg",
+ "proc-macro-hack",
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "futures-task"
+version = "0.3.17"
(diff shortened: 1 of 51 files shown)


alexcrichton:
One thing I'm not happy about here is how the Rust exposes exported functions. Currently you're required to define a `struct` of a particular name and implement a generated trait for the struct. By using traits this means that we have to use `async_trait` which is a bit of a bummer since it's not really otherwise necessary. I've never felt great about this interface, though, and this is basically just adding more fuel to the fire of trying to figure out something better. The main reason I used traits originally was because the type errors are great if you don't implement something or use the wrong signature. I think we'll just have to give up on that though.

PR #79, opened 2021-09-24
Fix names defining resource intrinsics in Wasmtime

### LINKED ISSUE #76: canonical_abi has not been defined
I have a very trivial setup to play with witx types. To reproduce the problem just run `cargo run` from the root directory. (requires nightly)  Use a vscode devcontainer to reproduce my exact environment.

https://github.com/carlsverre/wasi-witx-sandbox/tree/play_with_types

Generated code for both sides is in the repo as well in `output.rs`:
https://github.com/carlsverre/wasi-witx-sandbox/blob/play_with_types/host/output.rs
https://github.com/carlsverre/wasi-witx-sandbox/blob/play_with_types/component/output.rs

The fundamental issue seems to be that the code generated for the host is referencing canonical_abi which doesn't seem to be linking correctly. But honestly I am still trying to wrap my head around this repo so I probably just made a mistake.

The host's code is in `host/src/main.rs` and the clients code is in `component/src/lib.rs`.

alexcrichton:
There was an accidentaly `.to_snake_case()` when it should have been
using the raw name from the witx.

Closes #76

### DIFF
--- a/crates/gen-wasmtime/src/lib.rs
+++ b/crates/gen-wasmtime/src/lib.rs
@@ -990,20 +990,21 @@ impl Generator for Wasmtime {
                     self.src.push_str(&format!(
                         "linker.func_wrap(
                             \"canonical_abi\",
-                            \"resource_drop_{}\",
+                            \"resource_drop_{name}\",
                             move |mut caller: wasmtime::Caller<'_, T>, handle: u32| {{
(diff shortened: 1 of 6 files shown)


carlsverre:
Thanks for the fast turn around! Works great!

PR #86, opened 2021-10-18
Swap `import` and `export` in the Wasmtime bindings API.

sunfishcode:
Swap the names of the `import` and `export` macros in the Wasmtime bindings
API. This makes them the opposite of the `Direction` argument to the
underlying generator code.

As background, `Import` and `Export` in the generator code imply different
ABIs. In a wasm-calling-wasm use case, one wasm module would use the `Import`
ABI, the other would use the `Export` ABI, and there would be an adapter
layer between the two that translates from one ABI to the other.

With wasm-calling-host, we don't go through a separate adapter layer;
the binding code we generate on the host side just does everything
itself. So when the host is conceptually "exporting" a function to
wasm, it uses the `Import` ABI so that wasm can also use the `Import`
ABI and import it directly from the host.

The key is, from the user's perspective, these are all just
implementation details. For users, `export` means exporting functions
to be called, and `import` means importing functions to call, in both
wasm modules and host code.

This is also described in comments in crates/wasmtime-impl/src/lib.rs.

### DIFF
--- a/crates/test-helpers/src/lib.rs
+++ b/crates/test-helpers/src/lib.rs
@@ -14,7 +14,7 @@ use witx_bindgen_gen_core::Generator;
 #[derive(PartialEq, Eq, Copy, Clone)]
 enum Direction {
     Import,
-    Export
+    Export,
 }
 
 #[proc_macro]
--- a/crates/wasmtime-impl/src/lib.rs
+++ b/crates/wasmtime-impl/src/lib.rs
@@ -27,7 +27,7 @@ use witx_bindgen_gen_wasmtime::Async;
 /// modules and host code. The enum here represents this user perspective.
 enum Direction {
     Import,
-    Export
+    Export,
 }
 
 /// Generate code to support consuming the given interfaces, importaing them


alexcrichton:
Hm this is one of the reasons I fear about this change is that I've at least tried to be consistent everywhere that import/export is used but this seems to be introducing internal inconsistency which is hard to follow. I realize the import/export definitions are confusing and I want to fix that, but I'm not sure that making the internals swap around in various places is the best solution? I'm not sure why new `Direction` enums that all look the same are added here or why new filters needed to be added to tests?

I was mostly hoping that this could just be a surface-level distinction in the macro and the CLI, but everything else internally would use import/export as it does today.

sunfishcode:
I did start with the simple surface-level thing, but it quickly got confusing because the CLI shares a bunch of code between the various bindings.

The new test filters are because I removed this special-filter that was previously hard-coded in the harness:
```rust
        if dir == Direction::Export {
            match test.file_name().unwrap().to_str().unwrap() {
                "wasi_snapshot_preview1.witx" | "typenames.witx" | "legacy.witx" => continue,
                _ => {}
            }
        }
```
because in that context, `dir` is the user direction, not the ABI direction, and I didn't want to have it to special-case "wasmtime" here.

The new `Direction` enums are to make the distinction between "user direction" and "abi direction", so that we don't accidentally pass one to code expecting the other. It might be better to have a single "user direction" enum though; I'm open to suggestions as to where that should live.

Another possible change that might help would be to rename `witx2::abi::Direction` to something like `AbiDirection` or `AbiVariant` or so, so that `Direction` only ever means the user-visible direction. Would that help?

PR #88, opened 2021-10-18
Implement host profile parsing.

peterhuene:
This PR implements host profile parsing.

Profiles are expected to have a file extension of `profile`.

Profiles are pretty simple as they have four possible statements:

* `extend` : used to extend one profile from another.
* `provide` : used to specify an interface provided by the host.
* `require` : used to specify an interface required by the host.
* `implement` : used to specify a default interface implementation for a host.

### DIFF
--- a/crates/witx2/src/ast/profile.rs
+++ b/crates/witx2/src/ast/profile.rs
@@ -12,7 +12,7 @@ pub struct Ast<'a> {
 }
 
 impl<'a> Ast<'a> {
-    pub fn parse(input: &'a str) -> Result<Ast<'a>> {
+    pub fn parse(input: &'a str) -> Result<Self> {
         let mut lexer = Tokenizer::new(input);
         let mut items = Vec::new();
 
@@ -25,6 +25,30 @@ impl<'a> Ast<'a> {
     }
 }
 
+pub struct Id<'a> {
+    pub name: Cow<'a, str>,
+    pub span: Span,
+}
+
+impl<'a> Id<'a> {
+    fn parse(tokens: &mut Tokenizer<'a>) -> Result<Self> {
+        match tokens.next()? {
+            Some((span, Token::Id)) => Ok(Id {
+                name: tokens.get_span(span).into(),
+                span,
+            }),
+            Some((span, Token::StrLit)) => Ok(Id {
+                name: tokens.parse_str(span).into(),
+                span,
+            }),
+            other => {
+                let (span, msg) = tokens.format_expected_error("an identifier or string", other);
+                bail!(Error { span, msg })
+            }
+        }
+    }
+}
+
 pub enum Item<'a> {
     Extend(Extend<'a>),
     Provide(Provide<'a>),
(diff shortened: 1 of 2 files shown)


peterhuene:
This PR is still a WIP as I'm adding the test coverage for profiles today.

It also contains a refactoring of the witx2 tokenizer to take the token type as a generic type parameter, thereby allowing a sharable implementation with differing token types (shares parsing of whitespace, comments, and strings).

Contains clippy fixes as well.

peterhuene:
> I got about halfway through this when I started thinking more about this profile, and I was wondering if it may also be worth it to support basically inline syntax for this? Something like:
> 
> ```
> require foo {
>   the_fn: function()
> }
> 
> provide bar {
>   other_function: function()
> }
> ```
> 
> where auxiliary `*.witx` files aren't actually required and it's all contained within one file? That may actually be nice to automatically implemente support for in the CLI as well where a `--profile` flag (or just an argument with `*.profile`) would automatically translate to the `--import` and `--export` flags of today. Similarly this could help sharing things where we don't have to share multiple files but rather just the simple `*.profile`. Finally it wouldn't need to refactor the lexer with two kinds of tokens but we could go ahead and just have the same set of tokens for the whole format.

I considered supporting inline interface definitions, but the end goal here is to strongly prefer referring to interface entities coming from a registry with some sort of unique identifier (e.g. `foo/bar/1.2.3` using bindle's naming scheme), which the component tooling will handle resolving with `parse_with`.

That said, inline definitions are really no different from loading one from a local file, so we should either only support a custom resolver (i.e. everything is a string which the tooling resolves) or allow resolution from files and also inline definitions by default.  Okay...I think you've convinced me to support inline definitions too 👍, but we'll absolutely need a way for the tooling to resolve external interface definitions.  I'll consolidate the lexer implementation too.

alexcrichton:
Hm well now I'm also less certain on inline syntax. We also don't have inline syntax in `*.witx` itself for things like `use` directives in that you can't define the module inline it's required to be somewhere else on the filesystem (which I see as a bit of a bummer). I was theoretically hoping that this `*.profile` syntax could be a collections of `*.witx` modules that could `use` between one another but that may not 100% work out either.

I could go either way sort of and could see this just evolving over time as we see fit, but in either case personally I don't think it's worth it to have a generic tokenizer and it should probably be fine to have just one lexer.

peterhuene:
I'll revert the changes to the tokenizer and consolidate the token implementations; the intent was to have them completely separate to cut down on possible sources of confusion (e.g. parsing a profile might say `unexpected keyword 'function'` when there's really no such keyword in a profile). It also means that uses will have to quote their identifiers in witx that just so happen to be profile keywords (e.g. `"extend": function(...)`), which might be confusing to witx users unfamiliar with profiles (much more developers will use witx to define their component than author host profiles).

Re: the inline syntax for profiles open question: perhaps we can revisit this in the future if it proves needed? The current requirement of profiles is to easily source what interfaces are provided to a component, required from a component, and what implementations should be used for the host environment for component interface dependencies.

peterhuene:
@alexcrichton okay I've pushed up some requested changes. I think it makes enough sense to keep interface and profile parsing completely separate so they can evolve entirely independently of one another; to that end, I reorganized the code a bit so that it's clear that interface code goes here and profile code goes there.

Happy to undo that if you aren't a fan.

PR #90, opened 2021-10-20
Add support for parsing `.witx.md` files

tschneidereit:
This adds support for extracting `witx` code blocks from Markdown files, and then using those as input for the `witx` parser.

The idea is to make it easy to add documentation that shouldn't go into API documentation, e.g. because it's about implementation notes not relevant to API users.

This works, in that it results in the right AST, and hence correct bindings being generated, etc. However, it really only "works" in its current state, because of two limitations:
1. The emitted bindings have `.witx` in their name. E.g. for an input file `test.witx.md`, JS bindings will be generated under the name `test.witx.js`, instead of the expected `test.js`.
2. Since code blocks are naively extracted and then smooshed together, all the line offsets are wrong, which is bad for error diagnostics.


tschneidereit:
Ok, this now pads the unwrapped source with newlines as needed, so lines in error reporting are correct. Thank you for the suggestion, @sunfishcode!

alexcrichton:
Oh oops it appears I was late to the party about error messages, could some tests be added in the `parse-fail` directory to make sure the error messages look ok?

PR #98, opened 2021-10-27
Rename Direction to AbiVariant.

sunfishcode:
And rename Import and Export to GuestImport and GuestExport.

As discussed in #94, this renames `witx2::abi::Direction` to `witx2::abi::AbiVariant`, and its `Import` and `Export` to `GuestImport` and `GuestExport. More refactoring is needed; this is a first step. I'm also open to other names here.

The big picture is that the direction switch happens at the `Generator` interface. Code calling the `Generator` interface uses `Direction` and the developer perspective, and code inside `Generator` implementations use `AbiVariant` and the ABI perspective. In particular, the `imports` and `exports` members of the `Wasmtime` type are inside the wasmtime `Generator` implementation, so they use the ABI perspective.

The `preprocess_one` is the one outstanding exception to this I'm aware of; it currently takes an `AbiVariant` argument passed in from the outside. ~~If this overall plan makes sense, I'll submit a follow-up PR to adjust `preprocess_one` to be consistent with the `Generator` boundary.~~ Edit: It was pretty straightforward, so I've now added a patch to adjust `preprocess_one` to this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,11 +127,9 @@ $ witx-bindgen rust-wasm --export my-interface.witx
 $ witx-bindgen wasmtime --import host-functions.witx
 ```
 
-Here "import" means "imported by WebAssembly" and "export" means
-"exported by WebAssembly". This means that wasm files import APIs with
-`--import`, whereas hosts provide imports to wasm modules using `--import` (the
-terminology here [can be
-confusing](https://github.com/bytecodealliance/witx-bindgen/issues/34)
+Here "import" means "I want to import and call the functions in this interface"
+and "export" means "I want to define the functions in this interface for others
+to call".
 
 Finally in a sort of "miscellaneous" category the `witx-bindgen` CLI also
 supports:
--- a/crates/gen-c/src/lib.rs
+++ b/crates/gen-c/src/lib.rs
@@ -2,7 +2,7 @@ use heck::*;
 use std::collections::{BTreeSet, HashMap, HashSet};
 use std::mem;
 use witx_bindgen_gen_core::witx2::abi::{
-    Bindgen, Bitcast, Direction, Instruction, LiftLower, WasmType, WitxInstruction,
+    Bindgen, Bitcast, AbiVariant, Instruction, LiftLower, WasmType, WitxInstruction,
 };
 use witx_bindgen_gen_core::{witx2::*, Files, Generator, Ns};
 
@@ -703,9 +703,9 @@ impl Return {
 }
 
 impl Generator for C {
-    fn preprocess_one(&mut self, iface: &Interface, dir: Direction) {
+    fn preprocess_one(&mut self, iface: &Interface, dir: AbiVariant) {
         self.sizes.fill(dir, iface);
-        self.in_import = dir == Direction::Import;
+        self.in_import = dir == AbiVariant::GuestImport;
 
         for func in iface.functions.iter() {
             let sig = iface.wasm_signature(dir, func);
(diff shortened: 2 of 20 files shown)


sunfishcode:
#### Component author vision:

Suppose we wrote a new crate, `witx-bindgen-component`, which expands to either `witx_bindgen_rust::export` or `witx_bindgen_wasmtime::export`, depending on how it's being compiled. That way, a user source file could look like this:

```rust
witx_bindgen_component::export!("something.witx");

struct Something;

impl something::Something for Something {
    fn frob(&mut self) -> u32 {
        // do stuff
     }
}
```

(Such a crate isn't quite possible today, because witx-bindgen-rust currently emits a trait that doesn't have `self` arguments. But suppose we changed witx-bindgen-rust to implicitly create a `Something` in static memory, and then pass that into the trait functions, so that they always have a `&mut Self` argument.)

That way, the same source file can be compiled either as an actual wasm component, or, conceptually, a "host component".

To be sure, developers would need to avoid accessing static memory directly for this to work. But with that, developers opting into this model could think in terms of just writing components. Even when they're writing host code, they're still writing components. Maybe some day they'll want to migrate some of their wasm components into host components, or vice versa. Or maybe have a component that can be compiled two different ways, once in wasm and once in host code, using conditional compilation to provide extra features in the host case.

And as we look forward to greater Cargo integration, I think this mental model will continue to be useful. "Import" and "export" will still be part of the developer vocabulary, whether it's proc macros or Cargo.toml attributes or something else.

#### On the inside:

The basic idea is that witx-bindgen recognizes two distinct perspectives:

| Interface Types | Canonical ABI |
| ---------- | ------- |
| Components link with components\* | Hosts\*\* link with guests |
| *export* means another component is calling me | *export* means the host is calling the guest, aka "guest export" |
| *import* means I'm calling another component | *import* means the guest is calling the host, aka "guest import" |
| Code using the `Generator` trait works at this level | `Generator` trait implementations work at this level |
| `Direction` | `AbiVariant` |
| Component authors will only think at this level | Only toolchain developers will think at this level |

\* or component-like things, like host functions
\*\* or guests with wasmlink-like adapters

The current PR is mostly complete in terms of separating these two perspectives. The only further change I have in mind is merging `Generator::import` and `Generator::export` and just have a single function that takes a `Direction`. But either way, the current PR feels to me like it makes a pretty clear separation. As I was figuring this out, `Generator` repeatedly emerged as a natural place to draw the line. But I'm happy to iterate on this to figure out how best to organize things.

alexcrichton:
Hm ok yeah that makes a lot of sense to me. I like the idea of "export means to implement a `trait`" and "import means here's some functions you can call". While I'm pretty doubtful that the rust-compile-to-wasm and Wasmtime traits will ever look *exactly* the same at a high level I think the parallels make sense.

I'm not sure I understand what you mean though by differentiating interface types and the canonical ABI. Those are in theory both parts of the same standard, so how come they have different interpretations of the same words?

I think the another thing I'd still like to see done though is to improve what the generators are doing internally. All of their internal fields about imports/exports/etc are all built on my historical notion of guest import/export which are now all backwards from what they should be for hosts (js/python/wasmtime).

sunfishcode:
> I'm not sure I understand what you mean though by differentiating interface types and the canonical ABI. Those are in theory both parts of the same standard, so how come they have different interpretations of the same words?

Even though we currently define them in the same standard, the canonical ABI is a lower level of abstraction. The canonical ABI is just one possible ABI for implementing interface types.

"import" and "export" each have obvious meanings at each level, but the levels have some things that go in opposite ways (canonical ABI is asymmetric while IT is symmetric), which is what makes it confusing.

What if we renamed GuestImport and GuestExport to GuestCallingHost and HostCallingGuest? That's a little more verbose, but more descriptive.

> I think the another thing I'd still like to see done though is to improve what the generators are doing internally. All of their internal fields about imports/exports/etc are all built on my historical notion of guest import/export which are now all backwards from what they should be for hosts (js/python/wasmtime).

They're inside the `Generator` impls, so they're correctly using the canonical-ABI perspective. I've now renamed those `imports` and `exports` to `guest_imports` and `guest_exports` to match the `AbiVariant` enun names; does that help?

alexcrichton:
We talked a bit offline and I think I'll just deal with changes on my end.

Issue #99, opened 2021-11-04
support generating cffi's


dvc94ch:
briefly considered using wasi, but it simply isn't there yet. witx-bindgen seems usable for `wasm32-unknown-unknown`, and I think it may be suitable to generate a cffi and bindings using a higher level api than what is possible using a c header. any thoughts on if this would be considered upstream? might give it a go next week. (mainly interested in interfacing with dart/flutter and maybe kotlin/swift).

alexcrichton:
Thanks for the report, but can you clarify what you're asking for here? There's already generation modes for C right now (see the `crates/gen-c` generator), but are you thinking that this should support non-wasm language bindings?

dvc94ch:
>  are you thinking that this should support non-wasm language bindings?

exactly, yes. looks like the interface specification isn't specific to wasm.

dvc94ch:
basically the use case I have is we're building a rust p2p networking/crdt library that will be packaged for use in popular frontend frameworks like flutter/react. looks like witx-bindgen could help us keep the different language bindings in sync and avoid some error prone and tedious boilerplate.

alexcrichton:
Ah ok thanks for the clarification. In that case though I'm going to close this as out of scope for this project for now. While one day witx and friends may make cross-language communication easier than just for wasm but also for all embeddings it's an extremely ambitious project and we've got enough to figure out today with just wasm. Perhaps in the future having a tracking issue for this or a demo implementation, but for now there's too much else to do.

dvc94ch:
So we'll be using it for the wasm32-unknown-unknown target and I'll look into forking it and replacing our cffi/dart bindings.

dvc94ch:
so got it working, except for the async feature. the required changes are:

- support for i64 pointers
- fix prefixing/namespace inconsistency
- support passing RET_AREA as an argument
- handle needs some kind of redesign

looks like getting non-wasm support isn't that far off

PR #103, opened 2021-11-09
Native

dvc94ch:
Adds not too hacky support for generating non wasm interfaces as explained in #99 

This will only work on 32bit architectures at the moment, but can be tested on `i686-unknown-linux-gnu`. The push/pull buffers and async functions are not implemented yet and left for future work.

Here is the test setup: https://github.com/dvc94ch/wai-native

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,9 +52,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.44"
+version = "1.0.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "61604a8f862e1d5c3229fdd78f8b02c68dcf73a4c4b05fd636d12240aaa242c1"
+checksum = "ee10e43ae4a853c0a3591d4e2ada1719e553be18199d9da9d4a83f5927c2f5c7"
 
 [[package]]
 name = "async-trait"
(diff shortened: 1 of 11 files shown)


dvc94ch:
Is this ok to merge without all runtime tests working? the smoke test happens to work without proper 64bit support, but to enable all the runtime tests for the rust-native codegen it'll be required. I can either do it in this PR or do a followup. The reason I didn't do 64bit support first was that I couldn't get wasm64-unknown-unknown to work. As you know since you have a PR open to fix the target.

alexcrichton:
Thanks for the PR here but [as I mentioned in the related issue](https://github.com/bytecodealliance/wai-bindgen/issues/99#issuecomment-961357458) this is not something that I want to take on with this project at this time. I think that there's great possibilities with wasm interface types and native FFI between languages and libraries, but we've honestly just got our hands extremely full trying to get WebAssembly sorted out much less "all other FFI languages" as well. I at least am not in a position personally to keep this on the side and/or help maintaining it. If you're interested in pursuing this I'd recommend experimenting in a fork and seeing how it goes, perhaps with the intention of merging back in the future once the wasm interface types story has settled down more.

I'm going to close this for the above reasons. To reiterate I think it makes sense to revisit this in the future but otherwise interface-types is still nascent enough that this is a huge amount more burden to take on while dealing with wasm alone is already enough of a burden.

dvc94ch:
Is 64bit support something that is on the roadmap? I presume you'd want to get that working before completing the "canonical abi"?

alexcrichton:
Yes that's one thing which is still up-in-the-air, the wasm64 target for code is still quite new. Nevertheless that's definitely something we plan on implementing for wasm, yes.

PR #111, opened 2021-12-13
Fix the name of the `mod` in rust-wasm bindings for ".wit.md" files.

sunfishcode:
With ".wit.md" files, we need to strip both extensions, so that we don't
end up with Rust code like `mod foo.wit`.


alexcrichton:
👍

PR #112, opened 2021-12-13
Use `bitflags` in the rust-wasm bindings.

sunfishcode:
Use the `bitflags` crate to describe the flags in a flags type, in the
rust-wasm bindings.


alexcrichton:
👍 Nice!

Issue #115, opened 2021-12-14
Unit Type


theduke:
Wit currently doesn't seem to offer a unit type. (`()` in Rust, `null`/`void`/... in other languages).

It can be represented with an empty tuple `tuple<>`, but a builtin type would make sense to me.

This would be especially useful with `expected<unit, _>` but is generally useful.

I'd be happy to implement this.

sunfishcode:
An `expected` with no value on success can be written as `expected<_, error>`; an `expected` with no value on success or error can be written as `expected<_, _>`.

theduke:
That works indeed.

A unit type would be more general and coherent.

Is there a reason why this was implemented with a special construct instead of a proper type? (`_` can't be used in other places that expect a type)

Maybe because not every language has a direct mapping that makes sense?

sunfishcode:
I think all languages could support it; a bare `unit` argument, for example, could be lowered into any language by just omitting the argument. But in practice, it's not clear what kind of generality that would enable.

theduke:
> But in practice, it's not clear what kind of generality that would enable.

An example: there could (and probably will be) conversion tools that generates definitions based on existing types.

Let's say a Rust derive macro, or a Python attribute.

How would these convert types like 

rust```
struct S {
  x: (),
}
```

or 

```python
@dataclass
class C:
  x: None
```


Each tool would have to chose it's own definition of a unit type.

theduke:
What about: 

```
union U {
  _,
  bool,
  string,
}
```

Here the unit type carries semantic meaning and can't just be omitted.

alexcrichton:
FWIW in general this repository is implementing what's specified by [interface types](https://github.com/webassembly/interface-types), so discussion about changing the design of interface types or similar might be best to happen on the interface types repository rather than here.

theduke:
The newest `component-model` explainer features a `unit` type, so this is relevant again.

PR #116, opened 2021-12-15
Allow _ (unit type) in Unions

theduke:
Allow unions to have variants without a type.

This just requires changing the parser, since unions are just an alias
for variants with indexes as variant names, and variants don't need a type.

Notes:

* it wanted  to check that `_` is only present at most once, but there currently isn't any duplicate type checking for unions anyway so I just left it as is

* There are currently *a lot* of failing UI tests for the parser on main! None of those are affected
* The additional tests types in `types.wit` are at the bottom to keep the diff a bit smaller instead of renumbering

### DIFF
--- a/crates/parser/src/ast.rs
+++ b/crates/parser/src/ast.rs
@@ -594,12 +594,16 @@ impl<'a> TypeDef<'a> {
                 Token::LeftBrace,
                 Token::RightBrace,
                 |docs, tokens| {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Can you expand a bit on the motivation for this? Presumably instead of using a `union` a `variant` could be used instead I think?

theduke:
Superseded by general `unit` type support. See #115

PR #118, opened 2021-12-15
Implement `AsFd` etc. for handle types in the rust-wasm bindings.

sunfishcode:
Implement `AsFd`/`FromFd`/`IntoFd` and `AsRawFd`/`FromRawFd`/`IntoRawFd`
traits for handle types in the rust-wasm bindings.

This makes it more convenient to implement types like `File` and
`TcpStream` as wrappers around handles, as they support the usual
I/O handle APIs, and can simply be constructed directly from `FromFd`
without having to use raw handles and the unsafe `from_raw_fd`.

These traits are part of the [I/O safety RFC], which are
[implemented in Rust nightly]. In order to avoid having wit-bindgen
depend on nightly, this PR uses the [io-lifetimes] crate, which
contains standalone definitions of the main types and traits. Once
I/O safety is stabilized, the code can be switched to use std.

[I/O safety RFC]: https://github.com/rust-lang/rfcs/blob/master/text/3128-io-safety.md
[implemented in Rust nightly]: https://github.com/rust-lang/rust/issues/87074
[io-lifetimes]: https://crates.io/crates/io-lifetimes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,7 +155,7 @@ checksum = "4f8499797f7e264c83334d9fc98b2c9889ebe5839514a14d81769ca09d71fd1d"
 dependencies = [
  "cap-primitives",
  "cap-std",
- "io-lifetimes",
+ "io-lifetimes 0.3.3",
  "rustc_version",
  "winapi",
 ]
@@ -170,7 +170,7 @@ dependencies = [
  "errno",
  "fs-set-times",
  "io-extras",
- "io-lifetimes",
+ "io-lifetimes 0.3.3",
  "ipnet",
  "maybe-owned",
  "rustc_version",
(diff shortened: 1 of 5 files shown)


alexcrichton:
I'm a little hesitant to consider all resources an fd in the sense that fds are typically about I/O things but resources aren't particularly related to that. It should be possible to still implement these traits for generated types though. Perhaps this could be a macro option or otherwise just left to embedder crates?

sunfishcode:
Makes sense. I tend to think of all handles as being for I/O in a sense, from the perspective of a component, but you're right, it's not actually useful to have different types of handles accessible through the same type (`OwnedFd`). I'll move this to the toolchain.

PR #123, opened 2021-12-21
Replace string-literal syntax with a new '@' identifier syntax.

sunfishcode:
Following up on #119, this removes the string-literal syntax for
identifiers, and introduces a new '@'-prefixed syntax.

'@' identifiers are just like regular identifiers, except they start
with '@' and are never parsed as keywords.

Suggestions for better syntax are welcome.


alexcrichton:
Seems reasonable to me! I think it might be best to leave this to the new year though and make sure everyone who's out right now is ok with this as well. Otherwise though r=me

sunfishcode:
It's the new year! What do people think of `@`-prefixed identifiers?

alexcrichton:
Coming back to this (sorry meant to do this earlier!) one thing that now strikes me is that `@` is often reserved for "attributes" in various languages or rather it's often a popular choice for attribute-like things. If we use `@` for identifiers I think that may block off the route for doing that in the future? Not that I know of any current plans to add attribute-like things into `*.wit` but it seems like a possibility at some point in the future though.

sunfishcode:
That's reasonable. What would you think about `%` as the prefix?

alexcrichton:
Yeah I don't expect this to be really used all that much, if ever, so something like `%` seems reasonable to me (not that I've seen it used anywhere else before but hey no problem being the first)

fitzgen:
To be clear, this is just for raw identifiers, not all regular identifiers? If so, SGTM.

radu-matei:
Either `@` or `%` SGTM.
(I've seen `@` used in the past for keywords as identifiers (C#), or enclosing the identifier in double backticks (F#))

sunfishcode:
> To be clear, this is just for raw identifiers, not all regular identifiers? If so, SGTM.

Yes. This is just to have a way to support interfaces with names like `type` which is otherwise a keyword.

bsurmanski:
drive-by unsolicited feedback: 
have you considered $? Many languages, including bash, perl, and Rust macros, use a dollar sign to signify variables

sunfishcode:
I don't have a strong opinion about it, but my vague idea here is that we're more likely to want `$` for something in the future than `%`, and for the immediate purpose here, we just need something to disambiguate keywords from identifiers, so it won't be used very often, so it's ok if we use `%` even if it looks a little unusual.

sunfishcode:
This is now updated to use '%' instead of '@'.

PR #125, opened 2021-12-22
On wit sintax: formalizing documentation comments

### LINKED ISSUE #117: On wit sintax: formalizing documentation comments?
Hi there!

Thank you for the good work!

I read #108 and @sunfishcode mentioned there is interest in feedback on the current [wit sintax](https://github.com/bytecodealliance/wit-bindgen/blob/main/WIT.md).

I wanted to ask about [comment in wit sintax](https://github.com/bytecodealliance/wit-bindgen/blob/main/WIT.md#comments):
Should there be documentation and code comments?

## Current state

While trying out `wit-bindgen`, I noticed that the code generation for Markdown passes comments as documentation, which is really nice! The online demo accepts both `//` and `///` for (documentation) comments when generating the `.md` file.

I really like the current behaviour: taking comments before an item as its documentation. 

As an example of what I would write and use the Markdown generated file as documentation:
``` wit
/// person represents a physical, real person
record person {
  /// Legal name of the person
  name: string,
  /// Year of bird
  year: u32,
  // TODO: add more fields as they are needed
}

// This comment does not appear in the md file generated because there is no element that follows

/// Returns the greeting we shoud use with `person`.
hello: function(who: person) -> string
```

## "Use case"

For now, `.wit` files are expected to be small(?) so this difference is not really needed. But if they get bigger and more complex, this differentiation of comments might come in handy :)

Currently, "code comments" can be achieved by adding a new line after them: no element follows this comment, so they are not passed to the `.md` file. But maybe this behaviour of the `wit-bindgen-gen-markdown` should be included in the wit syntax?

saona-raimundo:
Closes #117 

The behaviour is the same a Rust (doc) comments, I hope.
Nesting in documentation works as follows:
- Whatever inside the outer documentation is taken as raw documentation text
- documentation inside comments are ignored, just as anything else inside comments

There were not many tests in the crate in general, so I was not sure if I had to include some(?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,22 +149,21 @@ dependencies = [
 
 [[package]]
 name = "cap-fs-ext"
-version = "0.21.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4f8499797f7e264c83334d9fc98b2c9889ebe5839514a14d81769ca09d71fd1d"
+checksum = "28c920cce6be66349c19b15f149741ac63a99c62c001eac873011cdea38801e7"
 dependencies = [
  "cap-primitives",
  "cap-std",
  "io-lifetimes",
- "rustc_version",
  "winapi",
 ]
 
 [[package]]
 name = "cap-primitives"
-version = "0.21.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c5998b8b3a49736500aec3c123fa3f6f605a125b41a6df725e6b7c924a612ab4"
+checksum = "a2737eb174c9482c865789a428bfbddfdf284317d466ebbc637eee2e1242bfaa"
 dependencies = [
  "ambient-authority",
  "errno",
@@ -173,7 +172,6 @@ dependencies = [
  "io-lifetimes",
  "ipnet",
  "maybe-owned",
- "rustc_version",
  "rustix",
  "winapi",
  "winapi-util",
@@ -182,9 +180,9 @@ dependencies = [
 
 [[package]]
 name = "cap-rand"
-version = "0.21.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fafda903eb4a85903b106439cf62524275f3ae0609bb9e1ae9da7e7c26d4150c"
+checksum = "d14abeb7657b6740639c7e33534633412643ddc96aa24143e9fe531ab56a23bb"
 dependencies = [
  "ambient-authority",
  "rand",
@@ -192,23 +190,22 @@ dependencies = [
 
 [[package]]
 name = "cap-std"
-version = "0.21.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "811de89a7ede4ba32f2b75fe5c668a534da24942d81c081248a9d2843ebd517d"
+checksum = "7a9ba199282865f3e3c3e7afc6907a3a27896435e24209e035c8a24d4fbcf43a"
 dependencies = [
  "cap-primitives",
  "io-extras",
  "io-lifetimes",
  "ipnet",
- "rustc_version",
  "rustix",
 ]
 
 [[package]]
 name = "cap-time-ext"
-version = "0.21.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85f263d62447efe8829efdf947bbb4824ba2a3e2852b3be1d62f76fc05c326b0"
+checksum = "ecbb6ba1f840b5b22fedaa6956ddf2e1e77604e986e23f7238c8ba509ce0bb9f"
 dependencies = [
  "cap-primitives",
  "once_cell",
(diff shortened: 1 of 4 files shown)


saona-raimundo:
Oh! and one noticeable change:

Doc comments are strictly starting with a whitespace after `///` or `/**`. 

I do not know if this is the wanted behaviour, but I thought it was nice.

saona-raimundo:
The implementation of balanced block comments is super basic and does not handle edge cases where the interleaving of blocks is not correct. To properly handle such cases I would introduce a new type of error, is that okay?

saona-raimundo:
> Doc comments are strictly starting with a whitespace after /// or /**.

After checking with Rust, I reverted this.

@alexcrichton, could you help me with the tests? I do not know how to fix them. I checked out if they were failing before my changes and yes, they were failing before (at least in my local repo).

WIT.md
```diff
@@ -44,16 +44,41 @@ whitespace ::= ' ' | '\n' | '\r' | '\t'
...
-          | '/*' any-unicode-character* '*/'
+implementation-comment ::= '//' character-that-isnt-a-newline*
+                         | '/*' any-unicode-character* '*/'
 ```
 
+#### Documentation comment
+
+A `documentation comment` token in `*.wit` is a documentation line preceded with `///` 
+which ends at the next newline (`\n`) character or it's a block doc comment which starts
+with `/**` and ends with `**/`. Note that block doc comments are allowed to be nested
```
alexcrichton, line 66: To match the Rust style I think the ending here will be `*/` because I believe the intention is that doc comments are subsets of normal comments.

WIT.md
```diff
@@ -44,16 +44,41 @@ whitespace ::= ' ' | '\n' | '\r' | '\t'
...
+just before the item they’re documenting. 
+
+The text considered for the documentation is constructed as follows.
+For line doc comments, all leading `/` are ignored. Then, leading and trailing whitespaces
+in each line (either in lines or blocks) is also ignored. Note that the number of lines is 
+preserved as given in the `*.wit`: different documentation lines in doc comments are different 
+lines in the final documentation.
+
+```wit
+documentation-comment ::= '///' '/'* character-that-isnt-a-newline*
```
alexcrichton, line 77: I don't think the `'/'*` is necessary here, right? (not sure why there'd be allowed even more slashes explicitly)

crates/parser/src/ast/lex.rs
```diff
@@ -29,6 +29,7 @@ pub struct Span {
 pub enum Token {
     Whitespace,
     Comment,
+    DocComment,
```
alexcrichton, line 32: Personally at least I'd prefer to keep the lexer the same as before where doc-comments are simply lex'd as normal comments and it's up to other systems externally to interpret doc-comments and process them.

alexcrichton (commented):
I think for now it'd be best to not change the lexer and only change the documentation. Either that or if you'd like to update processing of doc comments I think it'd be best to do it outside of the lexer in the parser rather than in the lexer itself.

crates/parser/src/ast/lex.rs
saona-raimundo: Okay! 

My first idea would be to change then `Resolver::docs` only, and no changes to lex. But then, I am not sure what should be returned by [parser::Resolver::docs](https://github.com/bytecodealliance/wit-bindgen/pull/125/files#diff-2164780efe09de34f76096a0bcfdefb5b1f4226ce229361f647c9174335a7961L511-R511), as it would return only doc-comments and there would be no access point to normal comments.

> it's up to other systems externally to interpret doc-comments and process them.

Is the `parser::Resolver` considered an "external system"? 

The `parser` crate only defines the `Doc` struct. To make comments available to external crates, should I define a `Comments` struct and define a new method `Resolver::comments`?

WIT.md
saona-raimundo: I wrote this because (currently) immediate `'/'` are trimmed and therefore are not part of the final documentation.

In other words, after `///` any character counts as part of the documentation, except `/`.

The implementation is [this line](https://github.com/bytecodealliance/wit-bindgen/pull/125/files#diff-2164780efe09de34f76096a0bcfdefb5b1f4226ce229361f647c9174335a7961L518-R518)

In Rust, if I am not mistaken, a line starting with `////` is considered a comment and is not part of the documentation. Should I do this?

WIT.md
saona-raimundo: True! My bad, I usually do not use block comments. I will change depending on the rest of the review.

crates/parser/src/ast/lex.rs
alexcrichton: Yeah the original intention was to discard normal comments and only pass through doc-comments, I unfortunately just never got around to it.

alexcrichton: Also, yes, by "external system" I basically mean external to this module, anything in `Resolver` is fine.

WIT.md
alexcrichton: I think that's just a weird artifact of the current implementation but I don't think there's any need to codify it, this just needs to say that doc-comments start with three slashes.

saona-raimundo:
Sorry for closing this PR, I had some problems with git branches. I opened a follow-up.

PR #126, opened 2021-12-27
Wasmtime import! demo

saona-raimundo:
This is an example with a basic usage of `import!` from the `wit-bindgen-wasmtime` crate.

Asked in #59 by @nitishm.

@nitishm, is this what you had in mind?

I hope I have the README correct in the `wit-bindgen-wasmtime` crate.


nitishm:
Lgtm

alexcrichton:
Thanks for this! I'm a bit hesitant to merge this in though for a few reasons which I think unfortunately are somewhat nontrivial:

* I don't really know how to best structure examples in this repository. Working with the example is a multi-step process due to needing to compile the wasm separately from the host. I don't know the best way to organize, orchestrate, and document that.
* Overall the idioms with the Wasmtime bindings aren't really the nicest, they're quite wonky in a few areas (e.g `Store::new(&engine, (wasi, renderer::RendererData {}));` feels pretty odd). This isn't your fault to be clear, it's just that not much effort has gone into making these usable and ergonomic just yet.
* In terms of structure I think an important part to consider is what happens with other languages. I don't know how we'd easily add JS/etc hosts and/or more guests such as C or JS either. I'd like to ideally have an organization that's consistent and allows everything to "work with everything" if that makes sense, but I don't know how to organize that off the top of my head.

On a slightly more mundane and easier-to-handle note I do think that we'll want examples at least built on CI if not executed as well to ensure that they're kept up to date and working.

saona-raimundo:
> I'm a bit hesitant to merge this in though for a few reasons which I think unfortunately are somewhat nontrivial:

No problem!

> Overall the idioms with the Wasmtime bindings aren't really the nicest, they're quite wonky in a few areas (e.g Store::new(&engine, (wasi, renderer::RendererData {})); feels pretty odd). This isn't your fault to be clear, it's just that not much effort has gone into making these usable and ergonomic just yet.

Should I start something in [wasmtime](https://github.com/bytecodealliance/wasmtime)? (you may know if it would be a timely request or not)

> In terms of structure I think an important part to consider is what happens with other languages. I don't know how we'd easily add JS/etc hosts and/or more guests such as C or JS either. I'd like to ideally have an organization that's consistent and allows everything to "work with everything" if that makes sense, but I don't know how to organize that off the top of my head.

Are you thinking of an orchestrating tool like [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) or [trunk](https://trunkrs.dev/), but for wasm applications that would run on `wasmtime` runtime (instead of the browser) and can be written in any language targeting Wasm?

> On a slightly more mundane and easier-to-handle note I do think that we'll want examples at least built on CI if not executed as well to ensure that they're kept up to date and working.

I will look into adding this as a test then :)

PR #128, opened 2022-01-07
On wit sintax: formalizing documentation comments (second attempt)

saona-raimundo:
Follows up #125, from a fresh start, taking into account all previous comments.

### DIFF
--- a/crates/parser/src/ast/resolve.rs
+++ b/crates/parser/src/ast/resolve.rs
@@ -514,7 +514,7 @@ impl Resolver {
         }
         let mut docs = String::new();
         for doc in doc.docs.iter() {
-        	// Comments which are not doc-comments are silently ignored
+            // Comments which are not doc-comments are silently ignored
             if let Some(doc) = doc.strip_prefix("///") {
                 docs.push_str(doc.trim_start_matches('/').trim());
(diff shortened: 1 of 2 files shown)


crates/parser/src/ast/resolve.rs
```diff
@@ -514,13 +514,13 @@ impl Resolver {
         }
         let mut docs = String::new();
         for doc in doc.docs.iter() {
-            if let Some(doc) = doc.strip_prefix("//") {
+            if let Some(doc) = doc.strip_prefix("///") {
                 docs.push_str(doc.trim_start_matches('/').trim());
                 docs.push('\n');
             } else {
-                assert!(doc.starts_with("/*"));
+                assert!(doc.starts_with("/**"));
```
alexcrichton, line 521: I think this assert will need to be conditional instead of an actual assert. Previously this could be an assert because comments only came in these two styles.

saona-raimundo: Wow, I am really sorry for this. It should be fine now.

I wanted to add tests... but I did not see unit tests. It would be weird to add some just for this, but let me know if I should add some.

alexcrichton:
👍

PR #138, opened 2022-01-26
Fix typed array canonical list lowering in JS bindgen

ospencer:
Hey there!

I was messing around with the JS output and noticed an issue with the way memory is copied when lowering lists of bytes. The argument is a typed array, but it copies the entire underlying buffer, which in some cases can be larger than the view of the typed array. This results in an error since we end up trying to write more data than what's expected. 

I fixed this by explicitly providing the byte offset and size.

Let me know if there's any other information I can provide or anything I can do to help get this merged.

Cheers!

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -1734,7 +1734,7 @@ impl Bindgen for FunctionBindgen<'_> {
                         ));
                         // TODO: this is the wrong endianness
                         self.src.js(&format!(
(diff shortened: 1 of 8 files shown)


alexcrichton:
Thanks! Would it be possible to add a test for this as well?

ospencer:
@alexcrichton I added a `list-roundtrip` to make it a bit easier to test this, and confirmed the test fails with the current implementation and passes with the fix. Let me know if the test is clear enough!

alexcrichton:
Looks great to me, thanks! I'll merge when CI is green

PR #139, opened 2022-01-26
Implement `Clone` for imported resources

### LINKED ISSUE #136: rust-wasm: Imported resources should implement `Clone`
The `resource_clone_*` functions aren't currently exposed to Rust for imported resources. I think they should be exposed as a `Clone` implementation, like they already are on `Handle` for exported resources.

Liamolucko:
Fixes #136


alexcrichton:
Thanks!

PR #142, opened 2022-01-27
Fix returned strings and lists in async wasmtime code

### LINKED ISSUE #141: wasmtime: list<u8> types do not work in async mode
I'm using the following:
```rust
wit_bindgen_wasmtime::import!({
    paths: ["./wit/my_module.wit"],
    async: *,
});
```
And `wit` contains something along the lines of:
```
record response {
    body: string,
}
handle_request: function() -> response
```

This fails in runtime with the following:
```
thread 'tokio-runtime-worker' panicked at 'must use `call_async` with async stores', /work/target/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmtime-0.33
.0/src/func/typed.rs:75:9
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/std/src/panicking.rs:525:12
   1: wasmtime::func::typed::TypedFunc<Params,Results>::call
             at ./target/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmtime-0.33.0/src/func/typed.rs:75:9
   2: wit_bindgen_wasmtime::rt::copy_slice
             at ./target/.cargo/git/checkouts/wit-bindgen-f98ccff846621481/8fdf702/crates/wasmtime/src/lib.rs:154:9
   3: edgedb_wasm_server::abi::http_server_v1::edgedb_http_server_v1::EdgedbHttpServerV1<T>::handle_request::{{closure}}
             at ./server/src/abi/http_server_v1.rs:1:1
```

The issue is, `copy_slice` function which also accepts destructor:
```rust
    pub fn copy_slice<T: Endian>(
        store: impl AsContextMut,
        memory: &Memory,
        free: &TypedFunc<(i32, i32, i32), ()>,
        base: i32,
        len: i32,
        align: i32,
    ) -> Result<Vec<T>, Trap> {
        let size = (len as u32)
            .checked_mul(mem::size_of::<T>() as u32)
            .ok_or_else(|| Trap::new("array too large to fit in wasm memory"))?;
        let slice = memory
            .data(&store)
            .get(base as usize..)
            .and_then(|s| s.get(..size as usize))
            .ok_or_else(|| Trap::new("out of bounds read"))?;
        let result = Le::from_slice(slice).iter().map(|s| s.get()).collect();
        free.call(store, (base, size as i32, align))?;   // <<< this line should be async!
        Ok(result)
    }
```

Seems easy enough to fix. What fix do you prefer, having `copy_slice_async` function? Or just moving the `free` call out of the function into the generated code? (or maybe a boolean flag argument?).

tailhook:
Fixes #141

Not sure how to add a test for this as `cargo test` does not run anything. Other than that, it works on my code base.

### DIFF
--- a/crates/gen-wasmtime/src/lib.rs
+++ b/crates/gen-wasmtime/src/lib.rs
@@ -1874,34 +1874,42 @@ impl Bindgen for FunctionBindgen<'_> {
                     self.gen.needs_copy_slice = true;
                     self.needs_functions
                         .insert(free.to_string(), NeededFunction::Free);
-                    let (stringify, align) = match element {
-                        Type::Char => (true, 1),
-                        _ => (false, self.gen.sizes.align(element)),
+                    let (stringify, align, el_size) = match element {
+                        Type::Char => (true, 1, 1),
+                        _ => (
+                            false,
+                            self.sizes().align(element),
+                            self.sizes().size(element),
+                        ),
                     };
                     let tmp = self.tmp();
                     self.push_str(&format!("let ptr{} = {};\n", tmp, operands[0]));
                     self.push_str(&format!("let len{} = {};\n", tmp, operands[1]));
-                    let result = format!(
+                    self.push_str(&format!(
                         "
-                                copy_slice(
-                                    &mut caller,
-                                    memory,
-                                    func_{},
-                                    ptr{tmp}, len{tmp}, {}
-                                )?
-                            ",
-                        free,
+                            let data{tmp} = copy_slice(
+                                &mut caller,
+                                memory,
+                                ptr{tmp}, len{tmp}, {}
+                            )?;
+                        ",
                         align,
-                        tmp = tmp
+                        tmp = tmp,
+                    ));
+                    self.call_intrinsic(
+                        free,
+                        // we use normal multiplication here as copy_slice has
+                        // already verified that multiplied size fits i32
+                        format!("(ptr{tmp}, len{tmp} * {}, {})", el_size, align, tmp = tmp),
                     );
                     if stringify {
                         results.push(format!(
-                            "String::from_utf8({})
+                            "String::from_utf8(data{})
                                     .map_err(|_| wasmtime::Trap::new(\"invalid utf-8\"))?",
-                            result
+                            tmp,
                         ));
                     } else {
-                        results.push(result);
+                        results.push(format!("data{}", tmp));
                     }
                 }
                 None => {
--- a/crates/wasmtime/src/lib.rs
+++ b/crates/wasmtime/src/lib.rs
@@ -137,7 +137,6 @@ pub mod rt {
     pub fn copy_slice<T: Endian>(
         store: impl AsContextMut,
         memory: &Memory,
-        free: &TypedFunc<(i32, i32, i32), ()>,
         base: i32,
         len: i32,
         align: i32,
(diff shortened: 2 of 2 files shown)


crates/gen-wasmtime/src/lib.rs
```diff
@@ -1881,27 +1881,29 @@ impl Bindgen for FunctionBindgen<'_> {
...
+                                ptr{tmp}, len{tmp}, {}
+                            )?;
+                        ",
                         align,
-                        tmp = tmp
+                        tmp = tmp,
+                    ));
+                    self.call_intrinsic(
+                        free,
+                        format!("(ptr{tmp}, len{tmp} as i32, {})", align, tmp = tmp),
```
alexcrichton, line 1897: Ah I remember now why I bundled the free in copy_slice, the `len` here is incorrect since it needs to be the byte size instead of the item-size. I think that should be reasonably easy to fix though with a multiplication here (we know it shouldn't overflow else the `copy_slice` would have failed)

alexcrichton:
Ah yeah unfortunately I haven't had a chance to add a suite of many async tests here yet. I hope to do that eventually but for now if local testing shows this is ok that's fine.

tailhook:
Updated. I'm not sure if `size_of_val` is the best idea here, but I didn't find anything better.

Tested with non-u8 types, although previously was tested too (I guess len is not used by memory allocators most of the time).

alexcrichton:
Oh I think you should be able to use the internal `SizeAlign` structure to get the size of the type to dealloate in the code generator rather than using the Rust `size_of_val` function

alexcrichton:
Actually, as I think about this, I realize now that there is already a test for async things in the test suite. That's actually a test for a different kind of async (async wit functions), but I think it'd be reasonable to add a new `tests/runtime/xxx` folder for a `host.rs` where the wasmtime host is async, would you be up for adding that?

tailhook:
I kinda lost what's wrong here. Also I find it weird that:
1. Approach with presumably invalid length run test fine: https://github.com/bytecodealliance/wit-bindgen/commit/9e9edff
2. Approach with `size_of_val` run test fine: https://github.com/bytecodealliance/wit-bindgen/commit/4ba42440e194d0c301b0bb956c2eb2df609a7de0
3. Then last appoach with using `self.sizes().size` failed the tests: https://github.com/bytecodealliance/wit-bindgen/commit/ec9edbfeff9dd2094d95b309fa34ed680aaf00da

So number of bytes freed is checked somewhere in tests (evidence is (3)) but it did not fail (1). I also don't know how to fix the latest issue.



> Actually, as I think about this, I realize now that there is already a test for async things in the test suite. That's actually a test for a different kind of async (async wit functions), but I think it'd be reasonable to add a new `tests/runtime/xxx` folder for a `host.rs` where the wasmtime host is async, would you be up for adding that?

I've spent few hours to figure out how to add another test variant. And I still need some time to figure out how to make async tests. So let's make it a follow up PR a bit later.

alexcrichton:
Ah I believe the issue is that `list<char>` is interpreted as `string` which loads bytes but the size of each element, `char`, is 4 bytes which means the size-to-free is incorrect. After fixing that locally the tests all pass.

No need to spend too much time working on a test for this though, if it takes too long I can add one later.

tailhook:
> Ah I believe the issue is that `list<char>` is interpreted as `string` which loads bytes but the size of each element, `char`, is 4 bytes which means the size-to-free is incorrect. After fixing that locally the tests all pass.

Ah, right. This makes sense. I've seen that hack for alignment and thought this is some tricky alignment quirk. Fixed.

alexcrichton:
Ok I'll go ahead and merge this and I can figure out the testing story later.

PR #150, opened 2022-02-07
feat(TS): make nullable fields use option `?`

willemneal:
Now fields and type aliases that are nullable are encoded as `T?` instead of `T | null`.

For fields (properties), this means `T | undefined` and for type aliases `T | null` as before.

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -174,7 +174,7 @@ impl Js {
                     TypeDefKind::Record(_) => panic!("anonymous record"),
                     TypeDefKind::Variant(v) if v.is_bool() => self.src.ts("boolean"),
                     TypeDefKind::Variant(v) => {
-                        if iface.is_nullable_option(v) {
+                        if self.is_nullable_option(iface, v) {
                             self.print_ty(iface, v.cases[1].ty.as_ref().unwrap());
                             self.src.ts(" | null");
                         } else if let Some(t) = v.as_option() {
@@ -355,6 +355,17 @@ impl Js {
         self.intrinsics.insert(i, i.name().to_string());
         return i.name().to_string();
     }
+
+    pub fn get_nullable_option<'a>(&self, iface: &'a Interface, ty: &Type) -> Option<&'a Type> {
+        iface.get_variant(ty).and_then(|v| v.as_option())
+    }
+
+    pub fn is_nullable_option(&self, iface: &Interface, variant: &Variant) -> bool {
+        variant.as_option().map_or(false, |ty| {
+            self.get_nullable_option(iface, ty)
+                .map_or(true, |ty| self.get_nullable_option(iface, ty).is_none())
+        })
+    }
 }
 
 impl Generator for Js {
@@ -414,8 +425,8 @@ impl Generator for Js {
                 .ts(&format!("export interface {} {{\n", name.to_camel_case()));
             for field in record.fields.iter() {
                 self.docs(&field.docs);
-                let (option_str, ty) = iface
-                    .get_nullable_option(&field.ty)
+                let (option_str, ty) = self
+                    .get_nullable_option(iface, &field.ty)
                     .map_or(("", &field.ty), |ty| ("?", ty));
                 self.src
                     .ts(&format!("{}{}: ", field.name.to_mixed_case(), option_str));
@@ -440,7 +451,7 @@ impl Generator for Js {
                 "export type {} = boolean;\n",
                 name.to_camel_case(),
             ));
-        } else if iface.is_nullable_option(variant) {
+        } else if self.is_nullable_option(iface, variant) {
             self.src
                 .ts(&format!("export type {} = ", name.to_camel_case()));
             self.print_ty(iface, variant.cases[1].ty.as_ref().unwrap());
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! I was gonna comment that I think it'd be good to have some tests for this as well, but it looks like the test updates to get CI green may suffice for that.

willemneal:
@alexcrichton I fixed the tests.  I apologize in advance for the code that unwraps a nullable type.  The update also fixed a bug in one of the tests when trying to access a method on a potentially undefined object.

crates/gen-js/src/lib.rs
```diff
@@ -2470,3 +2488,18 @@ impl Source {
         self.ts.push_str(s);
     }
 }
+
+fn is_nullable(iface: &Interface, ty: &Type) -> bool {
```
alexcrichton, line 2492: Could this perhaps return `Option<&Type>` so the match here doesn't have to be redone above?

alexcrichton: Also, would it be possible to remove this method and use `self.is_nullable_option` instead? (or some sort of shared helper method)

crates/gen-js/src/lib.rs
```diff
@@ -189,7 +189,7 @@ impl Js {
                     TypeDefKind::Variant(v) => {
                         if self.is_nullable_option(iface, v) {
                             self.print_ty(iface, v.cases[1].ty.as_ref().unwrap());
-                            self.src.ts(" | null");
+                            self.src.ts(" | null")
```
alexcrichton, line 192: I think this is a stray change now?

willemneal: No you still need it for cases like function signatures.  While you can have:
```ts
function foo(x?: string);
```

It has the requirement that the optional param comes after any non-optional ones, which isn't the case generally in wit, e.g. would fail

```ts
function foo(x?: string, y: number); 
function foo(x: string | null, y: number);  // <--- forces caller to put null
```

Perhaps if wit makes the same requirement about option types we could remove this.

alexcrichton: I mean moreso the drop of the semicolon

crates/gen-js/src/lib.rs
willemneal: Before I blindly copied and didn't read too much into the details, but now I'm a little confused.

```rust
match variant.as_option() {
        Some(ty) => match ty {
              Type::Id(id) => match &iface.types[*id].kind {
                  TypeDefKind::Variant(v) => !self.is_nullable_option(iface, v),
                  ty => true,
              },
              _ => true,
          },
          None => false,
      }
```

Why is this not just `variant.as_option().is_some()`?  Wouldn't anything with `option<T>` be a nullable option, regardless of `T`?

willemneal: Nvm I see.  `option<option<T>>` becomes `Option<T | null>` in TS.

crates/gen-js/src/lib.rs
```diff
@@ -140,6 +140,16 @@ impl Js {
         }
     }
 
+    fn unwrap_nullable_option<'a>(&self, iface: &'a Interface, ty: &Type) -> Option<&'a Type> {
```
alexcrichton, line 143: Instead of duplicating the logic fo `is_nullable_option` could there be only one method here? (probably this one) and callers of `is_nullable_option` switch to this method with a `.is_some()` or something like that?

Also I think this may be better named something like `get_nullable_option` since "unwrap" typically has panicking connotations in Rust.

willemneal: I moved the methods into the Interface struct, which opens the door for more refactoring in a later PR.

crates/parser/src/lib.rs
```diff
@@ -465,6 +465,27 @@ impl Interface {
...
+    pub fn get_nullable_option(&self, ty: &Type) -> Option<&Type> {
+        self.get_variant(ty).and_then(|v| v.as_option())
+    }
+
+    pub fn is_nullable_option(&self, variant: &Variant) -> bool {
+        variant.as_option().map_or(false, |ty| {
+            self.get_nullable_option(ty)
+                .map_or(true, |ty| self.get_nullable_option(ty).is_none())
+        })
+    }
```
alexcrichton, line 488: I don't think that this should live in the `wit-parser` crate because this is a JS-specific operation. Options are represented differently in each language and each one may need its own special casing.

willemneal: While I agree that different languages will represent them differently, this allows any language to check if the type is `option<T>` vs `option<option<T>>`, which isn't about representation but rather a fact about the variant.

alexcrichton: Perhaps, yeah, but I would personally prefer that this remain language-specific for now and in the future if there's duplication between languages we can deduplicate at that time.

syrusakbary:
This is great!

Would this PR make the following possible? (like in Typescript):

```
record person {
  name: string,
  age?: u32,
}
```

willemneal:
@syrusakbary  Yep!
```
record person {
  name: string,
  age: option<u32>
}
```
becomes
```ts
interface person {
  name: string,
  age?: number
}

alexcrichton:
Thanks!

syrusakbary:
Niice! Love it!

PR #151, opened 2022-02-07
fix(TS): use `/**` doc string in TS 

willemneal:
This way the comments will be proper TS doc strings and show up in hover overs, etc.

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -273,9 +273,9 @@ impl Js {
         };
         self.src.ts("/**\n");
         for line in docs.lines() {
-            self.src.ts(&format!("* {}\n", line));
+            self.src.ts(&format!(" * {}\n", line));
         }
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think this may need some custom handling to handle alignment in the source which might be good to implement, right now I suspect it generates:

```
/*
*
*
*/
```

where I suspect what we should instead generate is:

```
/*
 *
 *
 */
```

Otherwise seems reasonable though!

alexcrichton:
To confirm, does this look ok in generated code to you? I forget if appended code is automatically trimmed or not...

willemneal:
@alexcrichton It's trimmed.  Since it's a generated file and the comment does still show up, it's probably not a big deal.

Also in my other project I'm looking at using some JS prettier on the output.  So I think we could save properly formatted code for later.

alexcrichton:
Ok seems reasonable to go with this and we can fix formatting later if necessary.

PR #152, opened 2022-02-07
fix: add missing keywords to `WIT.md`

willemneal:


alexcrichton (commented):
Thanks for this! Technically though these types are on their way out and generally not really well supported today anyway. I think I'd prefer to hold off on documenting this while the state of them is still in flux, and ideally they'll just get naturally jettisoned soon-ish.

willemneal:
Works for me.  Just thought they should be highlighted as a keyword only (didn't provide any other context), just for signaling not to use those identifiers.

PR #166, opened 2022-03-08
Implement wit2wasm and wasm2wit tools.

peterhuene:
This PR implements simple CLI tools for encoding a single wit file to a
WebAssembly component (`wit2wasm`) or decoding a WebAssembly component to a wit
file (`wasm2wit`).

The latter tool right now only works with the output of `wit2wasm` and may
eventually be made more generic to spit out a wit representation for
all imported/exported interfaces of a WebAssembly component.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,20 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
-dependencies = [
- "gimli 0.25.0",
-]
-
-[[package]]
-name = "addr2line"
 version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
 dependencies = [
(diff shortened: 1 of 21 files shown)


alexcrichton (commented):
Nice! Could you add some tests as well for things like input/output to see the expected shapes of modules/components going in and coming out? I'm not sure if this requires `wast`-parsing of components though now that I think about it. II haven't done a meticulous review of `decoding.rs` and `encoding.rs` yet, but given the support in `src/lib.rs` I was surprised that encoding doesn't take a wasm module of some form or similarly how this works where `*.wit` right now only desribes either imports or exports but not both. I figured though that tests with example usages would probably clear all this up for me.

peterhuene:
This isn't the full `wit-component` tool that I'm working on to take a core wasm module along with imported/exported interface definitions to output a component; that does make use a bit of this code and I'm hoping to have that ready later this week.

Right now these are just tools for encoding a single interface as a wasm component and doing the reverse (the printing of the wit source in `wasm2wit` isn't as useful as getting the `wit_parser::Interface` to generate bindings via a macro or `wit-bindgen`, though, but it was simple enough to implement).

Encoding a wit file (especially one that imports other local wit files) as a single binary artifact to publish and one we'll be able to generate bindings from will be useful for certain future scenarios, though.

I'll add some round tripping tests for these tools that uses the applicable test wit files, encodes them to wasm, and then decodes them back to a matching wit AST.

alexcrichton (approved):
Ok seems reasonable to me!

Issue #56, opened 2021-07-27
Simplify bindings implementation


syrusakbary:
After taking a quick look on the project, I realized there was a bit of shared logic between the binding generators that could be simplified.
Eg: https://github.com/bytecodealliance/witx-bindgen/blob/main/crates/gen-wasmtime/src/lib.rs / https://github.com/bytecodealliance/witx-bindgen/blob/main/crates/gen-wasmtime-py/src/lib.rs

As more bindings get implemented this might become more challenging to maintain. I was wondering if there can be an implementation based on the visitor pattern.

## Visitor pattern

GraphQL designed a very simple pattern (visitor pattern) to walk through the AST in a simple an succinct way.  
Some overview: https://www.graphql-code-generator.com/docs/custom-codegen/using-visitor
Reference implementation in Python: https://github.com/graphql-python/graphql-core/blob/main/src/graphql/language/visitor.py

Example of schema printer based on the Visitor pattern: https://github.com/graphql-python/graphql-core/blob/main/src/graphql/utilities/print_schema.py

## Candid's approach
Candid (Dfinity's version of Interface Types), also follows a similar pattern where it's bindings are simple and concise to define:
https://github.com/dfinity/candid/blob/master/rust/candid/src/bindings/typescript.rs

I was wondering if it could make sense to refactor some of the codebase to allow easy maintainability of the generators while repeating as minimum as possible in the bindings generator code.

Thoughts?

alexcrichton:
Thanks for the report, but I'm not sure that this would really simplify the bindings all that much. Not to say I've written the simplest possible bindings as-is right now, I'm sure they could be improved! In any case I'm happy to review PRs to update the implementations (a prototype in one generator should be fine).

syrusakbary:
Looking a bit more into it, I think it might be useful to add another layer before the binding generation that processes the data to be easily consumed by the generator.

For example, right now this code:
```witx
record person {
  name: string,
  age: u32,
}
hello: function(who: person) -> string
hello2: function(who: person) -> string
```

Generates the following in Javascript:

```js
  imports["input"]["hello"] = function(arg0, arg1, arg2, arg3) {
    const memory = get_export("memory");
    const realloc = get_export("canonical_abi_realloc");
    const ptr0 = arg0;
    const len0 = arg1;
    const ret = obj.hello({
      name: UTF8_DECODER.decode(new Uint8Array(memory.buffer, ptr0, len0)),
      age: arg2 >>> 0,
[...]

PR #173, opened 2022-03-16
Implemented Wasmer Bindings

syrusakbary:
Today members from our community were asking for the latest up-to-date Wasmer wit-bindings, so I think it now makes sense to try to bring this effort upstream!

This PR implements both:
* Wasmer-Rust bindings
* Wasmer-Python bindings

Along with the end to end tests passing for both binding implementations.
Please let us know if there is anything that we need to do to iterate on the codebase. We'll be more than happy to iterate on them!

Thanks, and keep up the good work


tschneidereit:
Thank you for opening this PR. However, nothing's really changed in a fundamental way since Alex's comment in https://github.com/bytecodealliance/wit-bindgen/issues/56#issuecomment-888367561: wit-bindgen itself will continue undergoing churn for the foreseeable future, and we're not able to bear the inevitable cost that having additional 3rd-party bindings in the repository would bring.

To manage expectations for the future though, while we haven't decided on any kind of formal policy, I'm not sure we'll ever want to maintain an open-ended list of 3rd-party bindings in this repository. wit-bindgen is a project with specific stakeholders working on it with specific goals in mind, and we need to ensure that we stay focused on these. We'll base decisions around which bindings to host here strictly on whether we see it as beneficial to wit-bindgen as a project and its stakeholders, on a case-by-case basis.

Since we do appreciate that it's attractive to make use of the work done here without having to invest the considerable effort to do and maintain a full separate implementation, we structured wit-bindgen in a way that makes it easy to maintain bindings externally and make use of wit-bindgen through an API. While that API will necessarily undergo churn for a while, too, we aim to eventually have that settle down and follow semver versioning. I'd encourage you to investigate making use of this for your bindings.

Given the above, I'll close this PR.

PR #174, opened 2022-03-17
Per-language support summaries

Kylebrown9:
Some WIP docs that make it easier for users to tell what's supported the language(s) they're interested in.
Looking for feedback on the general idea of these docs before I get too much further in writing them up.

[Overview](https://github.com/Kylebrown9/wit-bindgen/blob/lang-support-docs/docs/OVERVIEW.md)

Goals:
1. Make language used to describe features (host bindings, guest bindings) more consistent and define them.
2. Make it easier for users to get a general overview of which languages are supported
3. Describe the support for each language independently of each other (e.g. go users shouldn't have to understand how Python is supported to get how go is supported)
4. Provide more prose, examples, and links for each language

TODO:
- [ ] Write more prose for go
- [ ] Write more prose for C++

### DIFF
--- a/docs/OVERVIEW.md
+++ b/docs/OVERVIEW.md
@@ -1,0 +1,27 @@
+# Overview
+WIT Bindgen
+
+## Host Bindings Support
+Host bindings in a language allows code in that language to talk to WASM modules using bindings generated from a WIT interface. This requires that that language has the ability to either embed a WASM runtime like `wasmtime` or interact with one built in to the language runtime (like V8).
+
+## Guest Bindings Support
+Guest bindings in a language allow WASM modules written in that language to implement interfaces defined by WIT interfaces. This requires that the language can be compiled to WASM.
+
+# WIT Support Table
+<!-- Whenever you update the status table, -->
+<!-- make sure the corresponding document is also updated. -->
+
+| Language | Host Bindings | Guest Bindings |
+| - | - | - |
+| Rust | ✔️ ([details](./langs/RUST.md#host-bindings))| ✔️ ([details](./langs/RUST.md#guest-bindings))|
+| JS | ✔️ ([details](./langs/JS.md)) | ❌ |
+| Python | ✔️ ([details](./langs/PYTHON.md)) | ❌ |
+| Go | ⏳ ([details](./langs/GO.md)) | ❌ |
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks for the PR! This is somewhat similar to https://github.com/bytecodealliance/wit-bindgen/pull/126 in the sense that I don't really know the best way to document everything in this repository right now. I'm a bit wary of having too many language-specific docs in lots of places as it can get sort of difficult to update everything and/or remember what needs updating. Things are honestly still pretty early for wit-bindgen and I expect a fair amount will change as the component module and/or interface types matures and more concrete use cases for that continue to arise. 

Basically I don't know how to balance things. This repository hasn't seen many major updates in awhile but that's because lots of work on other parts of the component model has been going on in the background (a spec, async, implementation in Wasmtime, etc). This repository sort of represents a snapshot of a "this is one trajectory we could go in" where lots still needs to be filled out but at the same time a lot of other pieces also work pretty well. I'm a bit hesistant to lock in too quickly to one trajectory when it's likely that things will alter course over time.

Kylebrown9:
In that case, I'm happy to leave this here for now and revisit it when the direction becomes more clear.

alexcrichton:
I think this is relatively outdated at this point so I'm going to close this. Lots has changed in the interim about where things live and such. I have at least a rewrite of the README for this repo  to push up once https://github.com/bytecodealliance/wit-bindgen/pull/463 lands which should cover some of the basics intended here.

That being said things are starting to settle nowadays and the component model is a bit more final then it was last year. The `wit-bindgen` repo here now hosts only guest generators which I think makes more sense to fill out documentation for perhaps. I'm still not certain that this is the final state of things but I'd hope that new-ish documentation wouldn't bitrot so quickly at this point.

Kylebrown9:
I agree with closing this PR, but I do still want to carry on the spirit of what I wrote here somewhere. I think people want to know how and to what extent it's possible to use component model tooling from the guest and host side in each applicable language and that having somewhere central that provides that is useful.

`wit-bindgen` is no longer that central place and I think even the guest generators may split apart eventually to be incorporated into languages tooling, and what we really need is some other place in the Bytecode Alliance that documents this information.

alexcrichton:
Yep I definitely agree, although I'm not sure where such a good central location would be at this point.

PR #181, opened 2022-03-24
chore: [wasmlink] Update wasmencoder to v0.10.0

saulecabrera:
@peterhuene as discussed via Zulip, this is part 1/n towards porting my changes to temporarily support dynamic linking modules, until the component model changes land completely. I'm aiming to make incremental patches rather than a big one to make it easier to review. 

This update will make it easier to keep track of the index space of each
section while constructing modules by relying on the provided `len`
method of each section rather than having to maintain internal data
structure for this purpose.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1603,15 +1603,6 @@ dependencies = [
 
 [[package]]
 name = "wasm-encoder"
-version = "0.4.1"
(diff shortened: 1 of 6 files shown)


peterhuene (approved):
👍 Looks good!

Issue #191, opened 2022-04-09
wasmtime: Allow Import of External Definitions


theduke:
Unless I'm missing something, it currently does not seem possible for one `wit_bindgen_wasmtime::export/import!()` macro to use definitions produced by another `export!/import!` macro.

My use case is allowing crates to use core definitions provided by another crate.

It could look something like:

`outer.wit`:
```wit
use {...} from core

...
```

```rust
wit_bindgen_rust::import!({
  paths: ["outer.wit"],
  external: [ "core" ],
});
```
This would prevent all the `core` definintions from `core.wit` to be generated, but instead expect a `core` Rust module to be in scope.

PR #193, opened 2022-04-12
wit-bindgen-rust: enable standalone code generation.

peterhuene:
This PR enables a "standalone" mode to the Rust bindings generator.

The general idea is that this mode may be used to generate bindings that can be
built as separate crates and depended upon like normal Rust crates.

For both import and export bindings, this means not nesting the bindings inside
of a module.

For export bindings, this means generating a macro that consuming crates can
use to export an interface for a given type. Additionally, this allows users to
export multiple interfaces from the same type (e.g. a single `Component` struct).

An optional `module` field was added to `Interface` to also allow users some
control over the exported and imported symbol names. When specified, it is used
as the module name for import bindings. For export bindings, it is used as part
of a formatted export name of `<module>#<function>`; this format is what is
currently expected of `wit-component` for exported (non-default) interfaces.

### DIFF
--- a/crates/gen-rust-wasm/src/lib.rs
+++ b/crates/gen-rust-wasm/src/lib.rs
@@ -46,12 +46,15 @@ pub struct Opts {
     #[cfg_attr(feature = "structopt", structopt(skip))]
     pub symbol_namespace: String,
 
-    /// The alias to use for the `wit_bindgen_rust` crate.
+    /// If true, the code generation is intended for standalone crates.
     ///
-    /// This allows code generators to alias the `wit_bindgen_rust` crate
-    /// to a re-export in another crate.
+    /// Standalone mode generates bindings without a wrapping module.
+    ///
+    /// For exported interfaces, an `export!` macro is also generated
+    /// that can be used to export an implementation from a different
+    /// crate.
     #[cfg_attr(feature = "structopt", structopt(skip))]
-    pub crate_alias: Option<String>,
+    pub standalone: bool,
 }
 
 #[derive(Default)]
@@ -79,6 +82,10 @@ impl RustWasm {
             Direction::Export => AbiVariant::GuestExport,
             Direction::Import => AbiVariant::GuestImport,
         }
+    }
+
+    fn ret_area_name(iface: &Interface) -> String {
+        format!("{}_RET_AREA", iface.name.to_shouty_snake_case())
     }
 }
 
@@ -150,12 +157,10 @@ impl Generator for RustWasm {
         self.in_import = variant == AbiVariant::GuestImport;
         self.types.analyze(iface);
         self.trait_name = iface.name.to_camel_case();
-        self.src
-            .push_str(&format!("mod {} {{\n", iface.name.to_snake_case()));
 
-        if let Some(alias) = &self.opts.crate_alias {
+        if !self.opts.standalone {
             self.src
-                .push_str(&format!("use {} as wit_bindgen_rust;\n", alias));
+                .push_str(&format!("mod {} {{\n", iface.name.to_snake_case()));
         }
 
         self.sizes.fill(iface);
@@ -456,6 +461,16 @@ impl Generator for RustWasm {
         self.src.push_str(";\n");
     }
 
+    fn preprocess_functions(&mut self, _iface: &Interface, dir: Direction) {
+        if self.opts.standalone && dir == Direction::Export {
+            self.src.push_str(
+                "/// Declares the export of the interface for the given type.\n\
+                 #[macro_export]\n\
+                 macro_rules! export(($t:ident) => {\n",
+            );
+        }
+    }
+
     fn import(&mut self, iface: &Interface, func: &Function) {
         let mut sig = FnSig::default();
         let param_mode = TypeMode::AllBorrowed("'_");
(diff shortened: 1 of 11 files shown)


alexcrichton:
This all seems pretty reasonable to me! I'd prefer though to not maintain multiple modes if we can. Could the new mode be just how the proc-macro works?

peterhuene:
I think I can remove the `standalone` option if we unconditionally use a wrapping `mod` on the generated code and the code generation for the component tooling can simply append a `pub use foo::*` to reexport everything at the crate-level.

The rest of the changes can remain and we'll need to update the examples/tests to use the generated `export` macros:

```rust

wit_bindgen_rust::export!("export.wit");
wit_bindgen_rust::import!("import.wit");

struct Component;

impl export::Export for Component {
    fn say_something() -> String {
        import::say_something()
    }
}

export::export!(Component); // one export macro invocation per exported interface
```

How does that sound?

alexcrichton:
Sounds reasonable to me!

peterhuene:
Argh...I've been trying to get this to play nice with resources and failing (not something necessary for the component model right this minute, so this PR punted on that for now with the "standalone" support); lots of assumptions on `super` in the codegen currently.

My first attempt was "oh I'll just add a type to the interface trait so that the type implementing the trait can specify resource implementation types" and that made some of the resource test cases happy.

But then I got to the handles test case where handles to resource types are inside of the type definitions and not just the exported function signatures; those need to exist outside of the `export!` macro and so there's no way to inform the type definitions of the implementer of the interface trait.

I'm still thinking about how best to accomplish this.

crates/gen-rust-wasm/src/lib.rs
```diff
@@ -941,10 +980,24 @@ impl Bindgen for FunctionBindgen<'_> {
         format!("ptr{}", tmp)
     }
 
-    fn i64_return_pointer_area(&mut self, amt: usize) -> String {
+    fn i64_return_pointer_area(&mut self, iface: &Interface, export: bool, amt: usize) -> String {
         assert!(amt <= self.gen.i64_return_pointer_area_size);
         let tmp = self.tmp();
-        self.push_str(&format!("let ptr{} = RET_AREA.as_mut_ptr() as i32;\n", tmp));
+        // For exports in standalone mode, the `RET_AREA` is accessed via a macro from another
+        // crate, so we need to access it from the expected crate name.
```
Michael-F-Bryan, line 987: I haven't run this to see the generated code so this might be a silly question, but do we need to use a separate `RET_AREA` for each use of the `export!()` macro? 

My interpretation based on the PR's discussion is that the generated `export!()` macro will expand to the code that gets generated by `wit_bindgen_rust::export!()` at the moment, minus the struct and trait definitions. Which means the `RET_AREA` static will be accessible in the current scope and generated functions can refer to it freely.

It might be easier to look at some code:

```rust
// wit_bindgen_rust::export!("export.wit") expands to something like this:

pub mod export {
    pub trait Export {
        fn say_something() -> String;
    }

    /// Given the name of something implementing the [`Export`] trait, generate
    /// glue code that uses the type's functionality to implement the `export.wit`
    /// interface.
    macro_rules! export {
        ($t:ty) => {
            mod export {
                use super::Export; // Assume the "Export" trait was imported in the parent namespace

                #[export_name = "export-0.1.0#say-something"]
                unsafe extern "C" fn __wit_bindgen_export_say_something() -> i32{
                    ...

                    let result0 = <$t as Export>::say_something();

                    ...

                    let ptr2 = RET_AREA.as_mut_ptr() as i32;

                    ...
                    ptr2
                }

                static RET_AREA: [...] = ...;
            }
        }
    }
}
```

Downstream users might then use it like so:

```rust
use common_interface_crate::export::Export;

struct ExportImpl;

impl Export for ExportImpl {
  ...
}

common_interface_crate::export::export!(ExportImpl);
```

Or if we don't want to assume the `Export` trait is available in the parent namespace we could change the `export!()` macro's syntax to something like `export!(ExportImpl as Export)`.

Michael-F-Bryan:
I like this `export::export!(Component)` approach! 

I had been thinking of something similar for work so we can publish a helper crate that wraps WIT file imports and exports, plus some convenience functions, so downstream users just need to add a single crate as a dependency.

> I think I can remove the standalone option if we unconditionally use a wrapping mod on the generated code and the code generation for the component tooling can simply append a `pub use foo::*` to reexport everything at the crate-level.

FWIW, you can do `pub use foo::*` to reexport the generated code from the current implementation. Is that what you were trying to achieve?

peterhuene:
> FWIW, you can do pub use foo::* to reexport the generated code from the current implementation. Is that what you were trying to achieve?

Effectively. I meant that I would put the generated `mod` back in whereas this PR currently removes it for the "standalone" mode because it desires the generated code to be put at the top-level of the generated crate.

By having the mod always present and just have the standalone tooling add an additional `pub use foo::*` to reexport everything at the top-level crate we'd be able to remove the "standalone" mode from the PR.

The problems I'm running into right now relate to what to do about resources as they also reference `super` in their code generation (primarily to use the _actual_ resource type users are implementing).

If we didn't support resources (we don't yet in the component model, but will soon enough), then there would be no issues with removing the "standalone" mode from this PR using the technique described above.

crates/gen-rust-wasm/src/lib.rs
peterhuene: `RET_AREA` currently can't be shared between interfaces as it's calculated to be a maximum space needed between all referencing functions from a single interface (each `export!` invocation can potentially know only about a single interface at a time).

The generated `export!` from these changes could definitely move the `RET_AREA` into the macro so it can just live static next to the functions and inaccessible from the outside world, at least. I can definitely make that change.

alexcrichton:
I think it's an open option to removing resource support for now and re-adding it back in when the component model gets it. We'll probably have to solve the same questions you're running into now but we can hope that the intervening time gave us inspiration.

In general this is starting to age relative to the recent changes in the canonical ABI as well so it's due for a number of updates.

theduke:
Great timing, I just also asked for this in #191 .

My suggestion there was to add an `external = [...]` setting to the proc macro that would disable code generation for the respective wit package, and instead expect a correctly named module to be in scope.

alexcrichton:
Ok that sounds reasonble to me yeah. Would you be ok writing up an issue on removing this option and using it by default so we can track that?

peterhuene:
@alexcrichton opened #243 to track that.

PR #197, opened 2022-04-19
Remove support for `usize` and `char8`

alexcrichton:
Further iteration from #195 and dropping more features that are only
needed for `*.witx`, the `usize` and `char8` types were only added for
backcompat with witx itself.


peterhuene (approved):

sunfishcode (approved):
`char8` was for UTF-8 code units; with interface types, users now use `string` instead.

`usize` was for lengths of buffers; with interface types, users now use `list` and streams instead.

PR #198, opened 2022-04-20
Rename f32/f64 to float32/float64

alexcrichton:
This commit renames the previously-known `f32` and `f64` interface types
to `float32` and `float64` to match the upstream specification.


sunfishcode (approved):
Looks good. The name change also signals the introduction of NaN canonicalization, but we can add that separately.

alexcrichton:
Good point! I'd actually forgotten that so I went ahead and made a checklist of known items at https://github.com/bytecodealliance/wit-bindgen/issues/201 -- let me know if I forgot something though

PR #202, opened 2022-04-20
Add a `Type::String` variant

alexcrichton:
This removes the implicit interpretation that `list<char>` is equivalent
to `string`, enabling `list<char>` to be equivalent to `&[char]` in
Rust, for example. This commit is plumbed throughout all generators with
new pseudo-instructions for lifting/lowering strings.

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -1672,7 +1672,7 @@ impl Bindgen for FunctionBindgen<'_> {
                 }
             }
             Instruction::StringLower { realloc } => {
-                // Lowering only happens when we're passing lists into wasm,
+                // Lowering only happens when we're passing strings into wasm,
                 // which forces us to always allocate, so this should always be
                 // `Some`.
                 let realloc = realloc.unwrap();
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -135,6 +135,7 @@ impl Js {
             Type::Float64 => Some("Float64Array"),
             Type::Char => None,
             Type::Handle(_) => None,
+            Type::String => None,
             Type::Id(id) => match &iface.types[*id].kind {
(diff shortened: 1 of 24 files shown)


crates/gen-js/src/lib.rs
```diff
@@ -1703,6 +1671,49 @@ impl Bindgen for FunctionBindgen<'_> {
                     None => results.push(result),
                 }
             }
+            Instruction::StringLower { realloc } => {
+                // Lowering only happens when we're passing lists into wasm,
```
peterhuene, line 1675:
```suggestion
                // Lowering only happens when we're passing strings into wasm,
```

crates/gen-rust-wasm/src/lib.rs
```diff
@@ -1037,15 +1032,44 @@ impl Bindgen for FunctionBindgen<'_> {
...
+                self.push_str(&format!("let {} = {}.len() as i32;\n", len, val));
+                if realloc.is_some() {
+                    self.push_str(&format!("core::mem::forget({});\n", val));
+                }
+                results.push(ptr);
+                results.push(len);
+            }
+
+            Instruction::StringLift { free, .. } => {
+                // This only happens when we're receiving a list from the
```
peterhuene, line 1059:
```suggestion
                // This only happens when we're receiving a string from the
```

crates/gen-wasmtime-py/src/lib.rs
```diff
@@ -1789,6 +1769,44 @@ impl Bindgen for FunctionBindgen<'_> {
                     None => results.push(result),
                 }
             }
+            Instruction::StringLower { realloc } => {
+                // Lowering only happens when we're passing lists into wasm,
```
peterhuene, line 1773:
```suggestion
                // Lowering only happens when we're passing strings into wasm,
```

peterhuene (approved):
👍 Just some minor comment nits that can be ignored if you want.

PR #204, opened 2022-04-21
Add new "unit" and "bool" primitive types

alexcrichton:
This splits out the `bool` type from `Variant` and additionally adds a
new `unit` type. The new `unit` type will need some more tests in the
future and this is just laying the groundwork for it for now. This'll
soon get automatically tested though as the type of each case in
`Variant` becomes required and functions will also switch to one return
type.


crates/gen-c/src/lib.rs
```diff
@@ -1354,6 +1348,15 @@ impl Bindgen for FunctionBindgen<'_> {
                 }
             }
 
+            Instruction::UnitLower => {}
+            Instruction::UnitLift => {
+                // TODO:
+                // results.push("None".to_string());
```
peterhuene, line 1354: Should we `todo!` this?

peterhuene (approved):

crates/gen-c/src/lib.rs
peterhuene: Never mind, I see that this will be implemented with a placeholder result in #206.

alexcrichton: Indeed yeah I wasn't sure how this was gonna pan out but it ended up getting solve well-ish in the next PR

PR #205, opened 2022-04-21
Fix missing items in C header files

### LINKED ISSUE #194: Returning a result of an option and error does not generate all structs for the C import bindgen
Consider the following interface:

```fsharp
// test.wit
enum error {
    success,
    failure
}

option-test: function() -> expected<option<string>, error>
```

Generating the C import bindings:

```
$ wit-bindgen c --import test.wit
```
Then, trying to compile results in the following error:

```./test.h:22:33: error: unknown type name 'test_option_string_t'
  test_error_t test_option_test(test_option_string_t *ret0);
                                ^
1 error generated.
```

Looking at the generated bindings, `test_option_string_t` is clearly missing:

```c

#ifndef __BINDINGS_TEST_H
#define __BINDINGS_TEST_H
#ifdef __cplusplus
extern "C"
{
  #endif
  
  #include <stdint.h>
  #include <stdbool.h>
  
  typedef struct {
    char *ptr;
    size_t len;
  } test_string_t;
  
  void test_string_set(test_string_t *ret, const char *s);
  void test_string_dup(test_string_t *ret, const char *s);
  void test_string_free(test_string_t *ret);
  typedef uint8_t test_error_t;
  #define TEST_ERROR_SUCCESS 0
  #define TEST_ERROR_FAILURE 1
  test_error_t test_option_test(test_option_string_t *ret0);
  #ifdef __cplusplus
}
#endif
#endif
```

Changing the interface to:

```fsharp
option-test: function() -> expected<option<string>, _>
```

This time, the header file includes the structure definition:

```c
  typedef struct {
    // `true` if `val` is present, `false` otherwise
    bool tag;
    test_string_t val;
  } test_option_string_t;
```

alexcrichton:
This fixes an issue where some structures of "anonymous" types such as
results/etc would accidentally only place types in the C implementation
files instead of the header files where they were needed.

Closes #194

### DIFF
--- a/crates/gen-c/src/lib.rs
+++ b/crates/gen-c/src/lib.rs
@@ -493,7 +493,7 @@ impl C {
             }
 
             TypeDefKind::Variant(v) => {
-                self.src.c("switch (ptr->tag) {\n");
+                self.src.c("switch ((int32_t) ptr->tag) {\n");
                 for (i, case) in v.cases.iter().enumerate() {
                     let case_ty = match &case.ty {
                         Some(ty) => ty,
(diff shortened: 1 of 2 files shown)


peterhuene (approved):

radu-matei (approved):
Thanks a lot for the fix!

PR #206, opened 2022-04-21
Change functions to have one return value

alexcrichton:
This commit changes wit-defined functions to have precisely one return
value instead of a list of return values. This matches the upstream
component-model draft specification and canonical ABI. The main fallout
here is that the spidermonkey tests stopped testing multiple-returns
since while multi-value was implemented struct lifting/lowering is not
yet implemented. Otherwise this is relatively straightforward although
there are a few gotchas here and there as some language like JS, Python,
and C don't have native tuple types and we still want to generate
reasonable-ish code.

### DIFF
--- a/crates/gen-c/src/lib.rs
+++ b/crates/gen-c/src/lib.rs
@@ -69,6 +69,7 @@ struct CSig {
 
 #[derive(Debug)]
 enum Scalar {
+    Void,
     OptionBool(Type),
     ExpectedEnum {
         err: TypeId,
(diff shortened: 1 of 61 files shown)


alexcrichton:
Note that this is stacked on https://github.com/bytecodealliance/wit-bindgen/pull/205 and https://github.com/bytecodealliance/wit-bindgen/pull/204, both of which are prerequisites for this to work.

peterhuene (approved):

PR #207, opened 2022-04-26
Change the ABI for multi-value returns

alexcrichton:
This commit aligns the implementation of wit-bindgen to the current
draft of the canonical ABI to have multi-value returns expressed via
pointers to the in-memory canonical ABI rather than the prior
pointers-to-the-lowered-values-in-memory ABI.

Due to the structure of wit-bindgen this ended up being a relatively
simple change where the lowered instructions generators are using were
already almost all appropriate. Mostly the parser ABI implementation
changed here along with some fiddly bits in `wasmlink`.

### DIFF
--- a/crates/parser/src/abi.rs
+++ b/crates/parser/src/abi.rs
@@ -10,7 +10,10 @@ pub struct WasmSignature {
     pub params: Vec<WasmType>,
     /// The WebAssembly results of this function.
     pub results: Vec<WasmType>,
-    /// TODO
+    /// Whether or not this signature is using a return pointer to store the
+    /// result of the function, which is reflected either in `params` or
+    /// `results` depending on the context this function is used (e.g. an import
+    /// or an export).
     pub retptr: bool,
 }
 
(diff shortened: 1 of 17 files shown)


crates/parser/src/abi.rs
```diff
@@ -10,9 +10,8 @@ pub struct WasmSignature {
     pub params: Vec<WasmType>,
     /// The WebAssembly results of this function.
     pub results: Vec<WasmType>,
-    /// The raw types, if needed, returned through return pointer located in
-    /// `params`.
-    pub retptr: Option<Vec<WasmType>>,
+    /// TODO
```
peterhuene, line 13: I imagine this TODO (and the one below for `return_pointer`) will be short lived, but can we add an explanation just in case?

crates/parser/src/abi.rs
```diff
@@ -1052,23 +1051,34 @@ impl<'a, B: Bindgen> Generator<'a, B> {
...
-                    // through memory. Remove that indirection here by loading
-                    // everything to simulate the function having many return values
-                    // in our stack discipline.
-                    if let Some(actual) = &sig.retptr {
-                        if self.variant == AbiVariant::GuestImport {
-                            assert_eq!(self.return_pointers.len(), 1);
-                            self.stack.push(self.return_pointers.pop().unwrap());
-                        }
-                        self.load_retptr(actual);
+                    // If a return pointer is in use
```
peterhuene, line 1054:
```suggestion
                    // If a return pointer is not in use
```

peterhuene (approved):
Looks good, just minor comments.

PR #208, opened 2022-04-26
Add support for params passed indirectly

alexcrichton:
This commit adds support for the `MAX_FLAT_PARAMS` variable in the
current canonical ABI draft. Argument lists which exceed a static number
of parameters are now always passed indirectly through memory. In export
this is dynamically allocated/freed and in imports this is statically
passed.


alexcrichton:
Note that the first commit here is https://github.com/bytecodealliance/wit-bindgen/pull/207

crates/gen-spidermonkey/src/lib.rs
```diff
@@ -48,191 +48,191 @@ lazy_static! {
             WasmSignature {
                 params: vec![],
                 results: vec![],
-                retptr: false,
+                retptr: false,indirect_params:false,
```
peterhuene, line 51: Looks like this file has some bad formatting, probably from conflict resolution?

crates/parser/src/abi.rs
```diff
@@ -654,6 +665,19 @@ def_instruction! {
             params: usize,
         } : [*params + 2] => [0],
 
+        /// TODO
```
peterhuene, line 668: Description needed.

crates/parser/src/abi.rs
```diff
@@ -654,6 +665,19 @@ def_instruction! {
...
+        Malloc {
+            realloc: &'static str,
+            size: usize,
+            align: usize,
+        } : [0] => [1],
+
+        /// TODO
```
peterhuene, line 675: Description needed.

peterhuene (approved):
Also looks good, just a few minor comments.

crates/gen-spidermonkey/src/lib.rs
alexcrichton: Oh this is me using a random macro and praying rustfmt will fix everything for me when, in fact, it did not.

PR #209, opened 2022-04-28
Split out the `flags` type from `record`

alexcrichton:
This commit creates a dedicated `flags` type which is distinct from the
`record` type rather than the previous inferred-from-the-structure
logic. This also additionally changes the canonical ABI of the `flags`
type where 64-bit flags are now passed as two `i32` values instead of
one `i64`. This ended up changing a significant amount of the logic
internally in each code generator, notably around the new lift/lower
behavior.

Along the way I tried to refactor code to support 64+ flags in a few
more places. While some support may be there, though, this is untested
and will need a full-fledged feature in the future.


peterhuene (approved):

codefromthecrypt:
what was the rationale for splitting a 64 bitflag into two fields? As you noticed in refactoring, this has a lot of impact and isn't universally desirable.

PR #211, opened 2022-04-29
Split out an `Enum` type from the `Variant` type

alexcrichton:
This was not 100% straightforward as the lowering of an enum wants to be
`the_value as i32` in the Rust-to-wasm generator, but that's not working
because sometimes `the_value` has type `&T` instead of `T` which can't
be cast to `i32`. This means that lowerings in Rust are now always
producing a `match` statement which should optimize to the same thing
but won't be as easy on the eyes.

Additionally a small change was made to the C code generator that
`expected<_, _>` is now represented as a struct-of-union instead of
special-cased to be an `enum`. The only reason it was special cased
prior was that it was accidentally interpreted as an `enum` due to the
`Variant::is_enum` check (which is now removed).


peterhuene (approved):
Looks great, just the tiny nit regarding a test case name now that enum and variant are distinct types.

PR #212, opened 2022-05-02
Remove custom hashing of types from `wit-component`

alexcrichton:
During the encoding process the `wit_parser::Interface` `types` array
already deduplicates types-via-hashing in the parsing/resolution stages
of creating an `Interface`. The `TypeId` and `Type` keys are also
hashable, so the duplication in `wit-component` shouldn't be necessary
since everything should already be handled externally.


peterhuene:
This exists because multiple `Interfaces` are deduped so that their type section representations can be condensed and I don't think the arenas for the ids are shared between multiple interfaces, no?

alexcrichton:
Ah yes indeed I forgot about that!

PR #219, opened 2022-05-05
Split out a `Union` type from `Variant`

alexcrichton:
This commit splits out the union specialization from `variant`,
finalizing the transition to the new AST of types available in the
current component model draft. This is mostly like the prior
splittings-out where there's some duplication but it's relatively
localized and hopefully not too hard to follow.


crates/parser/src/abi.rs
```diff
@@ -568,6 +568,21 @@ def_instruction! {
             ty: TypeId,
         } : [1] => [1],
 
+        /// TODO
```
peterhuene, line 571: TODO comment here and below.

peterhuene (approved):
Looks good 👍 Just the `TODO` comments, I think.

crates/parser/src/abi.rs
alexcrichton: Gah sorry I swear I'm not forgetting these intentionally, despite this being present in every PR I've sent so far...

PR #228, opened 2022-06-02
Implement parsing for the `stream` type.

sunfishcode:
Stream has two types, a stream element type and an end type. In the ABI,
it's represented as an i32 index. Lifting and lowering is not yet
implemented.


crates/gen-js/src/lib.rs
```diff
@@ -195,6 +195,7 @@ impl Js {
                     }
                     TypeDefKind::Variant(_) => panic!("anonymous variant"),
                     TypeDefKind::List(v) => self.print_list(iface, v),
+                    TypeDefKind::Stream(_) => panic!("anonymous stream"),
```
alexcrichton, line 198: I think this needs to be `todo!` since `panic!` here is used for "this should never happen"

crates/gen-rust/src/lib.rs
```diff
@@ -284,6 +285,9 @@ pub trait RustGenerator {
             TypeDefKind::Union(_) => {
                 panic!("unsupported anonymous type reference: union")
             }
+            TypeDefKind::Stream(_) => {
+                panic!("unsupported anonymous type reference: stream")
```
alexcrichton, line 289: I think that this will need to be supported so I think this will need to be `todo!`

crates/parser/src/ast.rs
```diff
@@ -514,6 +520,16 @@ impl<'a> Type<'a> {
...
 
+            // stream<T, Z>
+            Some((_span, Token::Stream)) => {
+                tokens.expect(Token::LessThan)?;
+                let element = Box::new(Type::parse(tokens)?);
+                tokens.expect(Token::Comma)?;
+                let end = Box::new(Type::parse(tokens)?);
+                tokens.expect(Token::GreaterThan)?;
+                Ok(Type::Stream(Stream { element, end }))
+            }
```
alexcrichton, line 531: Should syntax also be added for `stream<T>` where the `Z` is implicity `unit`?

alexcrichton (approved):
Could you open a tracking issue as well to track all the `todo!` macros? Otherwise I'm afraid that it's easy to get lost over time and forget that these aren't implemented.

crates/parser/src/ast.rs
sunfishcode: My inclination is to leave it explicit for now, and add implicit syntax later if needed.

sunfishcode:
> Could you open a tracking issue as well to track all the `todo!` macros? Otherwise I'm afraid that it's easy to get lost over time and forget that these aren't implemented.

I opened https://github.com/bytecodealliance/wit-bindgen/pull/228 to track the implementation of `stream`.

PR #229, opened 2022-06-03
Create development container for wit-bindgen

Kylebrown9:
This development container handles installing
* Rust
* Python/Pip
* Node/NPM
* the Rust targets
* the Python dependencies
* the NPM dependencies, and
* the WASI SDK.

Users should be able to open this container and immediately run `cargo test --workspace --release` and have it complete and pass.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -20,10 +20,9 @@ RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/
 # [Optional] Uncomment this line to install global node packages.
 # RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
 
-ENV WASI_SDK_VERSION=14.0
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Would it be possible to test this on CI as well?

ricochet:
I pulled and tested this on macos-darwin-aarch64. @Kylebrown9 tested on Windows+WSL.

As far as testing in the github workflow... it's a little difficult. The VS Code-y bits are perhaps the most difficult aspect. For the general docker build the breakdown would be something like:

- Build dev container docker image (ideally also push to a registry)
- We go off the rails a little since we don't have access to VS Code in this next step:
  - Exec into container. This part is annoying in github actions if you don't push to a registry
  - Run the steps that VS Code runs, e.g. install rust. Run the postCreateCommand script.
  - Then finally `cargo test --workspace --release`
  
Now a lot of this may be generally available in GitHub prebuilds : https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-codespaces-prebuilds. It might be better to wait on that.

alexcrichton:
Ah ok, well seems like a good improvement in the meantime!

PR #231, opened 2022-06-06
Updating WIT specification for f32 and f64

Kylebrown9:
Update WIT.md specification `f32` and `f64` to `float32` and `float64` to match [wit-bindgen lexer](https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/parser/src/ast/lex.rs#L249) and component model docs.


willemneal:
Thanks for this catch!  I think it would also be a good idea to add a version corresponding to the current `0.1.0` tag so that it's clear that the current syntax matches with that version of the parser crate.

Issue #163, opened 2022-02-26
function -> func


badeend:
WIT reuses the exact same names as Core Wasm & Interface Types for pretty much everything (variant, string, tuple, flags, bool, enum, option, union, expected, ...).

Except one: `function`

I suggest renaming this to `func` for consistency.

alexcrichton:
Completed in #232, thanks for the suggestion!

PR #232, opened 2022-06-06
Replace `function` with `func`

Kylebrown9:
This implements #163 changing the `function` WIT keyword to be `func`.


alexcrichton:
Thanks for this! I'll note though that I tagged #163 with `syntax` because I think we need to probably confirm we want to deviate from core wasm's and the component model's precedence of "func" with "function" but I don't think it's otherwise necessarily a clear-cut decision that we should migrate to "func". I think it would be good to get some feedback from other `*.wit` stakeholders and see what their thoughts are on a change like this as well.

Kylebrown9:
We'd love to start having those conversations. We're not dead-set on implementing this change, we just want to settle this question sooner than later because switching will only get harder.

peterhuene:
While I personally prefer `func` based on the reasoning that it corresponds to the underlying WebAssembly proposal (also less typing...), I don't have a strong opinion on the particular token we use; however, I strongly agree that this needs to be decided upon by the stakeholders (open question: how do we identify the stakeholders?) sooner than later to avoid having a larger wit corpus in the wild to update.

lukewagner:
Coincidentally, I was just having a chatting with a number of distinct wit-bindgen users and, taking a quick poll of the room, essentially everyone said they preferred `func` to `function` (but were "fine" with function and hence hadn't cared enough to raise the issue).  Personally, I think the argument that `function` is the only token that doesn't line up directly with the component model is pretty compelling.

Kylebrown9:
On a related note, the other thing that stands out is `float32`/`float64` which is consistent between `wit-bindgen` and the component model, but both disagree with the WebAssembly spec which uses `f32`/`f64`. What was the rationale for that split?

peterhuene:
@Kylebrown9 See [Luke's prior comment](https://github.com/WebAssembly/component-model/pull/6#issuecomment-1052323290) as to why there's a discrepancy between f32/f64 and float32/float64.

alexcrichton:
Ok sounds like there's no major objections to this, and personally I find the symmetry between core wasm and the component model with the `*.wit` syntax pretty compelling. For that I'm going to merge, thanks @Kylebrown9 for doing this refactor!

Issue #236, opened 2022-06-10
Update `WIT.md` to use `func` keyword instead of `function`


eyebrowsoffire:
I have been familiarizing myself with the .wit format, and playing around with it in the online bindings generator. One thing I noticed is that the `WIT.md` grammar and examples all use the keyword `function`, but that doesn't actually work in the bindings generator, it uses the keyword `func`. I'm not sure if the documentation is out of date and the implementation is correct, or the other way around, but it would be good if these two things matched up.

peterhuene:
It just changed to `func` very recently, so there's probably places where we failed to update the examples. If you spot any, we'd happily take a PR!

eyebrowsoffire:
Oh it actually *just* updated. I was using a newer version of the online bindings generator but reading a version of `WIT.md` from yesterday! That being said, I did find that the grammar itself uses the wrong string, so I did open a PR: https://github.com/bytecodealliance/wit-bindgen/pull/238

peterhuene:
Thanks again for putting up a fix for this issue!

Issue #237, opened 2022-06-10
update online demo to use func instead of function


attacker0211:
https://bytecodealliance.github.io/wit-bindgen/

peterhuene:
Hi @attacker0211. Thanks for reporting this. wit-bindgen recently changed the `function` keyword to `func`, but it appears the online playground wasn't updated.

#238 should fix this shortly.

peterhuene:
This should now be resolved. Thanks!

PR #238, opened 2022-06-10
Update grammar for `func` in `WIT.md`

eyebrowsoffire:
Just a documentation update for some instances of `function` that should be replaced with `func`.

https://github.com/bytecodealliance/wit-bindgen/issues/236

### DIFF
--- a/WIT.md
+++ b/WIT.md
@@ -368,7 +368,7 @@ union-cases ::= ty,
               | ty ',' union-cases?
 ```
 
-## Item: `function`
+## Item: `func`
 
 Functions can also be defined in a `*.wit` document. Functions have a name,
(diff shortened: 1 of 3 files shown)


peterhuene:
@eyebrowsoffire thanks for the PR.

I did notice from the error message in #237 that [it errors with `fn` instead of `func`](https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/parser/src/ast.rs#L225). Would you mind including an update to that string as part of this PR?

peterhuene (approved):
👍 Thanks again! Will merge when CI is green.

eyebrowsoffire:
Awesome! Thanks for the quick review.

PR #239, opened 2022-06-10
wit-component: update to latest component model proposal.

peterhuene:
This PR updates and refactors `wit-component` so that it produces
components compatible with the latest component model proposal changes.

Most of the big changes were to the encoding, which also was reimplemented in
places to make the assignment of indexes more manageable.

The `post-return` option is not yet implemented as wit-bindgen has not added
support for it yet, so right now any components created by this tool will leak
any data returned by exported functions.

I also made the tool more consistently use `IndexMap` and `IndexSet` (an
existing dependency) so that iterating of various things is stable.


peterhuene:
Looks like it might be time to remove (or at least stop testing) `wasmlink`, as it depends on the removed support for the previous module linking proposal that's now unsupported in wasm-tools.

PR #240, opened 2022-06-11
Remove `wasmlink`, `wasmlink-cli`, and `test-modules`.

peterhuene:
This PR removes `wasmlink`, `wasmlink-cli`, and `test-modules` from the
`wit-bindgen` repo.

`wasmlink` is based on an outdated module linking proposal and predates any of
the work on the component model. As such, it never really graduated beyond a
useful demo of how multiple core WebAssembly modules might exchange interface
types.

Now that the work on the component model is progressing, the underlying
wasm-tools no longer support the module linking proposal; therefore, `wasmlink`
is becoming a maintenance burden.

As the author of `wasmlink`, I think the time has come to remove it from the
tree. Any parties that need a copy of `wasmlink` for whatever reason may fork
it prior to this commit being merged.


peterhuene:
Pinging @saulecabrera for your thoughts on this too.

saulecabrera (approved):

PR #241, opened 2022-06-12
Correct *.wit format docs to reflect current implementation of resource-defs

Gearme:
Correct the documentation of the current *.wit format implementation to no longer falsely claim that function definitions in resource types need to (or even can) be divided by commas.

Currently works:
```
resource foo{
    bar: func()
    fizz: func()
}
```

Also works:
```
resource foo{
    bar: func() fizz: func()
}
```

Currently does not work:
```
resource foo{
    bar: func(),
    fizz: func()
}
```


alexcrichton:
Thanks!

PR #242, opened 2022-06-13
Improve naming of union cases in Rust

Liamolucko:
Rather than naming them by their index, 'V0', 'V1', 'V2' etc., this changes the Rust backend to name union cases after their types.

For example, this union:

```wit
union foo {
    string,
    u8,
}
```

would previously have generated this enum:

```rust
enum Foo {
    V0(String),
    V1(u8),
}
```

With this change, it will generate this enum:

```rust
enum Foo {
    String(String),
    U8(u8),
}
```

I think that this makes code using these unions much more self-explanatory; it's more obvious that `Foo::String(s)` contains a string than `Foo::V0(s)`.

This handles cases where a union contains more than one of a single type by reintroducing a suffix to those cases - so, `union { char, string, string }` would get case names of `Char`, `String0` and `String1`.

I've named the cases after the types' Rust names rather than their WIT names, so `s32` gets a name of `I32` and `expected<T, E>` gets a name of `Result`. I think that makes it make more sense within Rust, but I don't mind it either way.

This does have some degenerate cases. `option<T>` is named `OptionalT`, and `list<T>` is named `TList`, so `option<list<option<list<list<T>>>>>` would get the unwieldy name of `OptionalOptionalTListListList`. I doubt that types like that would come up much in practice though.

### DIFF
--- a/crates/gen-rust/src/lib.rs
+++ b/crates/gen-rust/src/lib.rs
@@ -1,4 +1,5 @@
 use heck::*;
+use std::collections::HashMap;
 use std::fmt::{self, Write};
 use std::iter::zip;
 use wit_bindgen_gen_core::wit_parser::abi::{Bitcast, LiftLower, WasmType};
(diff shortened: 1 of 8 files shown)


crates/gen-rust/src/lib.rs
```diff
@@ -379,6 +380,87 @@ pub trait RustGenerator {
...
+        // We do this instead of just modifying the strings so that we can
+        // still check against the original.
+        let mut case_names = Vec::new();
+        for case in union.cases.iter() {
+            let name = self.name(iface, &case.ty);
+            let mut suffix = None;
+
+            let mut same_named = case_names
+                .iter_mut()
+                .filter(|(other_name, _): &&mut (String, Option<usize>)| name == **other_name);
```
alexcrichton, line 436: To avoid potential future pitfalls could this use a hash map or similar instead of a linear scan?

crates/gen-rust/src/lib.rs
```diff
@@ -379,6 +380,87 @@ pub trait RustGenerator {
...
+            Type::Id(id) => {
+                let ty = &iface.types[*id];
+                match &ty.name {
+                    Some(name) => name.to_camel_case(),
+                    None => match &ty.kind {
+                        TypeDefKind::Option(ty) => format!("Optional{}", self.name(iface, ty)),
+                        TypeDefKind::Expected(_) => "Result".to_owned(),
+                        TypeDefKind::Tuple(_) => "Tuple".to_owned(),
+                        TypeDefKind::List(ty) => format!("{}List", self.name(iface, ty)),
+                        TypeDefKind::Stream(s) => format!("{}Stream", self.name(iface, &s.element)),
```
alexcrichton, line 410: Due to the recursive nature of this method could this method take `&mut String` and the methods push onto it instead of having an owned string returned from each level?

alexcrichton (approved):
Sounds like a great idea to me, thanks! Some minor implementation nits below but otherwise seems good to go to me 👍

PR #252, opened 2022-06-17
Introduce SourceBuilder utility

Kylebrown9:
This PR continues my work to simplify the way we create/use utilities for generating source code that I started when I introduced the `Dependencies` struct. The `SourceBuilder` struct is simply a collection of references to a `Source`, `Interface`, and `Dependencies` so that its methods can encapsulate all of the logic required for source generating, like marking requirements for intrinsics and looking up type info.

This reduces our reliance on `mem::take` and `mem::replace` which we were using to swap `Source`s into and out of a primary slot that all of our utilities assume the current source is in. In the future, I expect this will allow us to manage `Source`s more cleanly/separately and stop relying on the `mem` utilities.

### DIFF
--- a/crates/gen-wasmtime-py/src/source.rs
+++ b/crates/gen-wasmtime-py/src/source.rs
@@ -1,0 +1,495 @@
+use heck::*;
+use wit_bindgen_gen_core::wit_parser::*;
+
+use crate::dependencies::Dependencies;
+
+/// A [Source] represents some unit of Python code
+/// and keeps track of its indent.
+#[derive(Default)]
+pub struct Source {
+    s: String,
+    indent: usize,
+}
+
+impl Source {
+    /// Appends a string slice to this [Source].
+    ///
+    /// Strings without newlines, they are simply appended.
+    /// Strings with newlines are appended and also new lines
+    /// are indented based on the current indent level.
+    pub fn push_str(&mut self, src: &str) {
+        let lines = src.lines().collect::<Vec<_>>();
+        let mut trim = None;
+        for (i, line) in lines.iter().enumerate() {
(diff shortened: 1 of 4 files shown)


willemneal:
I love it! It is much nicer to read and keep all the benefits.  I do wonder how this could be made to be more generalizable, but that is the same issue for all generators.

crates/gen-wasmtime-py/src/source.rs
```diff
@@ -0,0 +1,495 @@
...
+    /// e.g.
+    /// > """
+    /// >
+    /// > Line one of docs node
+    /// >
+    /// > Line two of docs node
+    /// >
+    /// > """
+    pub fn docstring(&mut self, docs: &Docs) {
+        let docs = match &docs.contents {
```
willemneal, line 70: I'm still learning the rusty patterns and here I wonder what your thoughts are with using 
```rust 
docs.contents.as_ref().map(|docs| {...})
```
Or

```
if let Some(docs) = &docs.contents {
...
}
```
Again just a style choice, but personally like avoiding `returns` in rust and have come to like the latter choice.

Kylebrown9: This is carried over directly from the docstring PR, which itself pulled this code from the original `comment` implementation. I don't have any strong opinions on this myself.

willemneal: Aw makes sense.  Honestly all of them are readable so doesn't really matter.

alexcrichton (approved):

alexcrichton:
Seems reasonable to me, thanks for this!

PR #253, opened 2022-06-21
Update JS enum representation

Liamolucko:
Updates the representation of enums in JS to match the component model explainer.

Instead of being mapped to TypeScript enums (an integer discriminant, with constants to check against for each case), they're mapped to strings of the case names, the same as WebIDL enums.

I was originally going to make one big PR to update the representation of all the JS types, but I'm doing it one by one so that the whole thing doesn't get blocked if we decide to change any of the representations upstream.


alexcrichton:
Looks good to me, thanks for this!

PR #254, opened 2022-06-21
Update JS flags representation

Liamolucko:
Updates the representation of flags in JS to match the component model explainer.

Previously, flags were represented as bitfields in JS. This PR changes them to be represented as objects with boolean fields for each flag instead, matching the component model explainer.


alexcrichton (approved):
I agree it's best to follow the explainer, thanks for this!

alexcrichton:
Got the merge order right this time which makes this view much cleaner. Again sorry I forgot about this!

PR #255, opened 2022-06-22
Make `Docs::contents` `None` for regular comments.

Liamolucko:
There was already a check to make it `None` for something with no comments whatsoever, but it was still `Some` if all of the comments on something were regular comments. This fixes that so that it only becomes `Some` once an actual doc comment is found.

I ran into this because it was making the JS enum doc generation from #253 produce some odd results. This WIT:

```wit
// a comment
enum foo { bar }
```

Should have produced this type declaration:

```ts
/**
* # Variants
* 
* ## `"bar"`
*/
export type Foo = "bar";
```

But actually produced this:

```ts
/**
* 
* # Variants
* 
* ## `"bar"`
*/
export type Foo = "bar";
```

Note the extra leading line, caused because `foo`'s docs were `Some("")` instead of `None`. With this PR, it works as expected.

### DIFF
--- a/crates/parser/src/ast/resolve.rs
+++ b/crates/parser/src/ast/resolve.rs
@@ -543,16 +543,15 @@ impl Resolver {
     }
 
     fn docs(&mut self, doc: &super::Docs<'_>) -> Docs {
-        if doc.docs.is_empty() {
-            return Docs { contents: None };
-        }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #256, opened 2022-06-22
Update JS integer lowering

Liamolucko:
Updates JS integer lowering to match the component model explainer (assuming that WebAssembly/component-model#52 is merged).

The main differences are that out-of-range numbers now wrap instead of throwing, and that non-integers are now rounded down instead of throwing.


crates/gen-js/src/lib.rs
```diff
@@ -2488,6 +2495,64 @@ impl Js {
                 }
             "),
 
+
+            Intrinsic::ToInt32 => self.src.js("
+                export function to_int32(val) {
+                    return val >> 0;
```
alexcrichton, line 2501: Should one-liners like this perhaps be inlined instead of going through an intrinsic? I'm not sure of the perf implications of using this indirection.

crates/gen-js/src/lib.rs
```diff
@@ -2488,6 +2495,64 @@ impl Js {
...
+                export function to_uint8(val) {
+                    val >>>= 0;
+                    val %= 2 ** 8;
+                    return val;
+                }
+            "),
+
+            Intrinsic::ToBigInt64 => self.src.js("
+                export function to_int64(val) {
+                    return BigInt.asIntN(64, val);
```
alexcrichton, line 2547: Similar to above it may be better to do this inline instead of as an intrinsic perhaps?

alexcrichton (approved):

crates/gen-js/src/lib.rs
Liamolucko: I put together some quick benchmarks, and it does seem to be _slightly_ faster inline:

```
u32RoundtripBefore    7.65 µs/iter    (7.15 µs … 41.26 µs)   7.19 µs  21.81 µs  23.23 µs
u32RoundtripAfter     7.42 µs/iter    (6.84 µs … 67.78 µs)   7.12 µs  21.64 µs  22.67 µs

summary
  u32RoundtripAfter
   1.03x faster than u32RoundtripBefore

s32RoundtripBefore    7.34 µs/iter    (6.84 µs … 79.61 µs)   6.89 µs  21.56 µs  22.68 µs
s32RoundtripAfter     6.83 µs/iter     (6.75 µs … 7.12 µs)   6.89 µs   7.12 µs   7.12 µs

summary
  s32RoundtripAfter
   1.07x faster than s32RoundtripBefore

u64RoundtripBefore   66.25 µs/iter  (59.18 µs … 676.27 µs)  61.51 µs  133.4 µs 141.81 µs
u64RoundtripAfter    64.96 µs/iter  (57.94 µs … 323.74 µs)  61.22 µs 131.25 µs 143.86 µs

summary
  u64RoundtripAfter
   1.02x faster than u64RoundtripBefore

s64RoundtripBefore   64.33 µs/iter  (57.48 µs … 295.28 µs)  60.06 µs  130.5 µs 140.59 µs
s64RoundtripAfter    63.77 µs/iter   (57.17 µs … 404.7 µs)  59.76 µs 130.23 µs 136.73 µs

summary
  s64RoundtripAfter
   1.01x faster than s64RoundtripBefore
```

(the 'after' benchmarks are with the conversions inlined)

It's not a massive difference, though.

alexcrichton: Ah ok that seems negligible enough that having a uniform representation here is probably worth it, thanks for measuring though!

alexcrichton:
Apologies for not merging this at the time I'm ot sure why I didn't earlier. My merge with `main` appears to have confused GitHub but I'll land this once CI passes

PR #258, opened 2022-06-22
Update JS float lowering

Liamolucko:
Values are now coerced to numbers instead of throwing, matching the component model explainer.

I haven't implemented NaN canonicalisation, because I'm not quite sure how to. It's not too hard when writing to memory - you just have to write the bit pattern as an integer instead of a float - but I'm not sure how to properly do it when passing the floats as parameters directly.

[The spec for `ToWebAssemblyValue`](https://www.w3.org/TR/wasm-js-api/#towebassemblyvalue) isn't very clear about how NaNs are handled - it converts from a JS number to an `f64` without any extra clarification. This is strange because JS numbers aren't normal floats - they have a single NaN value, without a defined bit pattern or anything.

Worse, the behaviour does actually vary between engines in practice - take this code to roundtrip a NaN bit pattern through a float and back:

```wat
(module
    (func $f64toi64 (param $float f64) (result i64)
        local.get $float
        i64.reinterpret_f64
    )
    (func $i64tof64 (param $bits i64) (result f64)
        local.get $bits
        f64.reinterpret_i64
    )
    (export "f64toi64" (func $f64toi64))
    (export "i64tof64" (func $i64tof64))
)
```

```js
console.log(f64toi64(i64tof64(0x7ff8000000001234n)).toString(16));
```

Chromium preserves the bits and prints `7ff8000000001234`, whereas Firefox canonicalises them and prints `7ff8000000000000`.

So, I'm going to leave that out for now.

### DIFF
--- a/crates/gen-js/src/lib.rs
+++ b/crates/gen-js/src/lib.rs
@@ -57,8 +57,6 @@ enum Intrinsic {
     ClampHost,
     ClampHost64,
     DataView,
-    ValidateF32,
-    ValidateF64,
     ValidateGuestChar,
     ValidateHostChar,
     ValidateFlags,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Seems reasonable to me, thanks!

PR #260, opened 2022-06-23
Propose a Bindings Generators Versioning Policy

Kylebrown9:
PR introducing ADRs generally and a specific ADR to address versioning.

### ADR 0000 - Use Markdown Any Decision Records
Proposes the adoption of [Architectural Decision Records](https://adr.github.io/) (specifically the [MADR format](https://adr.github.io/madr/)) for recording design decisions.

### ADR 0001 - Separately Version Bindings Generators
Proposes that each bindings generator be versioned separately.
[GH preview of the this ADR's markdown](https://github.com/Kylebrown9/wit-bindgen/blob/release-process-adr/docs/decisions/0001-Separately-Version-Bindings-Generators.md)


alexcrichton:
Thanks for getting started on this! As Nick mentioned on Zulip this may be best done with an RFC rather than a new format done just in this repository. 

Otherwise though for a release process I think we'll also want more details about things like the stability policy, what's covered, what isn't, when releases are made, etc. The [Wasmtime 1.0](https://github.com/bytecodealliance/rfcs/blob/main/accepted/wasmtime-one-dot-oh.md) RFC has has some examples of this.

Of particular note for wit-bindgen is that the project is very much in the "not yet finished" state so nothing is basically stable and I don't think we're even at a point of when we can predict when everything will be stable and what the stable interface will look like. I don't personally know how best to have a release process within those constraints but that's where I'm hoping others can help!

Kylebrown9:
Thanks Alex, I've been bouncing ideas off of @ricochet and building up a full release process design draft that covers more of the details that you mentioned, this ADR proposal was intentionally just one part of that bigger picture. We had hoped that we could use ADRs to discuss each decision that motivates the release process incrementally so that there wouldn't be one large debate over the whole thing, but I can get a draft into RFC format in the near future if that's the better approach.

Issue #259, opened 2022-06-23
Upgrade to latest `wasmtime` and component model


dicej:
I've been experimenting with upgrading to the latest `wasmtime` (i.e. the main branch for now, but I would use the most recent release in a PR).  As a consequence of the upgrade, [instantiate_smw](https://github.com/bytecodealliance/wit-bindgen/blob/2b86ad0747a4b32eec98e17979dd2de773716a2d/crates/gen-wasmtime/tests/runtime.rs#L55) fails to build due to its use of the module linking API, which has been removed as of `wasmtime` 0.36.

I'd like to update `instantiate_smw` to use the component model, although I realize it's [not yet fully implemented](https://github.com/bytecodealliance/wasmtime/issues/4185) as of this writing.  I don't know whether `wasmtime`'s implementation is complete enough yet for `wit-bindgen`'s needs; one goal of this experiment is to answer that question.  Either way, I'm eager to start using the component model in other projects, so it will be helpful to understand the specific limitations of the current (partial) implementation -- particularly with regard to linking modules together.  I'd also be interested in helping move that implementation forward if there are opportunities to do so.

My current (possibly naive) strategy is to change `instantiate_smw` to create a `component::Component` which instantiates both the spidermonkey and input module, passing the former as an argument to the latter.  Then I would instantiate that component using a `component::Linker`, call `wizer.initialize` on the spidermonkey module instance, and return as before.  This raises a few questions:

- Given that the `wit_bindgen_wasmtime::export!` and `wit_bindgen_wasmtime::import!` macros (not to mention `wasmtime_wasi::add_to_linker`, etc.) expect a module `Linker`, not a component `Linker`, it's not clear to me how to use them in a component context.  Is there (or will there be) some way to add a module `Linker` into a component `Linker` for the purpose of dynamic instantiation?  Or am I thinking about it the wrong way?
- More generally, what's the vision for code which previously used module linking and will eventually need to transition to the component model?  Will there be a simple recipe to follow, or will big (e.g. architectural) changes be necessary?

dicej:
Reading over what I wrote above, I realize this might be more appropriate as an issue in the `wasmtime` repo.  If so, I'll be happy to move it over there.

Kylebrown9:
I think moving it to the `wasmtime` repo would make sense. You could make your first bullet point its own question here if you do move this one since it is more specifically about how `wit-bindgen` should work.
[...]

PR #261, opened 2022-06-24
chore: update to latest wasmtime release (0.38.0)

dicej:
This updates to `wasmtime` v0.38.0.  Because module linking support has been
removed from `wasmtime` as of v0.36.0, and the component model implementation is
not yet mature enough to use as a replacement (see
https://github.com/bytecodealliance/wasmtime/issues/4303 for details), it is
necessary to temporarily disable the Spidermonkey tests.  I've done this by
commenting-out the broken code in `instantiate_smw`, prefixed its name with and
underscore, and renamed the relevant host.rs files to ignore_host.rs to hide
them from the test infrastructure.

Once the component model implementation is mature enough and the broader
ecosystem has been updated to use it, we can update and restore the Spidermonkey
tests.

fixes https://github.com/bytecodealliance/wit-bindgen/issues/259


alexcrichton (approved):

dicej:
My original intention for https://github.com/bytecodealliance/wit-bindgen/issues/259 was to update the Spidermonkey tests to use the component model, which would hopefully serve as an example for other projects of how to migrate their code from module linking to the component model.  However, Nick and Alex explained why that's not worth the effort given the current state of the ecosystem.

So this PR won't have much value for projects which use both `wit-bindgen` and `wasmtime`'s old module linking support.  On the contrary, it's more likely to cause pain because now such projects will be unable to upgrade `wasmtime` _or_ `wit-bindgen` to the latest respective versions.  So I hesitated to even open the PR.

Before we merge this, I'd like to know: will this be helpful to anyone, or will it just cause pain?  If the latter, I'll happily close it.

alexcrichton:
I'm gonna go ahead and merge this and it's small enough we can back out if necessary in the near future. Thanks for the PR though @dicej!

PR #267, opened 2022-06-29
Implement `std::fmt::Write` for `Source`

### LINKED ISSUE #161: Design question: How to make source generation more readable
The current design of using a "source builder" has a draw back that it makes the code harder to read.

```rust
self.src.ts("Record<");
self.print_ty(iface, &r.fields[0].ty);
self.src.ts(", ");
self.print_ty(iface, &r.fields[1].ty);
self.src.ts(">");
```

vs

```rust

let first = record.fields[0].ty;
let second = record.fields[1].ty;
self.src.ts(&format!("Record<{first},{second}>"))

```

The latter assumes that a custom Display for `Type` was implemented. Obviously this would go against the current design.

So my question is if there is a middle ground?  A macro like `format!` but custom to this design.

```rust

to_src!( self.src.ts, iface, "Record<{first},{second}>")
// which becomes
self.src.ts("Record<");
self.print_ty(iface, &first);
self.src.ts(", ");
self.print_ty(iface, &second);
self.src.ts(">");
```

If you find this macro useful, what is your advice for writing it `macro_rules` or a `proc_macro`?  I'm leaning toward the latter.

Liamolucko:
This allows using `write!` and `writeln!` to generate code with format strings instead of `format!`, which is both shorter and more efficient.

This semi-resolves #161 for cases where the things being interpolated implement `Display`, although not for stuff like `print_ty`. I could imagine solving that with an `fn format_ty(...) -> impl Display`, but I haven't done that here; this just aims to allow simplifying the very common `src.push_str(&format!("some {thing}\n"))` to `writeln!(src, "some {thing}").unwrap()`.

### DIFF
--- a/crates/gen-core/src/lib.rs
+++ b/crates/gen-core/src/lib.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use std::collections::{btree_map::Entry, BTreeMap, HashMap};
+use std::fmt::{self, Write};
 use std::ops::Deref;
 use std::path::Path;
 use wit_parser::*;
@@ -395,6 +396,13 @@ impl Source {
 
     pub fn as_mut_string(&mut self) -> &mut String {
         &mut self.s
+    }
+}
+
+impl Write for Source {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        self.push_str(s);
+        Ok(())
     }
 }
 


Kylebrown9:
This seems useful for the reasons you say. If we're going to do it though, I think the PR should include some refactoring to use `write!` in generators (e.g. `gen-rust-wasm`) that use the type `Write` is now available on, to demonstrate this new pattern and make the PR more tangible.

I'll also note that while this will let us convert some `format!`s to `write!`s, source building as currently designed can have side effects like marking the need for an intrinsic. So there are limits to when this pattern will be usable.

alexcrichton:
Seems reasonable to me as well, thanks! If the main goal here is ergonomics perhaps a new macro could be introduced which bakes in the `unwrap()` as well though?

Kylebrown9:
Maybe something like `uwrite!` and `uwriteln!` for "unwrapped write"?

Liamolucko:
> Maybe something like `uwrite!` and `uwriteln!` for "unwrapped write"?

Good idea, I've implemented that.

PR #269, opened 2022-07-01
Implement parsing for the `future` type.

sunfishcode:
Future has one type. In the ABI, it's represented as an i32 index. Lifting and
lowering is not yet implemented.


Kylebrown9:
Please update `WIT.md` to reflect this new syntax.

PR #276, opened 2022-07-10
Remove duplicate definition of identifiers from WIT.md

Liamolucko:
WIT.md had an outdated definition of identifiers, which only allowed ASCII and included the old method of putting identifiers in quotes to escape keywords.

This removes that definition in favour of the up-to-date definition further down in the file.


alexcrichton:
Oop sorry looks like I just merged another PR which created a conflict, but otherwise looks good to me, thanks!

PR #277, opened 2022-07-10
docs: update wit.md according current parser impl

7sDream:
Updates:
- delete `[` in operators since we do not have this token
- delete `-` and `A-Z` from `identifier` since we force kebab case now
- add a sentence to make it clear that `id` is short for `identifier`
- add `unit` in `ty`
- remove `_` in `expected` since we do not support this grammer anymore

In current parse impl, `ty` has a `'handle' id` variant but behavor as same as `id` for resource type, it seems `handle` is deprecated?
I just ignore it for now.

### DIFF
--- a/WIT.md
+++ b/WIT.md
@@ -58,10 +58,10 @@ comment ::= '//' character-that-isnt-a-newline*
           | '/*' any-unicode-character* '*/'
 ```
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me, thanks!

PR #279, opened 2022-07-11
Ignore clippy warnings in generated code

7sDream:
Just meet `clippy::unused_unit` warning with a tiny test wit file:

because of generated code has redundant unit type/value in function signature/body, example:

```rust
    pub fn send(evt: EventParam<'_>) -> () {
        unsafe {
            // ... omit ...
            wit_import(result3_0, result3_1, result3_2, result3_3, result3_4);
            ()
        }
    }
```

This PR attempts to remove them, which in turn eliminates thoose warnings.

PS：Didn't find a good way to determine the return type during instruction translation, currently using string matching, if there is a better way please let me know.

### DIFF
--- a/crates/gen-rust-wasm/src/lib.rs
+++ b/crates/gen-rust-wasm/src/lib.rs
@@ -1580,10 +1580,8 @@ impl Bindgen for FunctionBindgen<'_> {
                 match amt {
                     0 => {}
                     1 => {
-                        if &operands[0] != "()" {
-                            self.push_str(&operands[0]);
-                            self.push_str("\n");
-                        }
+                        self.push_str(&operands[0]);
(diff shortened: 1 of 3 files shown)


alexcrichton:
Personally I would prefer to keep the code generator clean so I would rather not be appeasing clippy in the generated code. Is there a way to get clippy to ignore all generated code?

Liamolucko:
From rust-lang/rust-clippy#702, it looks like you can blanket `#[allow(clippy)]`.

7sDream:
> I would prefer to keep the code generator clean

it makes sense.

Add a `allow` attr on the top of each generated module seems a better idea, will try to change to use this method today.

PR #280, opened 2022-07-11
Implement syntax support for resource subtyping.

sunfishcode:
This implements a basic syntax for [resource subtyping], using the
`implements` keyword:

```
resource file implements preopen {
   ...
}
```

I'm not attached to this specific syntax or keyword name, but it is
useful to have something that we can start using to mock up wit files.

[resource subtyping]: https://github.com/WebAssembly/component-model/issues/60

### DIFF
--- a/crates/parser/src/ast.rs
+++ b/crates/parser/src/ast.rs
@@ -58,6 +58,7 @@ struct UseName<'a> {
 pub struct Resource<'a> {
     docs: Docs<'a>,
     name: Id<'a>,
+    supertype: Option<Id<'a>>,
     values: Vec<(bool, Value<'a>)>,
 }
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
This is small enough that it seems fine to throw in, but in general I have no idea how this will map to all the various features of each language and how usage will look like in the component model. Resources are already complicated enough as-is and not fully implemented anywhere (or spec'd), but do you have an idea in mind for how this will work out in all the generators in the long run?

sunfishcode:
I don't have a full vision for how this works in all the generators yet. I expect we'll do the work of fully fleshing out how these work when we add handles to the canonical ABI.

PR #281, opened 2022-07-14
wit-component: clean up type exports. 

peterhuene:
This PR refactors encoding in `wit-component` to only export types from the
default interface and stop exporting types from instance types used in imports.

It also makes "type only" encodings more explicit with `InterfaceEncoder` that
will encode the types from a single interface only.


peterhuene:
@alexcrichton I've changed this PR to continue to export types from `wit-component`, but it cleans things up a little bit so that we don't export types from the instance types used in imported interfaces.

alexcrichton (approved):

PR #284, opened 2022-07-19
Fix some minor bugs in the WIT parser

Liamolucko:
- The parser would panic on empty block comments (`/**/`), since it would interpret the `/**` as the start of a doc comment, leaving the block comment with `/` instead of `*/` at the end and tripping an assertion.
- The parser would panic on `.md` files, since it assumed they were `.wit.md` files and tried to strip the nonexistent `.wit` from the filename. I worked around this by changing the parser to take everything before the first `.` as the name of the interface. 

I found these while trying out `cargo-fuzz`.


alexcrichton:
Nice!

PR #296, opened 2022-08-03
wit-bindgen-rust: fix lowering and lifting of zero-length lists.

peterhuene:
This PR fixes the Rust bindings generated for lifting and lowering of lists
to account for lists of zero-length.

It is undefined behavior to call `std::alloc::alloc` with a zero-sized layout.

Additionally, as the behavior of the `canonical_abi_realloc` implemented in
`wit-bindgen` is to return a non-null dummy pointer for a zero-sized alloc, the
bindings were calling `std::alloc::dealloc` with a pointer that was not
allocated with `std::alloc::alloc` (also undefined behavior).

To fix this, the bindings now account for zero-length lists and skip allocation
and deallocation where appropriate.

### DIFF
--- a/crates/gen-rust-wasm/src/lib.rs
+++ b/crates/gen-rust-wasm/src/lib.rs
@@ -750,12 +750,16 @@ impl FunctionBindgen<'_> {
 
     fn emit_cleanup(&mut self) {
         for (ptr, layout) in mem::take(&mut self.cleanup) {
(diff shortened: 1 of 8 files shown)


alexcrichton (approved):
Would it be possible to add some tests exercising this as well which try to transfer empty lists or nonempty lists of zero-sized types? (also perhaps empty strings?)

Issue #244, opened 2022-06-15
Rename crate folders


Kylebrown9:
One thing I've noticed that is a "gotcha" for people exploring and using wit-bindgen is that it doesn't consistently use the guest/host terminology throughout its crate folders and names. This means that people are often confused when they try to figure out whether the code they're looking for is in `gen-rust` or `gen-wasmtime` because both generate Rust code and the user intends to use Wasmtime.

# Proposal
Rename the crate folders to follow a consistent naming scheme that makes guest/host more clear.

## Guest and Host Bindings Generators
Normalize all folder names for guest and host binding generators to
* Guest: `gen-<lang>-guest`
* Host: `gen-<lang>-host-<embedding>`

For example:
* `gen-c-guest` formerly `gen-c`
* `gen-rust-guest` formerly `gen-rust-wasm` 
* `gen-rust-host-wasmtime` formerly `gen-wasmtime`
* `gen-py-host-wasmtime` formerly `gen-wasmtime-py`
* `gen-js-host-??` formerly `gen-js` (definitely open to suggestions for what to call this embedding)

Other notes:
* Give `gen-rust` a suffix e.g. `-utils` or `-lib` that indicates that it's a library for generators but not really a generator itself
* `gen-markdown` doesn't have guest / host semantics, so it can stay as it is without a guest or host suffix.
* I'm honestly not familiar enough with the SpiderMonkey generator to classify how it should fit in this taxonomy, maybe someone has an idea of where it fits or it needs to be an exception.

## Rust Public Crate Folders
Rename the crates for use in Rust by our end users to follow a similar naming scheme to the generators

* `rust-guest` formerly `rust-wasm`
* `rust-guest-macro` formerly `rust-wasm-impl`
* `rust-host-wasmtime` formerly `wasmtime`
* `rust-host-wasmtime-macro` formerly `wasmtime-impl`

## Other Folders
* Potentially rename `gen-core` to `bindgen-core` so that it doesn't fit in the above convention as a "thing that generates core" (which will also conveniently sort it at the top of folder views)
* Potentially rename `parser` to `wit-parser` for clarity

# Wrap up
I've worked in some areas of wit-bindgen but not as much or at all in others, so if I categorize something incorrectly here definitely let me know and I'll update the proposal. I'll put together a PR tomorrow based on feedback so that we can get a sense of what it would look like. Thanks!

alexcrichton:
[...]

PR #301, opened 2022-08-15
Reorganize folder structure

Kylebrown9:
Implements #244

Folders are now organized in a consistent way that makes it clear which crates are generators and whether each generator is host/guest. Renames crates to match the new folder names and updates the associated usages and feature flags.


Kylebrown9:
This is ready for review, except the `README.md` changes, which I'm still working on.

alexcrichton:
This all seems reasonable enough to me but I think it would be good to get some other eyes on this as well. @peterhuene you're likely interested in this since you've been working a lot with the rust-to-component stuff recently. Additionally @sunfishcode or @kulakowski-wasm y'all may also be interested in weighing in on the particulars here.

sunfishcode:
The new naming here makes sense to me!

Kylebrown9:
Ready on my end.

I'd recommend keeping "releases" / version bumps separate from feature impl PRs.
So if we want to bump to `0.3.0` after this, which I think would be a good idea, I'll make a new PR after.

sunfishcode:
Sounds like there are no strong objections, so let's merge this, and iterate from here if there are any follow-up concerns.

Issue #299, opened 2022-08-08
Update wit-bindgen to parse the new multi-return syntax


kulakowski-wasm:
Now that https://github.com/WebAssembly/component-model/pull/69 landed, I'm planning to update wit-bindgen to parse this new syntax.

cc @sunfishcode and @lukewagner

PR #302, opened 2022-08-16
Bump version to 0.2.0

Kylebrown9:


sunfishcode:
wit-bindgen-cli isn't yet published on crates.io; is there more context here?

Kylebrown9:
We haven't had a tag since we made various syntactic changes (e.g. `func`), and I think it would be useful to have one before my PR to re-arrange the folder structure / command line args goes in. The idea of this PR is to update versioning so that our tree contents will match the new `0.2.0` tag.

sunfishcode:
There is another significant syntax change coming, as described in https://github.com/bytecodealliance/wit-bindgen/issues/299. Would it make sense to wait for that as well?

Kylebrown9:
I'd like to merge in #301 asap before things go in that make merging painful, so I'd like to go to `0.2.0` to record existing syntax changes now, then go to `0.3.0` for #301, then go to `0.4.0` when #299 goes in.

I think there are enough people relying on the prior syntax changes and the current CLI syntax (e.g. my employer SingleStore in it's quickstart tutorials) that it's worth giving them a version to point to for it. Plus, there's not much cost to tagging, and I think in lieu of a more detailed versioning/release process tagging for backwards-incompatible changes is a good idea.

alexcrichton:
I'm going to go ahead and merge this since there's not really a release process right now and the versions don't mean much anyway. Adding a tag doesn't seem too onerous and if users are relying on a tagged "thing" it seems like now's as good a time as any to make a tag before more changes are made.

PR #307, opened 2022-08-31
Remove `async func` support.

sunfishcode:
The `async` keyword has been removed from the [async proposal], in favor
of using plain functions that with `future` or `stream` types. See also
the [component-model PR to remove the syntax].

This removes the (now) old `async func` suppport from wit-bindgen, to make
it easier to develop the current Canonical ABI proposal.

[async proposal]: https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE/edit#slide=id.g1270ef7d5b6_0_111
[component-model PR to remove the syntax]: https://github.com/WebAssembly/component-model/pull/98


sunfishcode:
Unfortunately, the wit-bindgen demo depends on async infrastructure, so I've disabled running the demo in CI for now.

alexcrichton (approved):
I agree that this is the best way forward. I think it's best to get wit-bindgen into alignment with the currently-written spec and then subsequently re-add features like async as it fits within the current iteration of the component model. Otherwise trying to adapt everything here to the new model seems onerous. I do think that a good bit of what's removed here is going to come back, but it seems so likely to come back in a different form that it's not worth preserving outside of the git history.

It might be worth leaving this open for a day or two to ensure no one else has any objections, but otherwise seems reasonable to me to merge 👍

alexcrichton:
> Unfortunately, the wit-bindgen demo depends on async infrastructure, so I've disabled running the demo in CI for now.

The error I see in CI is:

```
../../static/main.ts(85,19): error TS2339: Property 'setWasmtimeAsync' does not exist on type 'Config'.
```

which I think should be easy to fix? I don't believe the demo use actual `async` function anywhere, that's just a config option which configures what the generated code looks like.

sunfishcode:
Ah, ok. I've now re-enabled the demo in CI, and just removed the `async` checkbox, which I think makes sense looking forward, as async will be implied by types rather than being a mode.

PR #312, opened 2022-09-13
Remove the special case for empty records.

sunfishcode:
WebAssembly/component-model#104 is closed, so remove the corresponding special case from wit-bindgen too.


alexcrichton (approved):

willemneal:
@alexcrichton can we get a 0.3.0 tag after this merge?

Issue #314, opened 2022-09-14
Reimplement host generators in terms of components


alexcrichton:
I'm opening this issue to document my thinking as a result of trying to resolve issues like https://github.com/bytecodealliance/wit-bindgen/issues/214, https://github.com/bytecodealliance/wit-bindgen/issues/34, and https://github.com/bytecodealliance/wit-bindgen/issues/201. I mentioned a number of issues with the current wit-bindgen architecture in #214, the biggest of which is that there was no clear location to slot a true "component" into the current hosts due to their centralized concept of using a singular wasm module for everything. To that end the best summary I can come up with is that the host-generators of wit-bindgen (`gen-host-wasmtime-{py,rust}` and `gen-host-js`) need to be reimplemented in terms of components, not core wasm modules.

In pivoting to components instead of core wasm modules I also believe that there's going to be to classifications of hosts: one where the host natively supports components and one where the host does not. For example `gen-host-wasmtime-rust` natively supports components, but `gen-host-wasmtime-py` (which uses Wasmtime's C API) and `gen-host-js` do not. This is not just a surface level distinction but will have deep implications on the capability of the code generators, namely hosts which do not have native component support will only support loading a single statically determined component. Hosts which natively support components (I'll just say Wasmtime from now on to refer to the Rust embedding of Wasmtime) will be able to consume any component that has a particular "shape", e.g. the same set of imports/exports with arbitrary structure internally.

## Hosts without component support 

I mentioned above but the major change for these hosts is that unlike today where they attempt to support any "shape" of core wasm module these hosts will only be able to generate bindings for a single statically known component. The rationale for this is that the imports/exports of a component, which are the most interesting from an embedder perspective, are only a small portion of the structure of a component which can have a lot going on internally within it. I don't think that `wit-bindgen` wants to get into the business of generating an entire component runtime for JS, for example, so instead these host generators will ingest a component and spit out bindings for just that one component.

At least for JS this lines up with how it was envisioned being used. In JS or web scenarios the intent is that there's a "thing compiled to wasm" that you want to run, so it's a statically known single module and there's no need for it to be able to instantiate multiple different shapes of modules at runtime. The Python generator for a Wasmtime host is mostly just proof-of-concept right now and I at least myself don't have preconcieved notions about how Python is most likely to be used.

[...]

PR #318, opened 2022-09-21
Synchronize the wit-bindgen type AST with the component model

alexcrichton:
This PR, with @kulakowski-wasm, brings the wit-bindgen AST up-to-date with the component model. Namely the changes implemented here are:

* Functions can now return multiple values. For now all code generators ignore the names of return values as I don't know how to implement that in any language other than JS, and JS can be implemented later. C notably has removed support for "splatting tuples" where it's exploded to multiple return values now that first-class multiple-returns are a thing. This more-or-less reverts https://github.com/bytecodealliance/wit-bindgen/pull/206.
* Variants now have an optional payload instead of a required payload. This reverts https://github.com/bytecodealliance/wit-bindgen/pull/220.
* The `unit` type is removed, partially reverting https://github.com/bytecodealliance/wit-bindgen/pull/204.
* The `result` type has its `ok` and `err` variants as now optional.

This PR was mostly test-driven where the core changes were made to the AST and ABI code and everything else fell out from basically getting `cargo test --workspace` working.

This is additionally one of the first steps I hope to make towards https://github.com/bytecodealliance/wit-bindgen/issues/314 since if Wasmtime is to consume components the generator and Wasmtime need to agree on encodings which requires `wasm-encoder` and such to be updated in `wit-component` which requires these AST changes.


peterhuene (approved):
Looks great! Thanks for updating `wit-component` as well.

PR #326, opened 2022-09-23
Move `RET_AREA` variables onto the stack.

sunfishcode:
Move `RET_AREA` variables out of static memory and onto the stack.

 - This ensures that the bindings are reentrant, which will be needed when the [wasi-threads] proposal makes threads available in wasm.

 - This makes it easier to use wit-bindgen to generate polyfill modules that import their linear memory.

[wasi-threads]: https://github.com/WebAssembly/wasi-threads

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -1347,7 +1347,10 @@ impl Generator for C {
             }
         }
 
-        if self.return_pointer_area_size > 0 {
+        // Declare a statically-allocated return area, if needed. We only do
+        // this for export bindings, because import bindings allocate their
+        // return-area on the stack.
+        if !self.in_import && self.return_pointer_area_size > 0 {
             uwrite!(
                 self.src.c,
                 "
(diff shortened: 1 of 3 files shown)


sunfishcode:
Another option here would be to add a flag to allow users to select whether to put `RET_AREA` on the stack or in static memory. I opted for just always putting it on the stack since that always works and it's simpler to avoid having flags, but I'm open to other options.

alexcrichton:
I agree there shouldn't be an option and we should just pick one, but this is technically UB right since it's returning a pointer to the local stack frame? If wasm were to implement some hypothetical canary hardening or similar where the frame was poisoned on exit this would trip that up, right?

Liamolucko:
> I agree there shouldn't be an option and we should just pick one, but this is technically UB right since it's returning a pointer to the local stack frame?

It's not actually *returning* the pointer, though - it's passed as an argument to an imported function, which exits before the function containing the return area and so the pointer is valid the whole time.

alexcrichton:
Ah I should clarify that I'm thinking about exports, although I do agree that for imports this should work. For exports this `*.wit` file:

```
test: func() -> (a: u8, b: u8)
```

generates:

```
$ cargo run guest rust --export foo.wit --rustfmt && cat bindings.rs
mod foo {
    #[export_name = "test: func() -> (a: u8, b: u8)"]
    unsafe extern "C" fn __wit_bindgen_foo_test() -> i32 {
        let (result0_0, result0_1) = <super::Foo as Foo>::test();
        let ptr1 = __FOO_RET_AREA.0.as_mut_ptr() as i32;
        *((ptr1 + 0) as *mut u8) = (wit_bindgen_guest_rust::rt::as_i32(result0_0)) as u8;
        *((ptr1 + 1) as *mut u8) = (wit_bindgen_guest_rust::rt::as_i32(result0_1)) as u8;
        ptr1
    }

    #[repr(align(1))]
    struct __FooRetArea([u8; 2]);
    static mut __FOO_RET_AREA: __FooRetArea = __FooRetArea([0; 2]);
    pub trait Foo {
        fn test() -> (u8, u8);
    }
}
```

where the pointer is returned.

@sunfishcode I believe your use case only deals with imports where this should work, but I think this may mean that both modes need to be simultaneously supported (`static` for exports, stack-local for imports)

sunfishcode:
Ah, right. I forgot to consider exports here. I've now updated the patch to use stack allocation only for imports.

alexcrichton (approved):

PR #327, opened 2022-09-27
add guest binding generator for TeaVM/Java

dicej:
This generator is for [teavm-wasi](https://github.com/fermyon/teavm-wasi), a friendly fork of TeaVM with support for WASI and the component model canonical ABI.

All the basic features are supported except for async functions, resources, and multi-value return.  The unsupported features shouldn't be hard to add in follow-up commits.

[Maven](https://maven.apache.org/) is required to run the tests.

### DIFF
--- a/crates/gen-guest-teavm-java/src/lib.rs
+++ b/crates/gen-guest-teavm-java/src/lib.rs
@@ -1,0 +1,1918 @@
+use heck::{ToLowerCamelCase, ToShoutySnakeCase, ToSnakeCase, ToUpperCamelCase};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Write,
+    iter, mem,
+    ops::Deref,
+};
+use wit_bindgen_core::{
+    uwrite, uwriteln,
+    wit_parser::{
+        abi::{AbiVariant, Bindgen, Bitcast, Instruction, LiftLower, WasmType},
+        Case, Docs, Enum, Field, Flags, FlagsRepr, Function, FunctionKind, Int, Interface, Record,
+        ResourceId, Result_, SizeAlign, Tuple, Type, TypeDefKind, TypeId, Union, Variant,
+    },
+    Direction, Files, Generator, Ns,
+};
+
+#[derive(Default)]
+pub struct TeaVmJava {
+    opts: Opts,
+    src: String,
+    stub: String,
+    aliases: HashMap<TypeId, Type>,
+    list_aliases: HashMap<TypeId, Type>,
+    sizes: SizeAlign,
+    anonymous_types: HashSet<TypeId>,
+    return_area_size: usize,
+    return_area_align: usize,
+    needs_cleanup: bool,
+}
+
+#[derive(Default, Debug, Clone)]
+#[cfg_attr(feature = "structopt", derive(structopt::StructOpt))]
+pub struct Opts {
+    /// Whether or not to generate a stub class for exported functions
+    #[cfg_attr(feature = "structopt", structopt(long))]
+    pub generate_stub: bool,
+}
+
+impl Opts {
+    pub fn build(&self) -> TeaVmJava {
+        TeaVmJava {
+            opts: self.clone(),
+            ..TeaVmJava::default()
+        }
+    }
+}
+
+impl TeaVmJava {
+    fn type_name(&mut self, iface: &Interface, ty: &Type) -> String {
+        self.type_name_with_qualifier(iface, ty, None)
+    }
+
+    fn type_name_with_qualifier(
+        &mut self,
+        iface: &Interface,
+        ty: &Type,
+        qualifier: Option<&str>,
+    ) -> String {
+        match ty {
+            Type::Bool => "boolean".into(),
+            Type::U8 | Type::S8 => "byte".into(),
+            Type::U16 | Type::S16 => "short".into(),
+            Type::U32 | Type::S32 | Type::Char => "int".into(),
+            Type::U64 | Type::S64 => "long".into(),
+            Type::Float32 => "float".into(),
+            Type::Float64 => "double".into(),
+            Type::Handle(_) => todo!("resources"),
+            Type::String => "String".into(),
+            Type::Id(id) => {
+                // TODO: can we rely on all relevant type aliases being known by the time we reach here?  If not,
+                // we'll need to refactor to ensure `type_name` isn't called until all aliases are known.
+                if let Some(ty) = self.aliases.get(id).copied() {
+                    self.type_name_with_qualifier(iface, &ty, qualifier)
+                } else if let Some(ty) = self.list_aliases.get(id).copied() {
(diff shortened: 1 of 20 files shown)


alexcrichton:
Bah it looks like `mypy` has changed and needs to be fixed on `main`, I'll try to look at that soon.

alexcrichton:
Otherwise though this is awesome! I'll try to take a look soon.

crates/gen-guest-teavm-java/src/lib.rs
```diff
@@ -0,0 +1,1918 @@
...
+                    Case {
+                        docs: Docs::default(),
+                        name: "err".into(),
+                        ty: result.err,
+                    },
+                ],
+            },
+            docs,
+        )
+    }
```
alexcrichton, line 542: Options/results are so commonly used and additionally aren't required to be named, so one thing I've done in the Python bindings for example is to create a helper `Result[T, E]` class. Would it be reasonable to use JS generics for something like this to have a `Result<T, E>` or `Option<T>` class?

Or perhaps for `Option<T>` would it perhaps be beneficial to specialize `Option<list<u8>>` for example to a `null` value but `Option<u32>` would generate a class wrapper?

crates/gen-guest-teavm-java/src/lib.rs
```diff
@@ -0,0 +1,1918 @@
...
+                name.to_upper_camel_case(),
+                operands[0]
+            )),
+
+            Instruction::ListCanonLower { element, realloc } => {
+                let op = &operands[0];
+                let (size, ty) = list_element_info(element);
+
+                // Note that we can only reliably use `Address.ofData` for elements with alignment <= 4 because as
+                // of this writing TeaVM does not guarantee 64 bit items are aligned on 8 byte boundaries.
```
alexcrichton, line 1292: We've had discussions historically about alignment in the canonical ABI and whether or not it should be required and traps if invalid. If this is too onerous it may be worth reopening the discussions to specify alignment and not require it like core wasm does.

crates/gen-guest-teavm-java/src/lib.rs
```diff
@@ -0,0 +1,1918 @@
...
+
+                results.push(array);
+            }
+
+            Instruction::StringLower { realloc } => {
+                let op = &operands[0];
+                let bytes = self.locals.tmp("bytes");
+                uwriteln!(
+                    self.src,
+                    "byte[] {bytes} = ({op}).getBytes(StandardCharsets.UTF_8);"
```
alexcrichton, line 1343: Are strings natively encoded as utf-8 in java?

crates/gen-guest-teavm-java/src/lib.rs
```diff
@@ -0,0 +1,1918 @@
...
+                );
+            }
+        }
+    }
+
+    fn return_pointer(&mut self, _iface: &Interface, size: usize, align: usize) -> String {
+        self.gen.return_area_size = self.gen.return_area_size.max(size);
+        self.gen.return_area_align = self.gen.return_area_align.max(align);
+        "RETURN_AREA".into()
+    }
```
alexcrichton, line 1722: One change we just landed to the C & Rust generators was that for imports the return area was stored on the stack, but I'm not sure if the equivalent would work in java. That being said it may be worth considering if you'd like.

alexcrichton (approved):
Looks great to me, thanks for implementing this! The python issues should be fixed on `main` so a rebase should turn up any issues in CI.

Otherwise some other items could be:

* Mind adding support to the web demo to see the generated java?
* Would you be up for implementing a few of the `tests/runtime/*/wasm.java` tests? I think it'd be good to test at least a little with the runtime rather than only that the compiler passes

crates/gen-guest-teavm-java/src/lib.rs
dicej: The JRE can technically encode them however it likes, and I believe the `String` implementation in OpenJDK has changed several times, but since `char` is 16-bits, the "most native" encoding is UTF-16, and TeaVM uses a `char[]` to implement `String`.

crates/gen-guest-teavm-java/src/lib.rs
dicej: Yeah, I saw that.  There's no shadow stack in TeaVM currently (and I don't know if it would make sense given you can't take the address of a stack-allocated thing in Java), so literally putting it on the stack isn't an option, but we could certainly allocate it on-the-fly on the GC'd heap.

crates/gen-guest-teavm-java/src/lib.rs
alexcrichton: Ah ok, I'm wondering if we can get a good use case for exercising the utf16 option of guests or similar, and it sounds like maybe? That's fine to happen at a later date though, this repo isn't prepared to actually do that yet.

crates/gen-guest-teavm-java/src/lib.rs
dicej: Correction: TeaVM _does_ have a shadow stack, since the GC needs to scan it for roots when collecting.  I don't know enough about how it works yet to safely allocate a return area on it, and I suspect TeaVM expects only heap references on that stack, so I think it would be a bad fit for this kind of thing.

Also, forget my idea of allocating the return area on the GC'd heap since TeaVM may consider it garbage once there are no more stack references to it.

Anyway, the static allocation works for now, and we can revisit if/when we have to deal with reentrancy.

dicej:
I'll investigate the latest CI failures.  Apparently the `actions/setup-java` doesn't include Maven on Windows?  Or maybe its something more subtle.

crates/gen-guest-teavm-java/src/lib.rs
dicej: I wouldn't say it's onerous.  If it turns out the extra copies for `list<u64>` values create real-world performance problems we can either revisit that discussion or I can see about modifying TeaVM to align `long[]` as desired.

crates/gen-guest-teavm-java/src/lib.rs
dicej: One more thought: I just realized that the C and Rust generators only use the stack for imports (which is what you said, now that I read more carefully).  So yes, we could at least use an on-the-fly heap allocation for import return areas, either using the GC'd heap or the manually managed one.  The static allocation is best for performance as long as reentrancy isn't a concern, though.  And as with C and Rust, static allocation is the only option for exports.

alexcrichton:
Looks great to me! Was there anything else you wanted to tackle before merging?

dicej:
> Looks great to me! Was there anything else you wanted to tackle before merging?

Still working on using generics for tuples, options, and results, as you suggested, and I was going to implement a few more runtime tests after that.  If you want to go ahead and merge now, I can do that work in follow-up PRs.  Either way works for me.

alexcrichton:
Ok I think this is good to do those as follow-ups so I'll merge this. If you'd like feel free to open an issue as well to track further improvements.

PR #347, opened 2022-10-05
Guest C custom sections containing component types

pchickey:
Adds the custom section component types generation to the guest C generator.

Also tests that TeaVM guests can be translated to components as well. In the TeaVM case, we explicitly pass the imports and export interface to the ComponentEncoder.

I added stub functions to the WASI adapter for the additional functions required by wasi-libc and the TeaVM guests.

### DIFF
--- a/crates/test-helpers/build.rs
+++ b/crates/test-helpers/build.rs
@@ -251,15 +251,35 @@ fn main() {
                 panic!("failed to build");
             }
 
+            let out_wasm = out_dir.join("target/generated/wasm/teavm-wasm/classes.wasm");
+
             wasms.push((
                 "java",
                 test_dir.file_stem().unwrap().to_str().unwrap().to_string(),
-                out_dir
-                    .join("target/generated/wasm/teavm-wasm/classes.wasm")
-                    .to_str()
-                    .unwrap()
-                    .to_string(),
+                out_wasm.to_str().unwrap().to_string(),
             ));
+
+            let imports = [Interface::parse_file(test_dir.join("imports.wit")).unwrap()];
+            let interface = Interface::parse_file(test_dir.join("exports.wit")).unwrap();
+
+            // Validate that the module can be translated to a component, using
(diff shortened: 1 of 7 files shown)


alexcrichton:
Nice! This looks great to me 👍

crates/test-helpers/build.rs
```diff
@@ -149,6 +151,24 @@ fn main() {
                 test_dir.file_stem().unwrap().to_str().unwrap().to_string(),
                 out_wasm.to_str().unwrap().to_string(),
             ));
+
+            // The "invalid" test doesn't actually use the rust-guest macro
+            // and doesn't put the custom sections in, so component translation
+            // will fail.
+            if test_dir.file_stem().unwrap().to_str().unwrap() != "invalid" {
```
alexcrichton, line 158: I think with my suggestion above about how to force linking the object file it should resolve this?

alexcrichton (approved):

crates/test-helpers/build.rs
pchickey: It turns out to be unnecessary now even without that fix, I just forgot to try it again

pchickey:
https://github.com/bytecodealliance/wit-bindgen/pull/347/commits/31b737c7cce9ae31248666f9a6afd81999ccbc1e cc @dicej, manually passing in the interfaces used by the TeaVM guests in test-helpers/build.rs is good enough for now, at some point it would be nice if the code generator put them in custom sections like we do for Rust and C, but its not urgent.

dicej:
> [31b737c](https://github.com/bytecodealliance/wit-bindgen/commit/31b737c7cce9ae31248666f9a6afd81999ccbc1e) cc @dicej, manually passing in the interfaces used by the TeaVM guests in test-helpers/build.rs is good enough for now, at some point it would be nice if the code generator put them in custom sections like we do for Rust and C, but its not urgent.

Thanks for the heads up.  It shouldn't be hard to do in TeaVM -- I'll take a look when I have some time.

PR #348, opened 2022-10-05
Remove special case for "invalid" test

alexcrichton:
While we don't actually use the `imports.wit` it should be possible to still generate the code for it as-if we used it to allow running the test through `wit-component`.


pchickey (approved):
Thanks, I didn't think of this

PR #349, opened 2022-10-05
Compile test cases with `wasm32-wasi` again

alexcrichton:
This commit is the next step in integrating `wasm32-wasi`, `wit-bindgen`, tests, and components all together. Tests are now again compiled with `wasm32-wasi` and use a repo-specific adapter module (with support from #338) to support transforming the final module into an actual component.

In supporting this feature the support from #331 is refactored into a new `extract` Rust module so the functionality can be shared between the ingestion of the main module as well as ingestion of adapter modules. Adapter modules now are also supported on the CLI as a standalone file without having to specify other options.

Note that the actual `wasi_snapshot_preview1.wasm` adapter is non-functional in this commit and doesn't do anything fancy. The tests in this repository don't really need all that much and I suspect all we'll really need to implement is `fd_write` for fd 1 (as that's stdout).

### DIFF
--- a/crates/wit-component/src/extract.rs
+++ b/crates/wit-component/src/extract.rs
@@ -1,0 +1,72 @@
+use crate::decode_interface_component;
+use anyhow::{Context, Result};
+use wit_parser::Interface;
+
+/// Result of extracting interfaces embedded within a core wasm file.
+///
+/// This structure is reated by the [`extract_module_interfaces`] function.
+#[derive(Default)]
+pub struct ModuleInterfaces {
+    /// The core wasm binary with custom sections removed.
+    pub wasm: Vec<u8>,
+
+    /// Imported interfaces found in custom sections.
+    pub imports: Vec<Interface>,
+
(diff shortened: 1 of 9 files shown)


crates/wit-component/src/extract.rs
```diff
@@ -0,0 +1,72 @@
...
+                    ret.exports.push(i);
+                }
+            }
+            _ => {}
+        }
+    }
+
+    // TODO: should remove the custom setions decoded above from the wasm binary
+    // created here, and bytecodealliance/wasmparser#792 should help with that
+    // to make the loop above pretty small.
```
pchickey, line 68: yessss

pchickey (commented):
Is it worth shipping a new wasmparser rn to resolve this todo?

alexcrichton:
In the interest of decoupling I'll work on that in parallel, github actions is having a fit right now and not actually running so it'll take a bit to land the wasmparser changes and for now the custom sections are otherwise harmless.

pchickey (approved):

PR #352, opened 2022-10-05
Add a `raw_strings` option.

sunfishcode:
Add a `raw_strings` option for the Rust guest bindings, to use `&[u8]` and `Vec<u8>` instead of `&str` and `String` for representing strings in bindings. This will allow polyfill modules that don't already have a `&str` to avoid doing a `str::from_utf8` to obtain one, which is redundant since the adapter will already be doing UTF-8 validation.

### DIFF
--- a/crates/gen-guest-rust/src/lib.rs
+++ b/crates/gen-guest-rust/src/lib.rs
@@ -57,6 +57,13 @@ pub struct Opts {
     /// If true, code generation should avoid any features that depend on `std`.
     #[cfg_attr(feature = "clap", arg(long))]
     pub no_std: bool,
+
+    /// If true, code generation should pass borrowed string arguments as
+    /// `&[u8]` instead of `&str`. Strings are still required to be valid
+    /// UTF-8, but this avoids the need for Rust code to do its own UTF-8
+    /// validation if it doesn't already have a `&str`.
+    #[cfg_attr(feature = "clap", arg(long))]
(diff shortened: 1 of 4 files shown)


alexcrichton:
Is this a case where the adapters could use `str::from_utf8_unchecked` instead?

sunfishcode:
That would probably work, though technically `from_utf8_unchecked` has UB if the bytes aren't UTF-8, and the polyfill's input doesn't guarantee that. Granted, passing ill-formed strings to wit interfaces will trap, so it's not going to be useful to do anyway, but it seemed better to avoid the potential UB.

alexcrichton:
Ok seems reasonable enough, could you add some codegen tests that exercise this option though? An example is a test [like this](https://github.com/bytecodealliance/wit-bindgen/blob/9649d344f39fa44876a5ba35d504fd4a9e758fd7/crates/gen-host-wasmtime-rust/tests/codegen.rs#L41-L54) for a wasmtime-specific option.

sunfishcode:
> Ok seems reasonable enough, could you add some codegen tests that exercise this option though? An example is a test [like this](https://github.com/bytecodealliance/wit-bindgen/blob/9649d344f39fa44876a5ba35d504fd4a9e758fd7/crates/gen-host-wasmtime-rust/tests/codegen.rs#L41-L54) for a wasmtime-specific option.

Ok, tests added.

PR #354, opened 2022-10-06
Build emits components; disable running all host generators

pchickey:
Change the test-helpers/build.rs script to write Components to the disk, and specify those files in the generated WASMS variable. This means all tests will use components.

Disable all of the host generator runtime tests. All of these generators and runtime tests are oriented to use canonical ABI modules, not components, as their input. As we switch to components, we will need to rewrite big chunks of each generator.

We can still expect the generators to still create code that compiles, so I am leaving the codegen tests intact.


alexcrichton:
Instead of disabling the existing tests, how about creating a parallel `COMPONENTS` array and then each host generator picks either `WASMS` or `COMPONENTS` to test? That way in the interim while the hosts are getting reimplemented the other hosts should continue to work. Eventually we'd drop the `WASMS` array once hosts are migrated of course.

pchickey:
Good point. Replaced with #359.

PR #355, opened 2022-10-06
rewrite gen-host-wasmtime-rust to use Wasmtime's component runtime

pchickey:
I rewrote the `wit-bindgen-gen-host-wasmtime-rust` crate to emit code which executes components, using the `wasmtime::component::` runtime. Previously, this crate emitted code which executed canonical ABI core modules, using the "core" wasmtime runtime.

There's not a ton to say about the rewrite. The tests show that it works.

* The wasmtime_runtime_tests macro now passes the component path to the code generator, rather than the core module path.
* Each `test/runtime/host.rs` had some minor trimming to get rid of old concepts that don't exist anymore: export structs arent part of the store context, this was required for handles iirc.
* I had to move some logic about adding the `#[component]` attr and `ComponentType, Lift, Lower` derivations down into `gen-rust-lib` behind a flag. This is a bit of a hack, but reduces code duplication.

## Ownership

I departed from the previous conventions in this crate to follow the behaviors available from the Wasmtime component runtime:

* Exports are passed `AllBorrowed` style (e.g. `&[&str]`) parameters, and return owned (e.g. `Vec<String>`) results.
* Imports are passed owned parameters, and return owned results.

When it becomes possible to pass imports any sort of borrowed parameter, we can make that improvement to this crate. This may end up saving a copy.

## No more `host-wasmtime-rust`

All significant code in the `wit-bindgen-host-wasmtime-rust` crate was a "runtime" for the previous implementation of components on top of core modules. This is all replaced by the Wasmtime implementation, so I was able to turn this code into a slim re-export of the proc macros, and the crate dependencies used by the proc macros (`pub use {anyhow, wasmtime, tracing};`).

## Wasmtime dependency

This PR requires a dependency on Wasmtime's `main` branch, at least until the 3.0.0 release is available. In particular, it requires the latest `wasmparser` https://github.com/bytecodealliance/wasmtime/pull/5010, and these fixes to the `wasmtime::component::flags!` macro https://github.com/bytecodealliance/wasmtime/pull/5030

## Tests

In the process, I had to fix a bunch of tests:

* I shortened `many_arguments` to only use 16 arguments (from 20), because that is the (arbitrary) limit on the TypedFunc argument tuple size.
* I translated `union` to have a host wasmtime rust implementation
* I fixed the behavior of `invalid` to reflect spec changes that out of bounds integers are truncated. I moved some testing of unaligned structures to this test as well, to assert that they trap.
* I am suspicious about some errors in the `host.py` test harness for `unions`, but they are corrected now at any rate.

Previously, the linker made WASI preview 1 imports available to canonical ABI core modules. Instead, we now use the `crates/wasi_preview_1` adapter, which will trap on calls to `fd_write` and other effectful functions. I did end up filling in trivial non-trapping implementations for `{environ,args}_{get,sizes_get}` because various guests call those as part of initialization.


crates/gen-rust-lib/src/lib.rs
```diff
@@ -492,11 +492,22 @@ pub trait RustGenerator {
...
+        derive_component: bool,
     ) {
         let info = self.info(id);
         for (name, mode) in self.modes_of(iface, id) {
             let lt = self.lifetime_for(&info, mode);
             self.rustdoc(docs);
+
+            if derive_component {
+                self.push_str("#[derive(wasmtime::component::ComponentType)]\n");
+                if self.lifetime_for(&info, mode).is_none() {
```
alexcrichton, line 504: I think `lt` can be used from above instead of calling `lifetime_for` again?

crates/gen-host-wasmtime-rust/tests/runtime.rs
```diff
@@ -11,43 +14,33 @@ fn default_config() -> Result<Config> {
...
     ) -> Result<(T, Instance)>,
-) -> Result<(T, Store<Context<I, E>>)> {
+) -> Result<(T, Store<Context<I>>)> {
     let engine = Engine::new(&default_config()?)?;
-    let module = Module::from_file(&engine, wasm)?;
+    let module = Component::from_file(&engine, wasm)?;
 
     let mut linker = Linker::new(&engine);
     add_imports(&mut linker)?;
-    wasmtime_wasi::add_to_linker(&mut linker, |cx| &mut cx.wasi)?;
```
alexcrichton, line 43: This is removing wasi entirely for tests, which I think means that tests can no longer have debug prints within them, right? Would you be ok updating the `preview1.wasm` adapter used to delegates writes on fd 1 to a simple `log: func(s: &str)` function?

crates/gen-rust-lib/src/lib.rs
```diff
@@ -602,6 +638,14 @@ pub trait RustGenerator {
         for (name, mode) in self.modes_of(iface, id) {
             let name = name.to_upper_camel_case();
             self.rustdoc(docs);
+            if let Some(derive_component) = derive_component {
+                self.push_str("#[derive(wasmtime::component::ComponentType)]\n");
+                if self.lifetime_for(&info, mode).is_none() {
```
alexcrichton, line 643: Similar to above I think the `let lt = ...` call below can be reordered to use the result here too

crates/host-wasmtime-rust/src/region.rs
```diff
@@ -1,312 +0,0 @@
...
-use crate::{Endian, GuestError, Le};
-use std::collections::HashSet;
-use std::convert::TryInto;
-use std::marker;
-use std::mem;
-use wasmtime::Trap;
-
-// This is a pretty naive way to account for borrows. This datastructure
-// could be made a lot more efficient with some effort.
-pub struct BorrowChecker<'a> {
```
alexcrichton, line 11: I'm so happy we get to remove this

tests/runtime/invalid/wasm.rs
```diff
@@ -17,51 +17,163 @@ extern "C" {
...
+                records.as_abi().1,
+                f32s.as_abi().0,
+                f32s.as_abi().1,
+                f64s.as_abi().0,
+                f64s.as_abi().1,
+                strings.as_abi().0,
+                strings.as_abi().1,
+                lists.as_abi().0,
+                lists.as_abi().1,
+            );
```
alexcrichton, line 176: I think technically to fully test this we'd have to split out these two tests into N different tests, one for each type. Otherwise anything which checks alignment for the `u16s` correctly would trap and then not make it to anything else.

Additionally only the Rust host generator actually checks alignment at this time, changes would be needed for JS & Python to implement the current component model semantics which is to check alignment. I want to get to that eventually but it means that the test here is going to have different behavior in wasmtime and in other hosts for now. 

Given all that maybe this test could just be commented out for now with an issue filed to come back and fix it?

crates/gen-host-wasmtime-rust/src/lib.rs
```diff
@@ -1,31 +1,19 @@
...
-    needs_char_from_i32: bool,
-    needs_invalid_variant: bool,
-    needs_validate_flags: bool,
-    needs_raw_mem: bool,
-    needs_bad_int: bool,
-    needs_copy_slice: bool,
-    needs_buffer_glue: bool,
-    needs_le: bool,
-    needs_custom_error_to_trap: bool,
-    needs_custom_error_to_types: BTreeSet<String>,
```
alexcrichton, line 27: Could the `Opts::custom_error` field get removed as well?

Also if `tracing` isn't implemented yet mind removing that too?

alexcrichton (commented):
This looks great to me, thanks again for this! I'm quite pleased at how simple the Wasmtime generator ended up being (or at least removing of now-vestigal complexity).

> There's not a ton to say about the rewrite. The tests show that it works.

I'm glad the tests worked well for this!

> I departed from the previous conventions in this crate to follow the behaviors available from the Wasmtime component runtime:

Can you file an issue on merging to improve this? I don't want to lose track of this but I do think that what you've implemented here is suitable for the time being. I think it's best to learn more about embeddings and what traits are expected given to guide the design here rather than the prior implementation which "maximized borrows where possible". 

I do think we'll need something in this area to make this truly viable, especially to avoid copies with parameters like `(list u8)` given to a print-like function. This is all fine to defer to later though.

crates/gen-rust-lib/src/lib.rs
pchickey: thanks!

crates/gen-host-wasmtime-rust/tests/runtime.rs
pchickey: I can go add that in, but maybe as a follow on PR. I don't think tests currently have debug prints because if they did, they'd trap in the current implementation.

crates/gen-host-wasmtime-rust/src/lib.rs
pchickey: Yes, will get rid of custom error.

Tracing is actually implemented for exports, but not imports - trivial to add it to imports, and configure the tests to turn it on.

pchickey:
Filed https://github.com/bytecodealliance/wasmtime/issues/5050 to capture the import func param ownership issue.

tests/runtime/invalid/wasm.rs
pchickey: https://github.com/bytecodealliance/wit-bindgen/issues/370

pchickey:
https://github.com/bytecodealliance/wasmtime/pull/5051 fixed the problems that cropped up when merging with main

crates/gen-host-wasmtime-rust/tests/runtime.rs
pchickey: https://github.com/bytecodealliance/wit-bindgen/pull/371

PR #357, opened 2022-10-06
Fix hashing and equality in `wit-component` encoding

alexcrichton:
The `params_equal` closure accidentally forgot to take into account parameter lists of different lengths which would mean that everything would be equal to the empty list. While reviewing this code I also noticed a number of cases where the `Hash` needed to improve to ensure that lengths of lists, variant numbers, and `None` cases all are hashed to improve the quality of the hash.

Long-term my hope is that this need of `wit-component` will go away as "world files" are pushed deeper into `wit-parser` which will push this global-level deduplication into the type resolver instead of needing to duplicate it here manually.


alexcrichton:
This should fix the nondeterministic test failures we've been seeing on CI

PR #362, opened 2022-10-07
Fix warnings in lex.rs

kulakowski-wasm:


crates/wit-parser/src/ast/lex.rs
```diff
@@ -135,17 +135,6 @@ impl<'a> Tokenizer<'a> {
         Ok(ret)
     }
 
-    pub fn parse_str(&self, span: Span) -> Result<String> {
```
alexcrichton, line 138: @fibonacci1729 is this a function you'd like to keep around for now?

(if so @kulakowski-wasm would you be ok adding `#[allow(dead_code)]` here with a comment?)

fibonacci1729: Yeah, i'll be using this in the near future.

kulakowski-wasm: ofc happy either way!

alexcrichton (approved):

PR #363, opened 2022-10-07
Refactor and reorganize how testing works

alexcrichton:
This commit is a large refactor of how testing works in this repository for wit-bindgen and hosts/guests to simplify the flow of tests and make it easier to modify and less monolithic. Previously tons of logic was located in a `test_helpers` macro crate but now that large crate has been split up to delegate more work to each individual test. Some common functionality is still present in a `test_helpers` crate but it's more clear that the `test_helpers::codegen_test` macro is purely responsible for simply expanding a glob at compile time and delegating to a macro-per-test.

The tests of Rust & other languages are now also more uniform where everything "just invokes a macro" where Rust's macros further invoke proc-macros for the respective generators and other languages delegate to shared functionality for "generate code" then "run verification program".

Overall I'm hoping this makes the testing easier to follow and understand, and in upcoming changes for components it will make it easier to change each generator, individually, over to components ideally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -685,6 +685,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "guest-rust-test-macro"
+version = "0.0.0"
+dependencies = [
+ "heck",
+ "proc-macro2",
+ "quote",
+ "wit-parser",
+]
+
+[[package]]
 name = "hashbrown"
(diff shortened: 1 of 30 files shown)


kulakowski-wasm:
This is definitely clearer for me! 👍

pchickey (approved):

Issue #333, opened 2022-09-29
Bump version to 0.3.0 since tag 0.2.0 is out? 


Mossaka:
We have the main branch versioned at `0.2.0` while released a tag `v0.2.0` last month. This might cause some confusions on what version of the wit-bindgen CLI is using.

alexcrichton:
This is done now, so closing.

PR #365, opened 2022-10-08
Bump version to 0.3.0

Mossaka:
Feel like #327 is a significant feature added to wit-bindgen. In addition, we already have tag `0.2.0` out so I am bumping the version to `0.3.0`.  See #333 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2021"
-version = "0.2.0"
+version = "0.3.0"
 
 [workspace.dependencies]
 anyhow = "1.0.65"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -32,27 +32,28 @@ wasmparser = "0.92.0"
...
+wit-parser = { path = 'crates/wit-parser', version = '0.3.0' }
+wit-component = { path = 'crates/wit-component', version = '0.3.0' }
 
 [[bin]]
 name = "wit-bindgen"
 test = false
 
 [dependencies]
 anyhow = { workspace = true }
-clap = { workspace = true }
```
alexcrichton, line 55: I think the changes in the `[dependencies]` and `[build-dependencies]` sections may not be needed?

alexcrichton (approved):

Cargo.toml
Mossaka: you're right~

PR #368, opened 2022-10-10
feat: added build info to `--version`

Mossaka:
Inspired by cargo, this pr adds some build-time information for the wit-bindgen cli. 

image


alexcrichton:
Having extra information here seems reasonable to me, but the `vergen` dependency here seems quite large proportional to the value and amount of information it's bringing in. Would it be possible to slim this down perhaps?

Also, to confirm, the date printed is the date of the commit, not the date of the build?

Mossaka:
> Also, to confirm, the date printed is the date of the commit, not the date of the build?

yes

> Would it be possible to slim this down perhaps?

That's a good point. I just pushed a new commit that uses `git` command to fetch commit sha and date. This is inspired by  https://github.com/rust-lang/cargo/blob/master/build.rs

src/bin/wit-bindgen.rs
```diff
@@ -1,12 +1,21 @@
...
 use wit_parser::Interface;
 
+lazy_static! {
+    pub static ref VERSION: String = build_info();
+}
+
+/// Helper for passing VERSION to opt.
+fn version() -> &'static str {
+    &VERSION
+}
```
alexcrichton, line 15: This is a bit of an iteration beyond what Cargo does but I think this entire `VERSION` string can be built in the build script? Could that be done to avoid needing `lazy_static!` and building it at runtime here?

Additionally I don't think this build will work when `.git` is missing or the `git` executable itself because then the env vars aren't set by the build script.

Mossaka: This is really a good point!

I build the enriched version information in `build.rs` and thus got rid of `lazy_static!`. For the second comment, thanks for pointing out a error handling case! What I did is to use `CARGO_PKG_VERSION` as the version info if `.git` is missing or the `git` executable itself is missing.

alexcrichton (approved):

PR #369, opened 2022-10-11
Embed named type information in components

alexcrichton:
In writing support for the JS host generator recently I found myself unable to take a `wit-component`-generated component and generate bindings for it, namely because the names of types were not present so it wasn't clear what records and such would be named.

Currently `wit-component` has a solution to this in the form of an "interface encoder" which produces a `*.wasm` file which is a theoretical bijection between a component and a `*.wit` interface as defined today. This `wit2wasm` and `wasm2wit` tooling, however, was all geared towards individual interfaces and didn't clearly fit within a full-fledged component produced by `wit-component`. In this PR I've decided to smash all of this together and get everything working in concert at the same time.

Specifically in this PR I've done an large number of changes to `wit-component` along the lines of:

* The `wit2wasm` tooling is not gone and replaced by `wit-component --types-only`. This was a large shift in the direction of `wit2wasm` as well as it wasn't merely moved but was instead integrated with `wit-component`'s idea of "worlds" where multiple interfaces are expressed simultaneously within a component (e.g. imports, exports, and the default interface).

* The `--types-only` output, taken from `wit2wasm` is updated to support simultaneously describing multiple interfaces. Namely imported interfaces are instance imports, exported interfaces are exports of an instance type, and the default export exports types representing all the items in the default interface.

* The default output of `wit-component` was updated to always embed type naming information. For example imported instances now also export type names on each individual instance. Exported instances export type names in addition to actual functions. This means that components always have named type information in them now when run through `wit-component`.

* The `tests/interfaces.rs` test for components was overhauled to look more like `tests/components.rs`: multiple interfaces are supported. The test itself is still testing the same thing, though, namely it produces a `types-only` component which is verified against `types_only.wat`. The `types_only.wat` component is then additionally decoded using `wasm2wit` and the input `*.wit` files are asserted to all be correctly inferred from the input. Finally a second test was added to additionally produce a `component.wat` which is not `types-only` to check that interfaces can be inferred from full components in addition to types-only components.

The "types only" interface of components may not be long-lived in `wit-component` depending on how things shape up over time. This isn't too bad though because the support for this mode is quite minimal. Otherwise there's a lot of effective churn in this PR largely around tests but the changes to Rust source files should all hopefully be more straightforward.

One worry I have about this PR is that this is much more strongly leveraging a part of `wit-component` which isn't necessarily "battle-tested" yet. For example the mapping from a set of `*.wit` interfaces onto a component should be a bijection where one can always be generated from the other (modulo documentation at this time) but there are a number of features of `*.wit` which won't be recovered, for example type aliases. I would ideally like to write a fuzzer soon to fuzz-generate an `Interface` and round-trip that through a component to ensure the same interface pops out. This may not be the best thing to prioritize so early on though but I am still fearful that this can otherwise create really subtle and hard-to-understand bugs for users if the low-levels of tooling here have basic issues in them.

For now though this should be sufficient to get the JS generator back on track since there's type information present and usable for generating human-readable and meaningful bindings.


crates/wit-component/src/decoding.rs
```diff
@@ -1,190 +1,196 @@
...
                 }
-                Chunk::NeedMoreData(_) => unreachable!(),
+                Payload::ComponentExportSection(s) if depth == 1 => {
+                    for export in s {
+                        let export = export?;
+                        let prev = exports.insert(export.name, export);
+                        assert!(prev.is_none());
+                    }
+                }
+                _ => {}
```
pchickey, line 62: Thanks, the improved parser abstraction really improves this implementation

crates/wit-component/src/decoding.rs
```diff
@@ -1,190 +1,196 @@
...
+/// expected shape. At this time not all component shapes are supported here.
+pub fn decode_interface_component(bytes: &[u8]) -> Result<ComponentInterfaces<'_>> {
+    let info = ComponentInfo::new(bytes)?;
+    let mut imports = IndexMap::new();
+    let mut exports = IndexMap::new();
+
+    for (name, ty) in info.imports.iter() {
+        // Imports right now are only supported if they're an import of an
+        // instance and an instance is self-describing in the interface that it
+        // represents.
```
pchickey, line 123: Can you clarify what self-describing means here - I don't think any of the instance imports presently available in the  binary format are not self-describing? or is this about future *-imports?

pchickey (approved):
Thanks, I like this improvement.

crates/wit-component/src/decoding.rs
alexcrichton: Oh I just meant that the type of the import is sort of the serialization of the `Interface` type, I reworded a bit to capture that.

PR #373, opened 2022-10-13
js: Take a component as input, not interfaces 

alexcrichton:
This commit is the implementation of https://github.com/bytecodealliance/wit-bindgen/issues/314 for the JS host generator.
The commit here covers basically everything in that issue for JS except
it generates a slightly different structure of the JS output. Otherwise
the main highlights are:

* The JS host generator no longer implements the core `Generator` trait
  since it doesn't really fit in the component-as-input world. For now I
  created an `InterfaceGenerator` trait since similar functionality is
  used just not precisely the same. I expect that this will get iterated
  on over time as worlds and other host generators take shape.

* The `wasmtime-environ` crate from Wasmtime itself, typically a "private"
  dependency of Wasmtime, is used to parse the input component and
  generate an `instantiate` function in JS. Wasmtime does all the heavy
  lifting of creating a linear list of initializers for the component
  and this empowers the generator to simply generate JS that is the list
  of initializers.

* The `wit-component` crate is used to "extract" the `Interface`
  descriptions from the input component. This is used to generate type
  information for TypeScript as well as types for lifting/lowering.
  Internally a correlation is done from a lowering/lifting to an
  `Interface` function to connect the dots when instantiating a component.

Lots of pieces needed updating here such as the runtime tests, the
`wit-bindgen` CLI tool, and the demo web page. These are all updated for
the new structure of the JS host generator.

Overall this surprisingly went much more smoothly than I expected. With
`wit-bindgen` being able to extract `Interface` representations from a
component most of the prior JS host code generation was able to be
reused. Along the way I also felt that the addition of "worlds" would
have relatively obvious insertion points and would be relatively easily
handled in all the places.

The demo and runtime tests are proof enough to me at least that this all
works internally, and this feels like a solid foundation to iterate from
with the addition of worlds and continued support for JS hosts.


alexcrichton:
Procedurally this is built on https://github.com/bytecodealliance/wit-bindgen/pull/372 and additionally depends on https://github.com/bytecodealliance/wasmtime/pull/5053, so will want to hold off on merging until those are both in.

crates/gen-host-js/tests/helpers.ts
```diff
@@ -0,0 +1,23 @@
...
+import { argv, stdout } from 'node:process';
+
+// This is a helper function used from `host.ts` test in the `tests/runtime/*`
+// directory to pass as the `instantiateCore` argument to the `instantiate`
+// function generated by `wit-bindgen`.
+//
+// This function loads the module named by `path` and instantiates it with the
+// `imports` object provided. The `path` is a relative path to a wasm file
+// within the generated directory which for tests is passed as argv 2.
+export async function loadWasm(path: string, imports: any) {
```
pchickey, line 13: dumb question because I do not know Typescript: I'd expect this to have a type annotation saying it returns a `WebAssembly.Instance`, but I guess that is just an inferred type?

crates/gen-host-js/tests/runtime.rs
```diff
@@ -31,8 +28,9 @@ fn execute(name: &str, wasm: &Path, ts: &Path, imports: &Path, exports: &Path) {
     };
 
     fs::copy(ts, dir.join("host.ts")).unwrap();
-    fs::copy("tests/helpers.d.ts", dir.join("helpers.d.ts")).unwrap();
-    fs::copy("tests/helpers.js", dir.join("helpers.js")).unwrap();
+    // fs::copy("tests/helpers.d.ts", dir.join("helpers.d.ts")).unwrap();
+    // fs::copy("tests/helpers.js", dir.join("helpers.js")).unwrap();
```
pchickey, line 32:
```suggestion
```

crates/gen-host-js/tests/runtime.rs
```diff
@@ -31,8 +28,9 @@ fn execute(name: &str, wasm: &Path, ts: &Path, imports: &Path, exports: &Path) {
     };
 
     fs::copy(ts, dir.join("host.ts")).unwrap();
-    fs::copy("tests/helpers.d.ts", dir.join("helpers.d.ts")).unwrap();
-    fs::copy("tests/helpers.js", dir.join("helpers.js")).unwrap();
+    // fs::copy("tests/helpers.d.ts", dir.join("helpers.d.ts")).unwrap();
```
pchickey, line 31:
```suggestion
```

pchickey (commented):
review still in progress, just pushing these easy ones

pchickey (approved):
This is awesome, lets ship it. I spent some time going through this with @guybedford and we satisfied a bunch of curiosity we had about why the generated code looks like it does.

crates/gen-host-js/tests/helpers.ts
alexcrichton: I believe it's inferred, yeah, although I'm no typescript-whiz myself!

PR #374, opened 2022-10-13
wasmtime: Remove lifetime and type parameter from exports

alexcrichton:
This commit removes the type parameter since it's unused at this time. It was historically used for handles but that's under development again so it's no longer needed. This commit then additionally removes the lifetime parameter from the structure, opting to store `Func` instead of `TypedFunc` to avoid the need to name types and their lifetime parameters. Functions internally then use `TypedFunc::new_unchecked` to re-acquire the typed view with appropriate anonymous lifetimes in parameters.


pchickey (approved):
Yep, this is a better design.

PR #380, opened 2022-10-18
guest-c: use string include

guybedford:
Using a string include was needed for compilation with clang to work out.

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -1166,7 +1166,7 @@ impl Generator for C {
             self.src.c,
             "\
                 #include <stdlib.h>
-                #include <{}.h>
+                #include \"{}.h\"
 
                 // The following symbols are never called, but they are sufficient
                 // to get the custom sections in the component type object linked


guybedford:
It would be great to know where this should be tested as well!

pchickey:
C codegen is tested here. I don't know why this issue wasn't caught by this test https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/gen-guest-c/tests/codegen.rs#L39

sunfishcode:
The `cmd.arg("-I").arg(dir);` line adds the directory containing the header file to the `<...>` search path. With the PR here, it looks like we should be able to delete that line.

kulakowski-wasm:
> C codegen is tested here. I don't know why this issue wasn't caught by this test https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/gen-guest-c/tests/codegen.rs#L39

`cmd.arg("-I").arg(dir);` looks to be putting the directory on the search path to make it visible to <> inclusions.

pchickey:
Thanks I forgot how <...> search path works

alexcrichton (approved):

alexcrichton:
Oh oops jumped the gun, but yes we do have tests and I believe a good test for this would be to [remove this argument](https://github.com/bytecodealliance/wit-bindgen/blob/dd51446a5d34f8b95285db76529c463ec9b34597/crates/test-helpers/macros/build.rs#L134-L135)

PR #382, opened 2022-10-18
Implement the component-model lexing rules for identifiers.

sunfishcode:
The [component-model grammer] for kebab-case identifiers now looks like this:

```
name           ::= <word>
                 | <name>-<word>
word           ::= [a-z][0-9a-z]*
                 | [A-Z][0-9A-Z]*
```

Implement the rules. This continues to use XID rules for the initial lexing, as that corresponds to what users might accidentally use, so that we can issue appropriate errors in those cases. The precise grammer is validated in a separate step.

[component-model grammer]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#instance-definitions


alexcrichton:
Given the current CI errors, I think one of two things needs to happen:

* Specify upstream in the component model uppercase isn't allowed for now
* Disallow overlapping identifiers which only overlap in case in `wit-bindgen`

I'd sort of prefer to avoid having to generate separate bindings for `apple` and `APPLE` and support that within wit-bindgen since that'd be a lot of changes for not a ton of benefit.

sunfishcode:
I think we should do something in the upstream component model definition. I filed https://github.com/WebAssembly/component-model/issues/118 to track this.

PR #383, opened 2022-10-19
host-wasmtime-rust: support async stores

pchickey:
This implements support for async wasmtime stores.

It is disabled by default, and configured in the macro with extra arguments `async: true` or at the command line with `--async`.

I updated the wasmtime git dep to after the merge of https://github.com/bytecodealliance/wasmtime/pull/5055.

Tokio is now a dev dependency. Runtime tests (via `#[tokio::test]`) are generated when a `tests/runtime/$suite/host.async.rs` file is found. I only added one for smoke, because I don't see a ton of use for the runtime test beyond smoke testing. Codegen tests  for `async: true` are generated for each suite.

Also includes fix for synchronous wasmtime "unsplatting" in https://github.com/bytecodealliance/wasmtime/pull/5065.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2230,8 +2230,8 @@ name = "wit-bindgen-host-wasmtime-rust"
 version = "0.3.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "bitflags",
- "thiserror",
  "tracing",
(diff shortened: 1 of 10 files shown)


alexcrichton:
Looks great to me, thanks!

PR #385, opened 2022-10-19
Implement case-insensitive kebab-case comparison

sunfishcode:
Implement case-insensitive comparisons for kebab-case names as described in the component-model thread:

https://github.com/WebAssembly/component-model/issues/118

This also adds previously missing validation for duplicate names in some contexts.

### DIFF
--- a/crates/wit-component/src/decoding.rs
+++ b/crates/wit-component/src/decoding.rs
@@ -82,6 +82,9 @@ struct InterfaceDecoder<'a> {
     // key is the result after `TypeId` lookup.
     type_map: IndexMap<PtrHash<'a, types::Type>, Type>,
     name_map: IndexMap<PtrHash<'a, types::Type>, &'a str>,
+
+    function_names: KebabNamespace<()>,
+    type_names: KebabNamespace<()>,
 }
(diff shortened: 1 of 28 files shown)


sunfishcode:
This doesn't yet implement case-insensitive comparisons for wit-component's decoder, though that could be added using the same `KebabNamespace` utility introduced here.

alexcrichton (approved):

alexcrichton:
Would you be ok retargeting this at the `wasm-tools` repository? 

Additionally for the changes to `decoding.rs` in wit-component, would it be ok to leave those out for now? Eventually these changes need to make their way into `wasmparser`'s validation where after that it shouldn't be necessary to have the changes in `decoding.rs`, and for now it seems unlikely a bug will crop up due to that.

peterhuene:
Just an FYI, but I also have a PR nearly ready for `wasm-tools` that brings the tooling up to date with kebab case validation in `wasmparser`

sunfishcode:
@peterhuene Cool, I'll wait for your changes then. Closing this PR as the code is no longer in this repo.

PR #386, opened 2022-10-20
Reimplement the host Python generator with worlds

alexcrichton:
This commit is the equivalent of #381 but for Wasmtime Python host bindings. This is a large-scale refactor of the internals of the Python host generator in which I took quite a few liberties in internal restructuring as well as output restructuring. The tests are probably the best to review and better reflect what changed, but there are some aspects of worlds that the tests are not currently exercising which we'll want to add with the introduction of worlds in the future.

This means that all host generators are now working with worlds as input rather than individual `*.wit` files, and the only two remaining generators are the C and Java generators for guests (which I hope are easier).

The high level summary of the new output is:

    out_dir/
      __init__.py             # top-level component exports/definitions
      types.py                # shared type information, currently just `Result`
      imports/
        __init__.py           # only here if something is imported
        foo.py                # type and protocol definition per interface
      exports/
        __init__.py           # only here if an instance is exported
        bar.py                # one per exported instance

"Default exports" will show up on the generated structure in `out_dir/__init__.py` so all runtime tests, for example, do not generate `exports` at this time. Lots of fiddly stuff went into structuring this all right to get past `mypy` and additionally try to avoid name conflicts. It's still somewhat easy to have name conflicts but ideally they're of the more esoteric category rather than "really easy to run into".


pchickey (approved):
awesome! I won't pretend to understand very much of the code generator, but the way the tests got simplified is really great

PR #387, opened 2022-10-20
Remove the markdown generator

alexcrichton:
This was originally added to showcase the old version of generating markdown from a `*.witx` file, but this functionality has since been moved and used from https://github.com/WebAssembly/wasi-tools. I don't think there's any particular reason to have duplicate functionality here so this commit removes the support from this repository.


sunfishcode (approved):

alexcrichton:
@sunfishcode discussed with me a bit about this and neither of us really remembers why `wasi-tools` is split out from this repository. We concluded that it's probably best for the code to actually live here as opposed to there, so I'm going to close this and I'll work on updating it for worlds and eventually `wasi-tools` can be replaced with the version in this repository.

PR #388, opened 2022-10-21
Host-wasmtime-rust: import functions are able to Trap execution

pchickey:
The new test https://github.com/pchickey/wit-bindgen/blob/wasmtime_trapping/tests/runtime/results/host.rs demonstrates how this behavior works.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,12 +38,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ambient-authority"
-version = "0.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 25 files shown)


alexcrichton (approved):
Looks good to me!

I think CI can be fixed by copying [this block](https://github.com/bytecodealliance/wit-bindgen/blob/09687166aedb82e38c97f49badd4498b27430e2e/crates/test-rust-wasm/Cargo.toml#L51-L53) for the new test.

At least that's how existing tests don't trip up over this error. It may also be fixable by using `#[cfg]` on the `include!` macro invocation so only run the `include!` on wasm which would avoid the need to update `Cargo.toml` as well (which would be nice!)

PR #390, opened 2022-10-21
gen-guest-c: align wit-component naming, ws tweaks

guybedford:
This fixes up the naming issue between the C guest generator and wit-component, where it seems like wit-component is expecting a name of the form `interfacename#fnname` while the C generator is currently only outputting `fnname`.

Some minor whitespace tweaks are also included to avoid some more unnecessary newlines.

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -1106,7 +1106,8 @@ impl Generator for C {
         // canonical ABI.
         uwriteln!(
             self.src.c_adapters,
-            "\n__attribute__((export_name(\"{}\")))",
+            "\n__attribute__((export_name(\"{}#{}\")))",
+            iface.name.to_kebab_case(),
             func.name
         );
         let import_name = self.names.tmp(&format!(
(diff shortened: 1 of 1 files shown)


guybedford:
Tests are currently failing, but this change is necessary to get an end-to-end example running. Not sure which side this is supposed to be fixed, so guidance would be welcome.

crates/gen-guest-c/src/lib.rs
```diff
@@ -1151,7 +1152,8 @@ impl Generator for C {
         if iface.guest_export_needs_post_return(func) {
             uwriteln!(
                 self.src.c_fns,
-                "\n__attribute__((weak, export_name(\"cabi_post_{}\")))",
+                "\n__attribute__((weak, export_name(\"cabi_post_{}#{}\")))",
+                iface.name.to_kebab_case(),
                 func.name
```
guybedford, line 1157: Combining kebab case and snake case?

guybedford:
Fix is to use the `wit-component --interface` option not `wit-component --export` in component generation.

peterhuene:
This will be fixed properly once the C generator moves to world-based generation (I also have a change ready to make it consistent between the guest generators and what `wit-component` expects).

Issue #53, opened 2021-07-26
JS host generator uses native endianness when copying "canonical" lists 


alexcrichton:
Currently it just uses slices of the underlying array buffer and doesn't attempt to swap endianness, which means that little-endian integers from wasm will be exposed as native-endian integers in JS, which means that on big-endian this will all be backwards.

PR #392, opened 2022-10-22
gen-host-js: utf16 support

guybedford:
Adds UTF16 support to the js-host generator.

In the process, this also refactors to use all camelcase for variables - removing the uppercase module scope variables, and underscore naming, as usually only window-level globals sometimes have capitalization in JS, as well as implementing a shorter data_view function. Feel free to tell me not to meddle though!


guybedford:
I've updated this PR to go all-in on camel casing, I'm happy to remove those parts though if it seems weird.

Pointers re a good way to get the tests integrated would be very useful, as I'm not quite sure how utf16 should fit into the current test pipeline.

crates/gen-host-js/src/lib.rs
```diff
@@ -418,69 +417,91 @@ impl Js {
...
-                    UTF8_ENCODED_LEN = writtenTotal;
+                    if (allocLen > writtenTotal)
+                        ptr = realloc(ptr, allocLen, 1, writtenTotal);
+                    utf8EncodedLen = writtenTotal;
+                    return ptr;
+                }
+            "),
+
+            Intrinsic::Utf16Encode => self.src.js("
+                const isLE = new Uint8Array(new Uint16Array([1]).buffer)[0] === 1;
```
alexcrichton, line 486: Handling endianness here is interesting since in general the JS generator doesn't do great with endianness today. Handing out typed array views of wasm content currently always hands out a little-endian view of memory and doesn't try to account for endianness. Single-value loads and stores, however, handle endianness at this time although that was because `DataView` provided the flag and the only reason `DataView` was used historically was for unaligned accesses which are now traps in the component model so we should probably replace `DataView` with typed arrays as well.

In any case as-written this PR I think handles endianness during encoding here but not during decoding. I'm not sure how well supported the `utf-16le` encoding is but that could possibly be an easy fix for the decoding portion.

Do you have a need to support little-endian at this time? If not I'd be tempted to basically `assert(isLE)` at the top of the generated script and defer to https://github.com/bytecodealliance/wit-bindgen/issues/53 for fixing the underlying issue

crates/gen-host-js/src/lib.rs
```diff
@@ -418,69 +417,91 @@ impl Js {
...
+
+                function utf16Encode (str, realloc, memory) {
+                    const len = str.length, ptr = realloc(0, 0, 1, len), out = new Uint16Array(memory.buffer, ptr, len);
```
alexcrichton, line 489: I think the 1 here for alignment needs to be 2

crates/gen-host-js/src/lib.rs
```diff
@@ -713,12 +734,11 @@ impl Instantiator<'_> {
...
     ) {
-        // Technically it wouldn't be the hardest thing in the world to support
-        // other string encodings, but for now the code generator was originally
-        // written to support utf-8 so let's just leave it at that for now. In
-        // the future when it's easier to produce components with non-utf-8 this
-        // can be plumbed through to string lifting/lowering below.
-        assert_eq!(opts.string_encoding, StringEncoding::Utf8);
+        // Only Utf8 and Utf16 supported for now
+        assert!(matches!(
+            opts.string_encoding,
```
alexcrichton, line 739: Could this option get plumbed through into the `FunctionBindgen` which would push this assert into the `String{Lift,Lower}` portions?

alexcrichton (commented):
Looks reasonable to me! Happy to take all the naming bits, but I think the bigger piece to handle will be testing this. I do think we'll want to land this with end-to-end tests from a guest into the JS host to ensure everything works. To do that I think the way would be:

* Pick a guest generator and add a `--string-encoding utf16` option.
* Implement the option in the generator, adjusting lift/lower as necessary and the types printed for strings
* Get the strings-are-utf16 option plumbed into the final component with either:
  * Use a file name like `wasm.utf16.rs` for the test case and use that to decide whether to pass the utf-16 option to `wit-component`
  * Invent the ability to ferry data through a custom section to get fed into `wit-component` to describe the string encoding. For example right now the custom section is `component-type:name` and has a binary-encoded component type, and there could possibly be something like `component-string-encoding:name` with a one-byte payload that is the string encoding which matches the encodings of string encodings in the component model spec.


With all that it should be possible to have one language as a guest and at least JS as a host. Ideally this would support Wasmtime as well since it should already have utf16 support but I can work on adding a test for that later.

crates/gen-host-js/src/lib.rs
alexcrichton: Oh I definitely think that endianness needs to be handled correctly, but I mostly wanted to point out that the handling here is only one piece of the picture and there's more places. The decoding was one but I wanted to point out the preexisting locations where it's not handled as well. I believe the preexisting locations are in `ListCanon{Lower,Lift}` where we're reading a typed array from wasm (`Lift`) and copying a host-defined typed array into wasm (`Lower`)

guybedford: Understood, can we keep the scope of this PR to UTF16 handling, while accepting partial endian support? I've separated out the endian detection intrinsic.

guybedford:
I've integrated the tests as discussed here for a C -> JS workflow with some string encoding checks etc.

With regards to the C helper functions, I had to implement a UTF16 strlen function - it might be better to entirely avoid these helpers using the strlen function and rather explicitly always take a length requiring the user to pass the length to the helpers. Shall we add that as well while we're on this?

Would be great to move on to error handling and output formatting work if we can get this landed soon.

crates/test-helpers/macros/build.rs
```diff
@@ -183,6 +184,106 @@ fn main() {
...
+            fs::write(&component_path, component).expect("write component to disk");
+
+            wasms.push((
+                "c_utf16",
+                stem,
+                out_wasm.to_str().unwrap().to_string(),
+                component_path.to_str().unwrap().to_string(),
+            ));
+        }
+    }
```
alexcrichton, line 285: I think this `if` is basically copy/pasted from above, so could this be refactored instead to a separate function to reduce the duplication here?

crates/test-helpers/macros/build.rs
```diff
@@ -183,6 +184,106 @@ fn main() {
         }
     }
 
+    if cfg!(feature = "guest-c-utf16") {
```
alexcrichton, line 187: I think it's ok to skip an extra feature here, it's ok to gate the utf16 tests for C behind `guest-c`

crates/wit-component/src/encoding.rs
```diff
@@ -2093,7 +2093,8 @@ impl<'a> ImportEncoder<'a> {
 #[derive(Default)]
 pub struct ComponentEncoder {
     module: Vec<u8>,
-    encoding: StringEncoding,
+    /// String encoding
+    pub encoding: StringEncoding,
```
alexcrichton, line 2097: Instead of making this field public I think you should be able to use the `ComponentEncoder::encoding` method.

tests/runtime/strings/wasm_utf16.c
```diff
@@ -0,0 +1,28 @@
+#include <assert.h>
+#include <imports.h>
+#include <exports.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+char16_t BASIC_STRING[] = { 'l', 'a', 't', 'i', 'n', ' ', 'u', 't', 'f', '1', '6', '\0' };
```
alexcrichton, line 8: Could this use the `u"foo"` syntax for the string literal here?

tests/runtime/strings/wasm_utf16.c
```diff
@@ -0,0 +1,28 @@
...
+}
+
+void exports_f1(exports_string_t *str) {
+  memcpy(&STR_BUFFER[0], str->ptr, str->len * 2);
+  STR_BUFFER[str->len] = '\0';
+}
+
+void exports_f2(exports_string_t *ret) {
+  exports_string_set(ret, STR_BUFFER);
+}
```
alexcrichton, line 28: Could you add a test here that asserts with a `memcmp` that the expected value of the string is correct in C? (may need to be a new function)

crates/gen-host-js/src/lib.rs
```diff
@@ -418,69 +423,89 @@ impl Js {
...
 
             Intrinsic::Utf8Decoder => self
                 .src
-                .js("const UTF8_DECODER = new TextDecoder('utf-8');\n"),
+                .js("const utf8Decoder = new TextDecoder('utf-8');\n"),
 
-            Intrinsic::Utf8EncodedLen => self.src.js("let UTF8_ENCODED_LEN = 0;\n"),
+            Intrinsic::Utf16Decoder => self
+                .src
+                .js("const utf16Decoder = new TextDecoder(`utf-16${isLE ? '' : 'be'}`);\n"),
```
alexcrichton, line 455: This was somewhat surprising to me in that I would expect this to always use `utf-16le` as the encoding since wasm is always little-endian. I get mixed up with endianness though so I'm not sure if the `be` is to swap things back to normal or not.

guybedford: `utf-16` means `utf-16le` per the spec, so that's why I left out the explicit `le`. I think `new TextDecoder()` without any arg is supposed to work with a Uint16Array as UTF16 and it seems to, I was just unsure if that mode would try to revert to UTF8 somehow so was being explicit. Happy to adjust further.

alexcrichton: > I think new TextDecoder() without any arg is supposed to work with a Uint16Array as UTF16 and it seems to,

Locally on Firefox I get:

```
(new TextDecoder()).decode(new Uint16Array([0, 0])) == "\u0000\u0000\u0000\u0000" 
(new TextDecoder('utf-16')).decode(new Uint16Array([0, 0])) == "\u0000\u0000" 
```

so I think that an argument is required here?

> utf-16 means utf-16le per the spec

Ah I was not aware of that, I mistakenly presumed `utf-16` was utf-16 with native-endianness. In that case `utf-16` should be sufficient to always pass here.

I don't believe it's correct to use `utf-16be` here. When I run the tests in `qemu-s390x` the tests fail, and when I remove the `be` here the tests pass.

crates/gen-host-js/src/lib.rs
```diff
@@ -418,69 +423,89 @@ impl Js {
...
+                    utf8EncodedLen = writtenTotal;
+                    return ptr;
+                }
+            "),
+
+            Intrinsic::Utf16Encode => self.src.js("
+                function utf16Encode (str, realloc, memory) {
+                    const len = str.length, ptr = realloc(0, 0, 2, len), out = new Uint16Array(memory.buffer, ptr, len);
+                    let i = 0;
+                    if (isLE) {
```
alexcrichton, line 495: Out of curiosity, is this measurably faster than `TextEncoder` which would handle endianness automatically?

tests/runtime/strings/wasm_utf16.c
```diff
@@ -0,0 +1,31 @@
...
+void exports_f1(exports_string_t *str) {
+  assert(str->len > 0);
+  memcpy(&STR_BUFFER[0], str->ptr, str->len * 2);
+  STR_BUFFER[str->len] = '\0';
+}
+
+void exports_f2(exports_string_t *ret) {
+  exports_string_set(ret, STR_BUFFER);
+  assert(ret->len > 0);
+  assert(memcmp(ret->ptr, STR_BUFFER, ret->len * 2) == 0);
```
alexcrichton, line 30: Sorry I meant moreso there are no assertions that a particular value is actually received in C. The assertions added here aren't too meaningful of the actual bytes received, e.g. that the endianness was all correct. Could a function be added to the C exports where it receives a known string and then asserts that it receives that string?

guybedford: Ah I see what you mean, sure.

crates/gen-host-js/src/lib.rs
guybedford: I believe `TextEncoder` only supports UTF8.

guybedford: See https://encoding.spec.whatwg.org/#interface-textencoder

alexcrichton: TIL, makes sense!

crates/gen-host-js/src/lib.rs
guybedford: > I don't believe it's correct to use utf-16be here. When I run the tests in qemu-s390x the tests fail, and when I remove the be here the tests pass.

I'm pretty sure according to the specs then that what we have currently for big endian handling is correct but that there may be another big endian related failure happening in the code perhaps due to the other intrinsics?

alexcrichton: Perhaps yeah? We don't test big-endian in CI but if you're looking to handle endianness here I think it would be good to test at least locally that the endianness is right (for me at least this PR as-is does not pass tests)

guybedford: I really think it's very important we sort out endianness here, so removing logic that works towards that goal seems counter productive to me.

guybedford: The techniques I'm using have been tested before on big endian hardware, even if unrelated bindgen support is likely causing further failures.

tests/runtime/strings/wasm_utf16.c
guybedford: I've added some direct assertions in the import pattern. The other pattern is a write / read pattern from the JS side so there isn't a reliable assertion to run unless we track the call count.

tests/runtime/strings/wasm_utf16.c
```diff
@@ -0,0 +1,40 @@
...
+#include <exports.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+char16_t BASIC_STRING[] = u"latin utf16";
+// 🚀 = 0xD83D 0xDE80
+// 𠈄 = 0xD840 0xDE04
+// 𓀀 = 0xD80C 0xDC00
+char16_t UNICODE_STRING[] = { 0xD83D, 0xDE80, 0xD83D, 0xDE80, 0xD83D, 0xDE80, ' ', 0xD840, 0xDE04, 0xD80C, 0xDC00 };
```
alexcrichton, line 12: Could this be `u"🚀🚀🚀 𠈄𓀀"`?

tests/runtime/strings/wasm_utf16.c
```diff
@@ -0,0 +1,40 @@
...
+char16_t STR_BUFFER[500];
+
+void assert_str(imports_string_t* str, char16_t* expected, size_t expected_len) {
+  assert(str->len == expected_len);
+  assert(memcmp(str->ptr, expected, 11) == 0);
```
alexcrichton, line 17: I think the 11 here may be a typo

alexcrichton: I also think that the length needs to be multiplied by 2 to compare the full byte size of the string?

tests/runtime/strings/wasm_utf16.c
guybedford: I did try that but it didn't seem to work for some reason.

tests/runtime/strings/wasm_utf16.c
guybedford: Thanks, yes this was left in from testing.

crates/gen-host-js/src/lib.rs
guybedford: Per further discussion, the decoder is now using UTF16 LE due to the underlying buffer being a Wasm little endian buffer.

PR #396, opened 2022-10-25
Update the guest C generator for world-based generation

alexcrichton:
This commit removes the `Generator` trait impl for the C generator and adds the `WorldGenerator` trait impl to replace it. This is the equivalent of #386 for the guest C generator.

The generated C code is largely the same, except that the entire world is represented in one header file instead of per-interface header files as before. Additionally a tiny amount of "type sharing" is now done where all interfaces share the same string type (namespaced by the world name). More type sharing should come with a more first-class implementation of worlds.

There was quite a lot of code movement within the generator as I got it working again, but at the surface level very little has changed and it should largely be the same as before.

### DIFF
--- a/crates/bindgen-core/src/lib.rs
+++ b/crates/bindgen-core/src/lib.rs
@@ -564,6 +564,7 @@ mod tests {
 
 pub trait WorldGenerator {
     fn generate(&mut self, name: &str, interfaces: &ComponentInterfaces, files: &mut Files) {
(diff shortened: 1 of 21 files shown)


alexcrichton:
This'll certainly conflict with https://github.com/bytecodealliance/wit-bindgen/pull/392 so I think it's best to hold off on merging this until that's in.

pchickey (approved):

PR #398, opened 2022-10-26
Allow adapters to import multiple interfaces

alexcrichton:
This commit lifts the restriction that an adapter module as part of `wit-component` can only import a single interface. That restriction was a temporary artifact of the initial implementation due to the fact that at the time there was no convenient way to communicate about multiple imported interfaces. With the `ComponentInterfaces` encoding of a "world" now, though, the `wit-component` CLI is updated to require that the incoming adapter module has a `component-type*` custom section describing the world that it's importing.


pchickey (approved):
Awesome!

PR #399, opened 2022-10-26
cli: default naming, default string encoding flags

guybedford:
Recent PRs made the `--string-encoding` and `--name` flags mandatory. This makes those optional again with defaults. `--name` defaults to the default interface name, first export name or import name. This also adds a new CLI error if no interface option is provided as well.

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -31,7 +31,7 @@ pub struct Opts {
     #[cfg_attr(feature = "clap", arg(long))]
     pub no_helpers: bool,
     /// Set component string encoding
-    #[cfg_attr(feature = "clap", arg(long, default_value_t = StringEncoding::UTF8))]
+    #[cfg_attr(feature = "clap", arg(long, default_value_t = StringEncoding::default()))]
     pub string_encoding: StringEncoding,
 }
 
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -31,7 +31,7 @@ pub struct Opts {
     #[cfg_attr(feature = "clap", arg(long))]
     pub no_helpers: bool,
     /// Set component string encoding
-    #[cfg_attr(feature = "clap", arg(long))]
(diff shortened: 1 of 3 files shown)


src/bin/wit-bindgen.rs
```diff
@@ -283,7 +283,21 @@ fn gen_world(
...
+        None => {
+            if let Some(default) = &interfaces.default {
+                &default.name
+            } else if interfaces.exports.len() > 0 {
+                &interfaces.exports.first().unwrap().1.name
+            } else if interfaces.imports.len() > 0 {
+                &interfaces.imports.first().unwrap().1.name
+            } else {
+                return Err(anyhow!("no interfaces provided to generate"));
+            }
```
alexcrichton, line 297: Eventually my intention was to make the `--name` flag optional and inferred from the "world" where eventually a world will be specified all in one file or the default world would have a name. For example I'd expect the invocation to look like `wit-bindgen guest c foo.wit` would infer the name "foo" for the generated bindings and the `--import`/`--export`/`--default` flags would all go away.

That's all for the future, though, and it seems reasonable to remove the required `--name` for now. That being said I'm a little uncomfortable just picking the first export or import, so could this require either `--name` or `--default`, otherwise returning an error saying that one of those must be specified?

crates/gen-guest-c/src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub struct Opts {
     #[cfg_attr(feature = "clap", arg(long))]
     pub no_helpers: bool,
     /// Set component string encoding
-    #[cfg_attr(feature = "clap", arg(long))]
+    #[cfg_attr(feature = "clap", arg(long, default_value_t = StringEncoding::UTF8))]
```
alexcrichton, line 34: Could this be `StringEncoding::default()`?

src/bin/wit-bindgen.rs
guybedford: Yeah, understood this will all change, but nice to keep the overall CLI expectations. Sure, added!

PR #400, opened 2022-10-26
JS host formatting

guybedford:
This is entirely a formatting PR, with some minor changes to the output:

* Fixes inconsistent newline whitespace appearing in the bindings
* Emits unquoted object properties when they are valid JS identifiers
* Refactors single line intrinsic functions into arrow functions

The perspective I'm taking with regards to output here is that output consistency and output size is really important with the JS stuff and that we should start to build up some more complex optimization code that can ensure the best final minified library size. I think it makes sense to continue in incremental directions like this even if it introduces quite a bit of additional code complexity - that is small improvements in output size should be worth the additional complexity, but I'm happy to discuss this topic further.

### DIFF
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -2202,7 +2202,7 @@ impl Bindgen for FunctionBindgen<'_> {
                 results.push(result.clone());
 
                 uwriteln!(self.src.js, "for (let i = 0; i < {len}; i++) {{");
-                uwrite!(self.src.js, "const base = {base} + i * {size};");
+                uwriteln!(self.src.js, "const base = {base} + i * {size};");
                 self.src.js(&body);
                 assert_eq!(body_results.len(), 1);
                 uwriteln!(self.src.js, "{result}.push({});", body_results[0]);
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -17,6 +17,41 @@ use wit_bindgen_core::{
 };
 use wit_component::ComponentInterfaces;
 
+fn is_js_identifier_start(code: char) -> bool {
+    return match code as u32 {
+        0..=35 | 37..=64 | 91..=94 | 96..=96 => false,
+        36 | 65..=90 | 95 | 97..=122 => true,
+        // leaving out non-ascii for now...
+        _ => false,
+    };
+}
+
+fn is_js_identifier_char(code: char) -> bool {
+    return match code as u32 {
+        0..=35 | 37..=47 | 58..=64 | 91..=94 | 96..=96 => false,
+        36 | 48..=57 | 65..=90 | 95 | 97..=122 => true,
+        // leaving out non-ascii for now...
+        _ => false,
+    };
+}
+
+fn is_js_identifier(s: &str) -> bool {
+    let mut chars = s.chars();
+    if let Some(char) = chars.next() {
+        if !is_js_identifier_start(char) {
+            return false;
+        }
+    } else {
+        return false;
+    }
+    while let Some(char) = chars.next() {
+        if !is_js_identifier_char(char) {
+            return false;
+        }
+    }
+    return true;
+}
+
 #[derive(Default)]
 struct Js {
     /// The source code for the "main" file that's going to be created for the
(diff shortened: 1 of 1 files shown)


pchickey:
Asking mostly for my own edification on JS tools ecosystem and community expectations - are there automatic formatting tools (like `rustfmt`) that we can use to automate some of this? What about linters or other analysis tools that would spot these sorts of problems?

guybedford:
Yes, we could certainly integrate some eslint rule checking into the JS generated output to converge on some style rules. We likely would also want to have some direct output snapshots, and to make reviewing these PRs easier as well.

For now, my focus is on the size of the output when it is passed into terser minification first and foremost, since JS file size is an important factor.

crates/gen-host-js/src/lib.rs
```diff
@@ -191,7 +226,12 @@ impl ComponentGenerator for Js {
     }
 
     fn finish_component(&mut self, name: &str, files: &mut Files) {
-        files.push(&format!("{name}.js"), self.src.js.as_bytes());
+        let mut bytes = self.src.js.as_bytes();
+        // strip leading newline
+        if bytes[0] == '\n' as u8 {
```
alexcrichton, line 231: As a shorthand you can use `b'\n'` instead of `'\n' as u8` here

crates/gen-host-js/src/lib.rs
```diff
@@ -17,6 +17,41 @@ use wit_bindgen_core::{
 };
 use wit_component::ComponentInterfaces;
 
+fn is_js_identifier_start(code: char) -> bool {
+    return match code as u32 {
+        0..=35 | 37..=64 | 91..=94 | 96..=96 => false,
+        36 | 65..=90 | 95 | 97..=122 => true,
```
alexcrichton, line 23: Could this use `'a'..='z'` style patterns instead of integer codes to be more readable? Additionally is there something online to link to as a reference here?

alexcrichton (commented):
This seems reasonable to me, but at a high level is there an actual need to focus on generated code size here? I have previously assumed that there's no need to worry about code size here so long as a minifier can minify the code to what you're otherwise changing this to produce. If minifiers can't productively minify the generated code then it makes sense to change it, but otherwise I would personally prefer to generate readable over minified code.

guybedford:
> This seems reasonable to me, but at a high level is there an actual need to focus on generated code size here? I have previously assumed that there's no need to worry about code size here so long as a minifier can minify the code to what you're otherwise changing this to produce. If minifiers can't productively minify the generated code then it makes sense to change it, but otherwise I would personally prefer to generate readable over minified code.

Yes agreed we don't have to output minified code. One simple metric would literally be to track the minified output sizes when run through terser.

The whitespace and identifier unquoting changes are purely aesthetic - but it is still nice to have output that looks good within reason (and these are small diffs). Agreed complexity at this level should be avoided.

The only change in the above that affects the minification metric slightly here is the function to arrow function rewriting, and this is not something minifiers will do automatically.

alexcrichton:
I understand whitespace is purely aesthetic and good to have improvements at any time, I'm specifically referring to your comment above

> since JS file size is an important factor

which I was presuming was the reason for changes such as:

```diff
-                function i32ToF32(i) {
-                    i32ToF32I[0] = i;
-                    return i32ToF32F[0];
-                }
+                const i32ToF32 = i => (i32ToF32I[0] = i, i32ToF32F[0]);
```

which I personally find less readable since I basically learned about `(expr, expr)` syntax from reading that.

I was also not aware that minifiers don't automatically convert between `function` and `=>` syntax, however.

Basically I wanted to emphasize that I would prefer, unless necessary, to prevent reducing the readability of code.

guybedford:
Makes sense. Yeah I'm happy to revert that change if you'd prefer the readability. It's definitely a micro-optimization, but from a high level think we should have the position that _any_ reductions in the minified (+ gzip / brotli of course) metric are worth considering.

alexcrichton:
Personally I find refactorings fine, I mostly want to make sure that the crate is aligned and consistent with how it treats generated code. In other words I'd prefer that the generated code is not optimized for size, but rather it's instead generated in such a way such that if fed to a minifier it can be optimized for size. If that requires changing `function` to `=>` then that seems fine, but the change to `i32ToF32` and other functions seems like more of a stylistic choice rather, which is also ok but I wanted to make sure to classify it correctly.

To fully take size seriously, though, I think it would be good to add some tests to this repository in one form or another. For example this could take a `*.wit` and assert that the compressed output stays below a particular threshold or something like that. Or otherwise a test could have a "golden output" to review the smaller generation diff through a PR. Basically I think it would be preferable to have something machine-checked here rather than golfing through PRs without backing assertions to guarantee it is indeed better.

For the time being though all the changes here are fine, so I'm going to merge.

PR #402, opened 2022-10-26
gen-host-js: support undefined values in lowering for nullable options

guybedford:
This makes it so that a function with optional nullable parameters can accept undefined values as a substitute.

Eg `fn(a, null)` can now be `fn(a)` for a nice optional param pattern.

### DIFF
--- a/tests/runtime/variants/host.ts
+++ b/tests/runtime/variants/host.ts
@@ -36,6 +36,8 @@ async function run() {
   wasm.testImports();
   assert.deepStrictEqual(wasm.roundtripOption(1), 1);
   assert.deepStrictEqual(wasm.roundtripOption(null), null);
+  assert.deepStrictEqual(wasm.roundtripOption(undefined), null);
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

alexcrichton:
Oh also mind adding a test for this?

tests/runtime/variants/host.ts
```diff
@@ -36,6 +36,10 @@ async function run() {
   wasm.testImports();
   assert.deepStrictEqual(wasm.roundtripOption(1), 1);
   assert.deepStrictEqual(wasm.roundtripOption(null), null);
+  // @ts-ignore
```
guybedford, line 39: For now I've just ignored this behaviour from a typescript perspective - indicating that function parameters are optional to TypeScript would be a nice thing to add but requires `isnullable` information at the param writer level.

alexcrichton: https://github.com/bytecodealliance/wit-bindgen/issues/404

PR #405, opened 2022-10-27
update the Java guest generator for world-based generation

dicej:
We now generate a top-level class for the world, which contains any tuple, result, and cleanup types, as well as the return area (if needed) and the custom section directive for the component type.  Each imported or exported interface also gets its own top level file in the same package as the world class.


alexcrichton (approved):
Looks great to me, thanks again for helping out with this!

PR #406, opened 2022-10-28
gen-host-js: instance ESM output by default

guybedford:
Outputs a direct ES module from the JS host generator by default that can be directly imported in JS environments.

For the custom instantiation API that was the previous output, a new `--instantiation` flag is provided.

Instance-level customization flags are also added including:
* `--compat`: Enables all compatibility mode flags (right now just `--nodejs-compat`, but based on the assumption that a new --tla-compat flag will be added in a follow-up PR).
* `--nodejs-compat`: Handle Node.js instantiation when no fetch global is present

The above flags are only compatible with instance output and will throw otherwise.

This PR also parallelizes the core Wasm instantiations, and converts the instantiation API from `instantiateCore(path, imports)` to `compileCore(path)` to enable this parallelization.

In addition a new `--map` is added with support for custom import specifier mappings that can apply to both output modes. For example - `--map "imports=./imports.js"` allows specifying that the imports should resolve to a relative JS specifier.

For testing, the first line comment of the JS file starting with `// Flags:` is used to pass these output options flags, with most tests now including `// Flags: --instantiation` for backwards-compat.

### DIFF
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -717,10 +717,16 @@ impl Instantiator<'_> {
                 let local_name = format!("module{}", idx.as_u32());
                 let name = format!("module{}.wasm", idx.as_u32());
                 if self.gen.opts.instantiation {
-                    uwrite!(self.src.js, "const {local_name} = compileCore(\"{name}\");\n");
+                    uwrite!(
+                        self.src.js,
+                        "const {local_name} = compileCore(\"{name}\");\n"
+                    );
                 } else {
                     let load_wasm = self.gen.intrinsic(Intrinsic::LoadWasm);
-                    uwrite!(self.src.js, "const {local_name} = {load_wasm}(new URL('./{name}', import.meta.url));\n");
+                    uwrite!(
+                        self.src.js,
+                        "const {local_name} = {load_wasm}(new URL('./{name}', import.meta.url));\n"
+                    );
                 }
             }
         }
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -711,7 +711,9 @@ impl Instantiator<'_> {
             if let GlobalInitializer::InstantiateModule(InstantiateModule::Static(idx, _)) = init {
                 // Get the compiled WebAssembly.Module objects in parallel
                 if first {
-                    self.src.js.push_str("\n");
+                    if !instantiation {
+                        self.src.js.push_str("\n");
+                    }
                     first = false;
                 }
                 let local_name = format!("module{}", idx.as_u32());
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -705,11 +705,33 @@ impl Instantiator<'_> {
             );
         }
 
-        // First do the core Wasm compilations in parallel
-        self.src.js.push_str("const [");
+        // Setup the compilation promises
         let mut first = true;
         for init in self.component.initializers.iter() {
             if let GlobalInitializer::InstantiateModule(InstantiateModule::Static(idx, _)) = init {
+                // Get the compiled WebAssembly.Module objects in parallel
+                if first {
+                    self.src.js.push_str("\n");
+                    first = false;
+                }
+                let local_name = format!("module{}", idx.as_u32());
+                let name = format!("module{}.wasm", idx.as_u32());
+                if self.gen.opts.instantiation {
+                    uwrite!(self.src.js, "const {local_name} = compileCore(\"{name}\");\n");
+                } else {
+                    let load_wasm = self.gen.intrinsic(Intrinsic::LoadWasm);
+                    uwrite!(self.src.js, "const {local_name} = {load_wasm}(new URL('./{name}', import.meta.url));\n");
+                }
+            }
+        }
+
+        // To avoid uncaught promise rejection errors, we attach an intermediate
+        // Promise.all with a rejection handler.
+        first = true;
+        self.src.js.push_str("Promise.all([");
+        for init in self.component.initializers.iter() {
+            if let GlobalInitializer::InstantiateModule(InstantiateModule::Static(idx, _)) = init {
+                // Get the compiled WebAssembly.Module objects in parallel
                 if first {
                     first = false;
                 } else {
@@ -718,29 +740,7 @@ impl Instantiator<'_> {
                 self.src.js.push_str(&format!("module{}", idx.as_u32()));
             }
         }
-        self.src.js.push_str("] = await Promise.all([");
-        first = true;
-        for init in self.component.initializers.iter() {
-            if let GlobalInitializer::InstantiateModule(InstantiateModule::Static(idx, _)) = init {
-                if first {
-                    first = false;
-                } else {
-                    self.src.js.push_str(", ");
-                }
-                // Get the compiled WebAssembly.Module objects in parallel
-                let name = format!("module{}.wasm", idx.as_u32());
-                if self.gen.opts.instantiation {
-                    uwrite!(self.src.js, "compileCore(\"{name}\")",);
-                } else {
-                    let load_wasm = self.gen.intrinsic(Intrinsic::LoadWasm);
-                    uwrite!(
-                        self.src.js,
-                        "{load_wasm}(new URL('./{name}', import.meta.url))",
-                    );
-                }
-            }
-        }
-        self.src.js.push_str("]);\n");
+        self.src.js.push_str("]).catch(() => {});\n");
 
         for init in self.component.initializers.iter() {
             self.instantiation_global_initializer(init);
@@ -848,7 +848,7 @@ impl Instantiator<'_> {
         uwriteln!(
             self.src.js,
             "
-                const instance{iu32} = await WebAssembly.instantiate(module{}, {imports});\
+                const instance{iu32} = await WebAssembly.instantiate(await module{}, {imports});\
             ",
             idx.as_u32()
         );
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -734,21 +734,25 @@ impl Instantiator<'_> {
         }
 
         // To avoid uncaught promise rejection errors, we attach an intermediate
-        // Promise.all with a rejection handler.
-        first = true;
-        self.src.js.push_str("Promise.all([");
-        for init in self.component.initializers.iter() {
-            if let GlobalInitializer::InstantiateModule(InstantiateModule::Static(idx, _)) = init {
-                // Get the compiled WebAssembly.Module objects in parallel
-                if first {
-                    first = false;
-                } else {
-                    self.src.js.push_str(", ");
+        // Promise.all with a rejection handler, if there are multiple promises.
+        if first == false {
+            first = true;
+            self.src.js.push_str("Promise.all([");
+            for init in self.component.initializers.iter() {
+                if let GlobalInitializer::InstantiateModule(InstantiateModule::Static(idx, _)) =
+                    init
+                {
+                    // Get the compiled WebAssembly.Module objects in parallel
+                    if first {
+                        first = false;
+                    } else {
+                        self.src.js.push_str(", ");
+                    }
+                    self.src.js.push_str(&format!("module{}", idx.as_u32()));
                 }
-                self.src.js.push_str(&format!("module{}", idx.as_u32()));
             }
+            self.src.js.push_str("]).catch(() => {});\n");
         }
-        self.src.js.push_str("]).catch(() => {});\n");
 
         for init in self.component.initializers.iter() {
             self.instantiation_global_initializer(init);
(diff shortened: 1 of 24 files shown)


crates/gen-host-js/tests/runtime.rs
```diff
@@ -75,3 +91,12 @@ fn execute(name: &str, lang: &str, wasm: &Path, ts: &Path) {
...
     );
 }
+
+fn get_first_line_flag_comment(src: &str) -> &str {
+    if src.starts_with("// Flags:") {
+        if let Some(nl_idx) = src.find(|c| c == '\r' || c == '\n') {
+            return &src[9..nl_idx];
+        }
+    }
+    return "";
```
alexcrichton, line 101: This can perhaps be simplified with:

```rust
src.lines().next().and_then(|s| s.strip_prefix("// Flags: ")).unwrap_or("")
```

tests/runtime/flavorful/host.ts
```diff
@@ -75,4 +71,5 @@ async function run() {
   assert.deepStrictEqual(r2, ['typedef4']);
 }
 
-await run()
+// TLA cycle avoidance
```
alexcrichton, line 74: What is "TLA"? (curious, haven't seen this before)

tests/runtime/flavorful/host.ts
```diff
@@ -1,53 +1,49 @@
...
-          [false, true],
-          [{ tag: 'err', val: undefined }, { tag: 'ok', val: undefined }],
-          ["a", "b"],
-        ];
-      },
-    },
-  });
+export async function run () {
+  // @ts-ignore
+  const wasm = await import('./flavorful.js');
```
alexcrichton, line 46: Is this possible to lift to a top-level `import`?

Additionally, how come this needs `@ts-ignore`?

crates/gen-host-js/src/lib.rs
```diff
@@ -250,15 +256,25 @@ impl WorldGenerator for Js {
     fn export(&mut self, name: &str, iface: &Interface, files: &mut Files) {
         self.generate_interface(name, iface, "exports", "Exports", files);
         let camel = name.to_upper_camel_case();
-        uwriteln!(self.export_object, "{name}: {camel}Exports;");
+        if self.opts.instantiation {
+            uwriteln!(self.export_object, "{name}: {camel}Exports;");
+        } else {
+            uwriteln!(self.src.ts, "TODO");
```
alexcrichton, line 262: Note to handle this before merging.

crates/gen-host-js/src/lib.rs
```diff
@@ -313,10 +333,20 @@ impl Js {
             files.push(&format!("{dir}/{name}.d.ts"), gen.src.ts.as_bytes());
         }
 
-        uwriteln!(
-            self.src.ts,
-            "import {{ {camel} as {camel}{extra} }} from \"./{dir}/{name}\";"
-        );
+        // TODO: instance mode imports typings
```
alexcrichton, line 336: Note to handle this before merging

crates/gen-host-js/src/lib.rs
```diff
@@ -74,20 +41,55 @@ struct Js {
...
-    pub fn build(self) -> Box<dyn ComponentGenerator> {
+    pub fn build(self) -> Result<Box<dyn ComponentGenerator>> {
         let mut gen = Js::default();
         gen.opts = self;
-        Box::new(gen)
+        if gen.opts.compat {
+            gen.opts.nodejs_compat = true;
+        }
+        if let Some(map) = gen.opts.map.as_ref() {
+            gen.import_maps = maps_str_to_map(&map)?;
```
alexcrichton, line 90: I think that you can use `#[clap(value_parser = maps_str_to_map)]` on the field in `Opts` to avoid needing to manually do the parse step here if you'd like.

crates/gen-host-js/src/lib.rs
```diff
@@ -729,8 +784,22 @@ impl Instantiator<'_> {
...
+                "const instance{iu32} = await instantiateCore(\"{name}\", {imports});"
+            );
+        } else {
+            let load_wasm = self.gen.intrinsic(Intrinsic::LoadWasm);
+            uwriteln!(
+                self.src.js,
+                "
+                const instance{iu32} = await {load_wasm}(new URL('./{name}', import.meta.url))
+                .then(m => WebAssembly.instantiate(m, {imports}));\
+            "
```
alexcrichton, line 800: I think this string literal may need one tab of indentation to match.

Additionally, is it worth trying to use `instantiateStreaming` instead of just `compileStreaming` since I think at least historically that's what browsers recommended? The `LoadWasm` intrinsic here could perhaps be replaced with `InstantiateWasm` where the imports are passed in.

crates/gen-host-js/src/lib.rs
```diff
@@ -875,11 +964,17 @@ impl Instantiator<'_> {
         iface: Option<&Interface>,
     ) {
         if exports.is_empty() {
-            self.src.js("{}");
+            if self.gen.opts.instantiation {
+                self.src.js("{}");
+            }
```
alexcrichton, line 969: Personally I would ideally prefer to have sort of a mish-mash sprinkling of `if self.gen.opts.instantiation` throughout since I feel that makes code difficult to read and change in the future. Would it be possible to instead change the return value of this function to something that works in both `--instantiation` and not modes? For example the `exports` here could return a form of `HashMap<???, String>` or something and then the caller would, once, check `instantiation` and create an export object or create a bunch of exported functions.

alexcrichton (commented):
Looks reasonable to me!

Could you expand on the longer-term story you see fo the `--instantiation`/etc flags? My experience in `wasm-bindgen` is that the flags for "what js environment are you targeting" can get out of hand quite quickly and become unwieldy to manage internally. The solution settled on over there was an all-encompassing `--target $foo` flag where there was no other need to have lots of other various flags.

Do you expect that the set of flags here will continue to grow? Are there lots of other targets that would grow custom flags that could have odd interactions with one another? Or perhaps would a `--target` flag make sense here as well?

tests/runtime/flavorful/host.ts
guybedford: Top-level await. Sorry forget it's not widely known!

tests/runtime/flavorful/host.ts
guybedford: This can't be lifted as it would create a top-level await execution cycle since the module imports the JS which imports this module (to fulfill the imports). We could just create another module, but it's quite nice to encapsulate the test in a single file as well.

crates/gen-host-js/src/lib.rs
```diff
@@ -368,8 +398,26 @@ impl Js {
...
+                self.src.js("
+                    const isNode = typeof process !== 'undefined' && process.versions && process.versions.node;
+                    let _fs;
+                    async function loadWasm (url) {
+                        if (isNode) {
+                            _fs = _fs || await import('node:fs/promises');
+                            return WebAssembly.compile(await _fs.readFile(url));
+                        }
+                        return fetch(url).then(WebAssembly.compile);
+                    }
```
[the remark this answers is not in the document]
guybedford: Well spotted. In theory you avoid one event loop tick by awaiting the promise in the async function instead of returning it. But it's completely negligible from any user perspective so that writing the shortest code is still the preference for me with these. Changing to be a non-async function for now.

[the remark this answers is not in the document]
guybedford: Actually the `import('node:fs')` does need the await, so I've left it as-is for now.

crates/gen-host-js/src/lib.rs
guybedford: Added, although as far as I can tell from the source formatter, it will remove that indentation anyway?

In theory if we wanted to instantiate the same core module multiple times, we would want the compiled but uninstantiated module. For this reason I quite like treating that as the go-to representation, and since it matches what will happen with import reflection as well.

It was just the sync former APIs which are not - that is `WebAssembly.instantiate` is not recommended when passing a buffer. But that it's overloaded form that takes a Module instance is fine to use despite the MDN warning, since the steps involved are equivalent from a spec perspective at least.

crates/gen-host-js/src/lib.rs
```diff
@@ -745,13 +814,33 @@ impl Instantiator<'_> {
...
+                self.gen.map_import(import_name),
+            );
+        }
+        uwrite!(
             self.src.js,
-            "const {callee} = imports.{}.{};",
-            import_name.to_lower_camel_case(),
-            func.name.to_lower_camel_case(),
+            "
+            function lowering{index}"
```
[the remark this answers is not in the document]
guybedford: The pattern I'm using for all intrinsics and top-level definitions since the formatting PR to ensure consistent double line spacing is to have a newline both before and after when they should be double spaced between eachother (and that's why there's the strip leading newline behaviour in the final write). Otherwise double spacing after each item leads to double spacing between items that intend to be single-spaced. I'm open to better layout suggestions though for sure! Updated to use `\n` for now.

crates/gen-host-js/src/lib.rs
```diff
@@ -745,13 +814,33 @@ impl Instantiator<'_> {
...
+                } else {
+                    format!("[\"{}\"]", import_specifier)
+                },
+                func.name.to_lower_camel_case(),
+            );
+        } else {
+            uwriteln!(
+                self.src.js,
+                "
+                    import {{ {} as {callee} }} from '{}';\
```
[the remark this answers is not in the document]
guybedford: Yes, there's no restriction they would be the first statement types. That said if we did have that restriction module parsing would have been a lot easier...!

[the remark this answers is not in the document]
guybedford: I do plan to add grouping to this output like we did with the C generator, just wanted to get the functionality right first.

crates/gen-host-js/src/lib.rs
```diff
@@ -2277,6 +2382,55 @@ fn to_js_ident(name: &str) -> &str {
     }
 }
 
+fn maps_str_to_map(maps: &str) -> Result<HashMap<String, String>> {
+    let mut map_hash = HashMap::<String, String>::new();
+    for mapping in maps.split(",") {
+        if let Some(eq_idx) = mapping.find('=') {
```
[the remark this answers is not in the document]
guybedford: Thanks.

guybedford:
In terms of the compat / target flags, one thing I don't like about versioning targets is it puts too much responsibility on the user to know their exact platform support, whereas having good defaults and some hints / options can get you 99% of the way.

I guess the model I'm thinking of is that the default output works in most environments, then by adding `--compat` you get something that works everywhere, or by digging into the specific `--x-compat` flags you can work out your own kind of compat matrix as necessary.

The features we're branching on here are currently really just:
* Top-level await support (~90% of browsers)
* fetch support (basically everywhere except Node.js)
* Node.js support (using readfile as a replacement for fetch)

That's really it as far as the current status is.

In future we might have import reflection to import the core modules instead of compileStreaming, but apart from that I think most other things would be at the level of core Wasm itself. Eg if we want to lean more heavily on new core Wasm features while still emitting legacy core modules and branching on that. But in theory even something like that could be captured by the current `--compat` system?

I'm not sure though, I may be going down that same path of progressive complexity here too, so would be interested your thoughts further on this and what you would have in mind for the above in a target-based system.

tests/runtime/flavorful/host.ts
guybedford: Turns out the `@ts-ignore` wasn't necessary.

crates/gen-host-js/src/lib.rs
guybedford: For now I've just exported the imports typings so they are at least available.

crates/gen-host-js/src/lib.rs
guybedford: By switching to namespaces there is now much more consistent handling between the modes. Overall I think this gating is a semantically important one as it represents the difference between exporting a namespace and a top-level declaration.

alexcrichton (approved):
Ok sounds reasonable to me, I'm happy to go with these flags for now and see how it ends up.

As part of this PR or as a follow up, would you be up for plumbing the new JS options into checkboxes and such for the demo page? The Wasmtime output has an example of this and I think it might be neat to get a demo online of the different outputs.

crates/gen-host-js/src/lib.rs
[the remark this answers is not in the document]
guybedford: I've added this hoisting now.

crates/gen-host-js/src/lib.rs
```diff
@@ -619,37 +694,82 @@ struct Instantiator<'a> {
...
+                } else {
+                    let load_wasm = self.gen.intrinsic(Intrinsic::LoadWasm);
+                    uwrite!(
+                        self.src.js,
+                        "{load_wasm}(new URL('./{name}', import.meta.url))",
+                    );
+                }
+            }
+        }
+        self.src.js.push_str("]);\n");
```
guybedford, line 743: I've added a new parallel instantiation here, and also switched `instantiateCore` to `compileCore` so that it can be called in parallel (since import objects otherwise require ordered instantiation). Please let me know if this model works for you?

guybedford:
I've added the demo UI options for "Compat" and "Custom Instantiation". Will merge when CI passes.

PR #408, opened 2022-10-29
Emit component name in core Wasm file names

guybedford:
This updates the generator to rather than outputting `module{n}.wasm` for the core Wasm files, to instead use the naming scheme `{name}.core{n}.wasm` and also in addition when `n == 0` to emit the `n`.

This way applications with a single Wasm file get an unnumbered Wasm, and the loaded Wasm file has a component-specific name eg in the network tab etc.

In addition this also fixes a one-liner regression in the Node.js compat output which wasn't using `compileStreaming`, and corrects the unhandled rejection handler attachment in the singular case for the JS generator.

### DIFF
--- a/crates/bindgen-core/src/component.rs
+++ b/crates/bindgen-core/src/component.rs
@@ -62,13 +62,7 @@ pub fn generate(
     // Insert all core wasm modules into the generated `Files` which will
     // end up getting used in the `generate_instantiate` method.
     for (i, module) in modules.iter() {
-        let i_str = if i.as_u32() == 0 {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Could this perhaps be threaded through to the generators in some form of a structure describing the names? That should avoid having three locations that all need to be kept in sync as the naming scheme gets fancier.

guybedford:
Good point, I've updated the PR to use a `core_file_name` trait function.

alexcrichton (approved):

PR #409, opened 2022-11-01
gen-host-js: Error wrapping and unwrapping for singular Result types

guybedford:
Updates the JS generator for functions that return a singular `Result` type to return the payload directly or throw the error payload if there was an error.

Conversely for imported functions, calls are now wrapped with a try-catch and errors are converted into intermediate result structs.

For both directions, the intermediate structs that are used should be possible to optimize away in future output. Tracking issue for this in https://github.com/bytecodealliance/wit-bindgen/issues/410.

The error type implemented here is the following when converting a `{ tag: 'err', val }` to a JS error:

```js
class ComponentError extends Error {
  constructor (payload) {
    super(payload);
    this.payload = payload;
  }
}
```

When in an exported function, the JS function stack will naturally have the throw position including the real and correct function name in the JS adapter code.

`toString` is implicitly applied to the error payload in the error message, which would result in `[object Object]` etc for non-primitive payloads. We could potentially explicitly branch on this and do something different although it is tough to know what would make sense here.

In lifting, the payload is extracted from the error with the following method:

```js
function getErrorPayload(e) {
  if (Object.prototype.hasOwnProperty.call(e, 'payload')) return e.payload;
  if (Object.prototype.hasOwnProperty.call(e, 'message')) return String(e.message);
  return String(e);
}
```

This composes with the component error, while also supporting common JS patterns of `throw new Error('any string')` forming an error string result type, allowing component synthesis without a full component error class via `throw Object.assign(new Error(), { payload: { any: 'data' } })` and also supporting legacy JS errors with as much observability from their message or string form to non-JS code to widely be able to handle a normal JS error as a `Result<ResultType, String>` in any other language.

This PR also fixes some minor formatting in the C generator and adds consistent single quoting in JS output. Suggestions / changes very welcome further.

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -221,12 +221,7 @@ impl WorldGenerator for C {
         if self.needs_string {
             uwrite!(
                 h_str,
-                "
-                   typedef struct {{
-                       {ty} *ptr;
-                       size_t len;
-                   }} {snake}_string_t;
-               ",
+                "typedef struct {{\n  {ty} *ptr;\n  size_t len;\n}} {snake}_string_t;\n",
                 ty = self.char_type(),
             );
         }
@@ -255,10 +250,7 @@ impl WorldGenerator for C {
         if self.return_pointer_area_size > 0 {
             uwrite!(
                 c_str,
-                "
-                   __attribute__((aligned({})))
-                   static uint8_t RET_AREA[{}];
-               ",
+                "\n__attribute__((aligned({})))\nstatic uint8_t RET_AREA[{}];\n",
                 self.return_pointer_area_align,
                 self.return_pointer_area_size,
             );
@@ -717,7 +709,7 @@ impl InterfaceGenerator<'_> {
         // canonical ABI.
         uwriteln!(
             self.src.c_adapters,
-            "__attribute__((export_name(\"{export_name}\")))"
+            "\n__attribute__((export_name(\"{export_name}\")))"
         );
         let import_name = self.gen.names.tmp(&format!(
             "__wasm_export_{}_{}",
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -219,6 +219,7 @@ impl WorldGenerator for C {
         c_str.push_str(&self.src.c_fns);
 
         if self.needs_string {
+            // Automatic indentation avoided due to `extern "C" {` declaration
             uwrite!(
                 h_str,
                 "typedef struct {{\n  {ty} *ptr;\n  size_t len;\n}} {snake}_string_t;\n",
@@ -248,6 +249,7 @@ impl WorldGenerator for C {
         // this for export bindings, because import bindings allocate their
         // return-area on the stack.
         if self.return_pointer_area_size > 0 {
+            // Automatic indentation avoided due to `extern "C" {` declaration
             uwrite!(
                 c_str,
                 "\n__attribute__((aligned({})))\nstatic uint8_t RET_AREA[{}];\n",
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -195,33 +195,43 @@ impl WorldGenerator for C {
             );
         }
 
-        let mut h_str = format!(
-            "#ifndef __BINDINGS_{0}_H\n\
-            #define __BINDINGS_{0}_H\n\
-            #ifdef __cplusplus\n\
-            extern \"C\" {{\n\
-            #endif\n\n",
+        let mut h_str = wit_bindgen_core::Source::default();
+
+        uwrite!(
+            h_str,
+            "#ifndef __BINDINGS_{0}_H
+            #define __BINDINGS_{0}_H
+            #ifdef __cplusplus
+            extern \"C\" {{",
             name.to_shouty_snake_case(),
         );
+
+        // Deindent the extern C { declaration
+        h_str.deindent(1);
+        uwriteln!(h_str, "\n#endif\n");
+
         self.include("<stdint.h>");
         self.include("<stdbool.h>");
 
         for include in self.includes.iter() {
             uwriteln!(h_str, "#include {include}");
         }
-        h_str.push_str("\n");
 
-        let mut c_str = format!("#include \"{snake}.h\"\n");
+        let mut c_str = wit_bindgen_core::Source::default();
+        uwriteln!(c_str, "#include \"{snake}.h\"");
         if c_str.len() > 0 {
             c_str.push_str("\n");
         }
         c_str.push_str(&self.src.c_fns);
 
         if self.needs_string {
-            // Automatic indentation avoided due to `extern "C" {` declaration
-            uwrite!(
+            uwriteln!(
                 h_str,
-                "typedef struct {{\n  {ty} *ptr;\n  size_t len;\n}} {snake}_string_t;\n",
+                "
+                typedef struct {{\n\
+                  {ty} *ptr;\n\
+                  size_t len;\n\
+                }} {snake}_string_t;",
                 ty = self.char_type(),
             );
         }
@@ -232,17 +242,17 @@ impl WorldGenerator for C {
         h_str.push_str(&self.src.h_fns);
 
         if !self.opts.no_helpers && self.src.h_helpers.len() > 0 {
-            h_str.push_str("\n// Helper Functions\n");
+            uwriteln!(h_str, "\n// Helper Functions");
             h_str.push_str(&self.src.h_helpers);
             h_str.push_str("\n");
         }
 
         if !self.opts.no_helpers && self.src.c_helpers.len() > 0 {
-            c_str.push_str("\n// Helper Functions\n");
+            uwriteln!(c_str, "\n// Helper Functions");
             c_str.push_str(self.src.c_helpers.as_mut_string());
         }
 
-        c_str.push_str("\n// Component Adapters\n");
+        uwriteln!(c_str, "\n// Component Adapters");
 
         // Declare a statically-allocated return area, if needed. We only do
         // this for export bindings, because import bindings allocate their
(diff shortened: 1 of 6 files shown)


guybedford:
Note this PR makes the `payload` object property on errors special in being a lifted interface. We could potentially use a more specific component model name or tagging, although `payload` feels relatively unique to me here. Another alternative would be to further tag errors by enforcing `err.tag === 'component_model_error'` or something equally specific. What is useful about the current output though is supporting the relatively straightforward - `throw new Error('any string error')` and `throw Object.assign(new Error(), { payload: { any: data } })`.

crates/gen-guest-c/src/lib.rs
```diff
@@ -221,12 +221,7 @@ impl WorldGenerator for C {
         if self.needs_string {
             uwrite!(
                 h_str,
-                "
-                   typedef struct {{
-                       {ty} *ptr;
-                       size_t len;
-                   }} {snake}_string_t;
-               ",
+                "typedef struct {{\n  {ty} *ptr;\n  size_t len;\n}} {snake}_string_t;\n",
```
alexcrichton, line 224: How come this changed? I personally find it easier to read the prior version and would prefer to keep that.

If all that's necessary is removing the leading newline then you can define the string as `"\` at the start which trims the leading newline.

(same for the adjustments below, I'd prefer to keep the multiline version if possible)

tests/runtime/flavorful/host.ts
```diff
@@ -63,7 +65,15 @@ export async function run () {
   assert.deepStrictEqual(wasm.fListInVariant2(), "list_in_variant2");
   assert.deepStrictEqual(wasm.fListInVariant3("input3"), "output3");
 
-  assert.deepStrictEqual(wasm.errnoResult().tag, 'err');
+  try {
+    wasm.errnoResult();
+    assert.ok(false);
+  }
+  catch (e: any) {
```
alexcrichton, line 72: Do you know if there's a way in typescript to annotate the type of the error that can be thrown so `catch` doesn't need type annotations?

tests/runtime/variants/host.ts
```diff
@@ -43,10 +43,17 @@ async function run() {
...
   const f = Math.fround(5.2);
-  assert.deepStrictEqual(wasm.roundtripResult({ tag: 'err', val: f }), { tag: 'err', val: 5 });
+
+  try {
+    wasm.roundtripResult({ tag: 'err', val: f });
+  } catch (e: any) {
+    assert.strictEqual(e.constructor.name, 'ComponentError');
+    assert.ok(e.message.includes('5'));
+    assert.strictEqual(e.payload, 5);
+  }
```
alexcrichton, line 56: Can this be updated to fail the test if the function does not throw?

crates/gen-host-js/src/lib.rs
```diff
@@ -421,6 +429,30 @@ impl Js {
...
+                ")
+            },
+
+            Intrinsic::ComponentError => self.src.js("
+                class ComponentError extends Error {
+                    constructor (payload) {
+                        super(payload);
+                        this.payload = payload;
+                    }
+                }
```
alexcrichton, line 453: Would it be best to remove the typescript mode for now? With the lack of import checking and new features like this aren't showing up in the typescript bindings I'm worried that the typescript support is falling pretty far behind. I was originally under the impression that typescript support was pretty important to have, but it seems like you're not using it much so it may be less important than I originally thought.

crates/gen-host-js/src/lib.rs
```diff
@@ -587,28 +619,31 @@ impl Js {
                         ptr = realloc(ptr, allocLen, 1, writtenTotal);
                     utf8EncodedLen = writtenTotal;
                     return ptr;
-                }
+                }\n
```
alexcrichton, line 622: FWIW I realize these are added from making the generated code look better, but from a maintainability point of view this is really hard to work with since there's just sort of random `\n` markers all sprinkled about without any particular rhyme or reason. Without tests whatever output is precisely done today is extremely likely to regress or change in the future. Personally I find that the readability is hindered with these sorts of markers because it makes me wonder why they're here at all which I have to actively combat with an understanding that it's just for generated code appearance at some point in the past.

crates/gen-host-js/src/lib.rs
```diff
@@ -1151,12 +1193,16 @@ impl<'a> JsInterface<'a> {
...
+                        if self.throw_ty_result {
+                            self.print_optional_ty(r.ok.as_ref());
+                        } else {
+                            self.needs_ty_result = true;
+                            self.src.ts("Result<");
+                            self.print_optional_ty(r.ok.as_ref());
+                            self.src.ts(", ");
+                            self.print_optional_ty(r.err.as_ref());
+                            self.src.ts(">");
+                        }
```
alexcrichton, line 1205: I think this may not be correct for something like `-> list<result<...>>` since it won't print the `result`?

This is sort of similar to the above question about removing typescript since it doesn't seem to be tested and is falling behind.

tests/runtime/flavorful/host.ts
guybedford: Unfortunately TypeScript only permits `any | unknown` for errors, it simply doesn't handle error type flow at all.

crates/gen-host-js/src/lib.rs
guybedford: I would say it's still very important to get TypeScript and documentation comments in TypeScript for the generated exports interface of the component, so that consumers get type hinting when importing the component. But that typing the component internals and imports could well be dropped.

crates/gen-host-js/src/lib.rs
guybedford: The rule I'm using for output is quite simple - in the generated output, top-level declaration blocks get double spacing. In generated functions, there is no unnecessary whitespace. No spaces more than a double line spacing. Then no leading whitespace, and trailing whitespace at the end of the file. An output linter rule could be configured to catch this I think.

crates/gen-host-js/src/lib.rs
guybedford: I'm using `throw_ty_result` as a single visitor marker more like an argument than a global state by immediately reverting its state after the call. See where this is being set. I'm open to actually treating it like an argument or better.

crates/gen-guest-c/src/lib.rs
guybedford: This is actually quite a specific output bug where the internal formatter needs a workaround. I've added a comment to give some more context. I think because of the `extern C {` brace being flattened, the source indentation system is getting confused. It was outputting:

```c
#ifndef __BINDINGS_LEXER_WORLD_H
#define __BINDINGS_LEXER_WORLD_H
#ifdef __cplusplus
extern "C" {
#endif

#include <stdlib.h>
#include "lexer_world.h"
#include <string.h>
#include <uchar.h>
#include <stdint.h>
#include <stdbool.h>

                   typedef struct {
                     char16_t *ptr;
                     size_t len;
                   } lexer_world_string_t;

typedef struct {
  bool is_some;
  lexer_world_string_t val;
} lexer_world_option_string_t;
```

It was only affecting that first declaration, and all subsequent indentation was correct in being at the base level (instead of the entire source being indented by the `extern C {` which I fixed in a previous PR.

With this change this fixes the output to properly indent:

```c
#ifndef __BINDINGS_LEXER_WORLD_H
#define __BINDINGS_LEXER_WORLD_H
#ifdef __cplusplus
extern "C" {
#endif

#include <stdlib.h>
#include "lexer_world.h"
#include <string.h>
#include <uchar.h>
#include <stdint.h>
#include <stdbool.h>

typedef struct {
  char16_t *ptr;
  size_t len;
} lexer_world_string_t;

typedef struct {
  bool is_some;
  lexer_world_string_t val;
} lexer_world_option_string_t;
```

Regressions could be avoided on this work by having proper code generation output snapshotting.

guybedford: The other corrections were actually a similar issue.

crates/gen-host-js/src/lib.rs
```diff
@@ -1217,7 +1263,11 @@ impl<'a> JsInterface<'a> {
         self.src.ts("): ");
         match func.results.len() {
             0 => self.src.ts("void"),
-            1 => self.print_ty(func.results.iter_types().next().unwrap()),
+            1 => {
+                self.throw_ty_result = func.results.throws(self.iface);
+                self.print_ty(func.results.iter_types().next().unwrap());
+                self.throw_ty_result = false;
```
guybedford, line 1269: See here for how `throw_ty_result` is treated as a single call state.

crates/gen-host-js/src/lib.rs
guybedford: Note that TypeScript is still being tested by the strong TSC checking in the tests.

crates/gen-guest-c/src/lib.rs
alexcrichton: Could this be fixed in `Source` rather than making the source less readable?

crates/gen-host-js/src/lib.rs
alexcrichton: That all makes sense but doesn't really address my point. I don't particularly care about the actual output of the generated code that much, the problem here is the source code itself. The source code has all these little changes sprinkled throughout it where the rhyme-and-reason is only apparent if you follow it through all the way to the end. Personally I don't think that it's the right balance since generated code is generated code and the source code here is pretty important to be readable.

I would prefer that, if possible, issues like this are fixed at the `Source` level to avoid the need to go through contortions at source code layer written in Rust.

crates/gen-guest-c/src/lib.rs
guybedford: Okay, I was able to resolve this by using the `source.deindent()` function.

crates/gen-host-js/src/lib.rs
guybedford: I will check more carefully before commenting - turns out your intuition was exactly right in that this newline was incorrect!

crates/gen-host-js/src/lib.rs
guybedford: Okay, in this case I believe we do need a better way to pipe this low-level information through eg via some core ABI information. I'd appreciate some advice on how to do that.

guybedford: I just inlined the type logic explicitly for now, avoiding the need for `throw_ty_result` tracking.

crates/gen-host-js/src/lib.rs
```diff
@@ -1253,7 +1293,18 @@ impl<'a> JsInterface<'a> {
         self.src.ts("): ");
         match func.results.len() {
             0 => self.src.ts("void"),
-            1 => self.print_ty(func.results.iter_types().next().unwrap()),
+            1 => {
+                if func.results.throws(self.iface) {
```
alexcrichton, line 1297: I'd recommend changing the signature of `throws` from returning a `bool` to returning `Option<(&Type, &Type)>` so you can avoid the `iter_types` and `if let` here to have it all bundled-in-one.

alexcrichton (approved):
👍

PR #411, opened 2022-11-02
gen-host-js: base64 option

guybedford:
Adds a `--base64` flag to the JS host generator for inlining the Wasm. ~~This is almost always a terrible idea and so the option has been hidden from the options display.~~ The option has been left in the help without warnings, so users can use it as appropriate.

It can be useful in certain contexts though, such as for es-module-lexer where the binary size is very carefully managed and a latency delay for a separate fetch during browser intialization wouldn't be acceptable. It doesn't add much code complexity overhead at all though thankfully.

### DIFF
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -66,8 +66,7 @@ pub struct Opts {
     )]
     pub instantiation: bool,
     /// Inline core WebAssembly modules as base64 strings.
-    /// Not recommended!
-    #[cfg_attr(feature = "clap", arg(long, hide = true))]
+    #[cfg_attr(feature = "clap", arg(long, conflicts_with = "instantiation"))]
     pub base64: bool,
     /// Comma-separated list of "from-specifier=./to-specifier.js" mappings of
     /// component import specifiers to JS import specifiers.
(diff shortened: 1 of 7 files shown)


crates/gen-host-js/src/lib.rs
```diff
@@ -48,35 +51,35 @@ struct Js {
...
             short = 'I',
             conflicts_with = "compatibility",
             conflicts_with = "compat"
         )
     )]
     pub instantiation: bool,
+    /// Inline core WebAssembly modules as base64 strings.
+    /// Not recommended!
+    #[cfg_attr(feature = "clap", arg(long, hide = true))]
+    pub base64: bool,
```
alexcrichton, line 71: I personally think it's ok to leave this here by default, I don't think we need to go out of our way to hide it. It's useful for at least one case so if it's there for that seems like others may find it useful too and it may as well not be proactively hidden.

alexcrichton (approved):
Can you add some tests for this as well?

crates/gen-host-js/src/lib.rs
guybedford: Ok, sure.

PR #412, opened 2022-11-02
gen-guest-c: weak attribute on post-return abi functions

guybedford:
I previously missed adding the weak attribute on these post-return functions as well, which is useful if you want to override the deallocation in the C guest generator.


alexcrichton (approved):
Can you add a test for this as well?

guybedford:
I've added a test with exports only that has a file size assertion on the base64 inlined output, which only passes with the change here and works quite nicely to avoid future file size regressions now as well.

Will merge once tests pass.

PR #413, opened 2022-11-03
syntax: world + interface

fibonacci1729:
**WIP**

This is the _initial_ implementation of [#83](https://github.com/WebAssembly/component-model/pull/83) and [#116](https://github.com/WebAssembly/component-model/pull/116) currently covering a subset of the proposed syntax, i.e. top-level `world` and `interface` items however `use` is currently disabled pending further design. 

An example of the syntax that is currently supported by this PR:
```
// one.wit

interface "consumer" {
    consume: func(arg: string)
}

interface "producer" {
    produce: func() -> string
}

world "component" {
    default produce-consume: interface {
        record info {
            v: string
        }

        produce: func() -> info
        consume: func(i: info)
    }
    import consumer: "consumer"
    import producer: "producer"
}
```

Signed-off-by: Brian H <brian.hardock@fermyon.com>


fibonacci1729:
cc @alexcrichton 

Also this code is pretty ugly at the moment but ill keep this PR up-to-date with changes as I churn.

crates/wit-parser/src/ast.rs
```diff
@@ -1,24 +1,329 @@
...
+
+        if self.worlds().count() > 1 {
+            let first = self.worlds().next().unwrap();
+            return Err(Error {
+                span: first.name.span,
+                msg: format!("too many worlds defined"), 
+            }
+            .into());
+        }
+        match self.worlds().next() {
```
alexcrichton, line 74: I might recommend modeling this as:

```rust
let mut worlds = self.worlds();
let world = match worlds.next() {
    None => { /* return error about no worlds */ }
    Some(world) => {
        if let Some(other) = worlds.next() {
            /* return error about `other` being second world
        }
        world
    }
};
// process `world`
```

crates/wit-parser/src/ast.rs
```diff
@@ -1,24 +1,329 @@
...
+                                },
+                            };
+                            if world.exports.insert(name.to_string(), interface).is_some() {
+                                return Err(Error {
+                                    span: span,
+                                    msg: format!("duplicate export name {}", name),
+                                }
+                                .into());
+                            }
+                        },
```
alexcrichton, line 156: This looks pretty similar to the import case so could this be refactored to a helper method?

crates/wit-parser/src/ast.rs
```diff
@@ -1,24 +1,329 @@
...
+            _ => None,
+        })
+    }
+
+    fn exports(&self) -> impl Iterator<Item = &Export<'a>> {
+        self.items.iter().filter_map(|item| match item {
+            WorldItem::Export(export) => Some(export),
+            _ => None,
+        })
+    }
```
alexcrichton, line 222: If possible I think it might be best to remove this style of iterators because it makes it harder to add more cases in the future and automatically have the compiler inform us what needs updating.

crates/wit-parser/src/ast.rs
```diff
@@ -1,24 +1,329 @@
...
+                Interface::parse_items(tokens).map(ExternKind::Interface)
+            },
+            other => {
+                Err(err_expected(tokens, "path, value, or interface", other).into())
+            }
+        }
+    }
+}
+
+#[allow(dead_code)] // TODO
```
alexcrichton, line 294: I think this may no longer be dead code?

alexcrichton (commented):
Looks reasonable to me! I think with tests passing this is a good first step towards integrating worlds.

alexcrichton:
Ok I've pushed up some more commits here for:

* The test suite should now be passing for all existing crates.
* All `wit-parser` tests are updated to use worlds
* Some basic world-using tests were added for `ui` and `ui/parse-fail`
* Added support for `default export ...` in `world`, at most once
* Removed all existing support for `use`
* Plumbed through `Docs` in a location or two

I think that this is now in a ready-to-go state. The plan after merging this is to, in parallel, remove `Interface::parse_file` and to continue to improve the text format. Notably the text format here is not intended to be final, it's a small step towards "let's have all the syntax in one text file". Further PRs will continue to make breaking changes to the text format and extend it, for example the removal of `use` is intended to be temporary while "more official" semantics are worked out.

sunfishcode:
@alexcrichton `use` is used in a few places, but only to import `u32` type aliases which previously were handle types, so we could just duplicate those type aliases in each wit that needs them for now.

alexcrichton:
Ok makes sense, in that case if it's ok I think it's best to land this and then fork off the work for `use`.

PR #414, opened 2022-11-04
gen-host-js: Top-level await compatibility via $init promise

guybedford:
This adds a new compatibility feature, `--tla-compat` which is enabled under the `--compat` flag as well. Top-level await support is wide, but still gives some gap on browser compatibility (around 5% of browsers support modules but not TLA), so is still useful for maximum JS support.

When used, all top-level Wasm loads and instantiations that rely on `await` get wrapped in a single `$init` promise, which is exported and must be awaited before any component functions can be called. This only applies to ESM instance mode and not `--instantiation`.

If calling any exported component functions before initialization has been completed, an error is thrown informing that the `$init` promise must first be awaited. The `$init` export name is used as it should be an invalid world export name I believe.

### DIFF
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -290,31 +290,42 @@ impl ComponentGenerator for Js {
             output.push_str(";");
         }
 
-        output.push_str(&self.src.js_intrinsics);
-
-        if self.opts.tla_compat {
-            let init_str: &str = &self.src.js_init;
-            uwrite!(
-                output,
-                "
-                let _initialized = false;
-                export const $init = (async() => {{
-                    {init_str}
-                    _initialized = true;
-                }})();
-
-            "
-            );
-            uwriteln!(
-                self.src.ts,
-                "
-                export const $init: Promise<void>;"
-            );
+        let (maybe_init_export, maybe_init) = if self.opts.tla_compat {
+            uwriteln!(self.src.js_init, "_initialized = true;");
+            (
+                "\
+                    let _initialized = false;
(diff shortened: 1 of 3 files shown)


alexcrichton (commented):
This seems like a reasonable feature, but the internals feel pretty heavy-handed with lots of if/else strewn throughout code generation. Instead of this structure could something be generated along the lines of:


```js
async function init() {
    // current init, always using `await`
}

// if --tla-compat
init()

// if not-compat
await init()
```

That way I think just the exports would get the `let`-vs-otherwise definition treatment?

guybedford:
@alexcrichton the problem is I don't know of a single JS optimizer short of closure that will unwrap `async function init () {}; await init()`. Byte size in JS is important and a feature of this output.

guybedford:
I've managed to remove one branch with an update commit. The remaining 6 branches are for:
* One branch for the $init function itself
* One branch for exported functions to throw if called when unitialized
* Three branches to hoist the `memory`, `realloc` and `postReturn` variable definitions from const function definitions to out-scope let definitions. I'd prefer not to collapse this branch since that would involve removing const declarations which could be useful optimization information for an optimizer on non-tla-compat source
* One branch to vary the WebAssembly instantiation routine from setting the local const versus global let variable in TLA compat.

With the current output size constraints I'm not sure I can simplify more than this, but I'm open to refactoring suggestions.

guybedford:
Ok, I've unified on just the single branch as discussed, leaving any unwrapping to rather be the job of an optimizer.

alexcrichton (approved):

PR #415, opened 2022-11-05
gen-host-js: --valid-lifting-optimization

guybedford:
This adds a new `--valid-lifting-optimization` flag to the JS generator to disable the runtime checks of the Wasm data structures. The assumption being that when the interfaces have been fully tested, this can be enabled in production for a significant reduction in output JS file size.

We could could possibly treat this as an explicit optimization or production option. Checks of user inputs should still remain though.

I've based this branch to the `--tla-compat` PR at https://github.com/bytecodealliance/wit-bindgen/pull/414 so it should land after that.

I've also started to unify on the `uwriteln!` / `uwrite!` patterns that feel a bit cleaner than the source functions, so am refactoring those as I go, let me know if you agree that these look better or have further suggestions on the refactoring. Would be nice to converge on code style with the formatting.

### DIFF
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -2016,14 +2016,12 @@ impl Bindgen for FunctionBindgen<'_> {
                     );
                 }
                 let variant_name = name.to_upper_camel_case();
-                if !self.gen.opts.valid_component_optimization {
-                    uwriteln!(
-                        self.src.js,
-                        "default: {{
-                            throw new TypeError('invalid variant specified for {variant_name}');
-                        }}"
-                    );
-                }
+                uwriteln!(
+                    self.src.js,
+                    "default: {{
+                        throw new TypeError('invalid variant specified for {variant_name}');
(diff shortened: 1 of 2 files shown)


alexcrichton:
One area I think it's fine to relax checks on is incoming JS values in the sense that those are embedder-specific and in theory already correct. I believe this PR is also relaxing checks for values coming out of the component itself, though, and I don't believe that's something that we can compromise on. The purpose of wit-bindgen is to be a component runtime polyfill and to implement that correctly many of these checks are required (e.g. variant discriminants all valid)

guybedford:
I've rebased this PR and changed the flag name to `--valid-component-optimization` where these optimizations are based on the assumption that the component being consumed is a safe trusted component binary that is known to produce valid outputs.

alexcrichton (approved):

guybedford:
To be completely clear about when this applies I've updated the flag name to `--valid-lifting-optimization` and removed all branches which did not apply to this case.

PR #416, opened 2022-11-05
host-gen-js: Output optimization

guybedford:
This is based to https://github.com/bytecodealliance/wit-bindgen/pull/415, which is in turn based to https://github.com/bytecodealliance/wit-bindgen/pull/414, so should land after those.

Updates the output to be more friendly to minifiers resulting in a smaller total JS file size - arrow functions are used for intrinsics, and the `dataView(memory)` pattern is replaced with a dedicated `dataView{n}` binding per memory, which should be able to be optimized out if not used for a particular memory.

### DIFF
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -85,6 +85,9 @@ pub struct Opts {
     /// via an async $init promise export to wait for instead.
     #[cfg_attr(feature = "clap", arg(long, group = "compatibility"))]
     pub tla_compat: bool,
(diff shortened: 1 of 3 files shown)


alexcrichton:
This looks reasonable, but the `dataView` optimization I don't think will work because `memory.grow` can invalidate the prior `ArrayBuffer` so JS needs to always check to see if the `ArrayBuffer`s match to get the most up-to-date view of memory.

guybedford:
Ah, ok I can't come up with a better alternative, will close this then.

PR #417, opened 2022-11-08
chore: fix clippy warnings

willemneal:
Currently only fixed and check for default lints. If there was a reason for not including clippy feel free to ignore.

I also added allows where it seemed fit; should those be blanket allows?

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -37,9 +37,10 @@ pub struct Opts {
 
 impl Opts {
     pub fn build(&self) -> Box<dyn WorldGenerator> {
-        let mut r = C::default();
-        r.opts = self.clone();
-        Box::new(r)
+        Box::new(C {
+            opts: self.clone(),
+            ..Default::default()
+        })
     }
 }
 
@@ -476,7 +477,7 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
         self.src.h_defs(" ");
         self.print_typedef_target(name);
 
-        if flags.flags.len() > 0 {
+        if !flags.flags.is_empty() {
             self.src.h_defs("\n");
         }
         for (i, flag) in flags.flags.iter().enumerate() {
@@ -513,7 +514,7 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
         self.src.h_defs("} ");
         self.print_typedef_target(name);
 
-        if variant.cases.len() > 0 {
+        if !variant.cases.is_empty() {
             self.src.h_defs("\n");
         }
         for (i, case) in variant.cases.iter().enumerate() {
@@ -596,7 +597,7 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
         self.src.h_defs(" ");
         self.print_typedef_target(name);
 
-        if enum_.cases.len() > 0 {
+        if !enum_.cases.is_empty() {
             self.src.h_defs("\n");
         }
         for (i, case) in enum_.cases.iter().enumerate() {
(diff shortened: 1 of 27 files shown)


pchickey:
We don't tend to use clippy on Bytecode Alliance projects, because we find that it doesn't increase code quality enough to justify the churn it adds to CI and local workflows. So, I think we will decline this contribution.

peterhuene:
Although I tend to run clippy as my rust-analyzer command in my IDE so my own code is clean from linter warnings before I commit, I agree that we don't want to enforce this at the CI-level.

Given I see the existing linter warnings as I edit, I sometimes fix them in files I'm working in, but generally leave the existing ones alone.

willemneal:
How about removing the addition to CI and accepting the other changes?  I do find that most of the changes improve readability.

crates/wit-parser/src/lib.rs
```diff
@@ -305,7 +309,7 @@ pub struct Function {
     pub results: Results,
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
pchickey, line 312: I believe what the PartialEq and not Eq constraint was expressing here is that equality is only valid on items from the same "world", and if you compare across worlds equality isn't necessarily meaningful.

crates/gen-guest-c/src/lib.rs
```diff
@@ -1623,7 +1626,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 for op in operands {
                     uwriteln!(result, "{},", op);
                 }
-                result.push_str("}");
+                result.push('}');
```
pchickey, line 1629: I don't think these substitutions of push_str("_") to push('_') are worthwhile, they make it less consistent to read.

crates/gen-guest-c/src/lib.rs
```diff
@@ -2306,6 +2309,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
 }
 
 #[derive(Default, Clone, Copy)]
+#[allow(clippy::enum_variant_names)]
```
pchickey, line 2312: Lets not put these in

crates/gen-rust-lib/src/lib.rs
```diff
@@ -5,7 +5,7 @@ use std::iter::zip;
 use wit_bindgen_core::wit_parser::abi::{Bitcast, LiftLower, WasmType};
 use wit_bindgen_core::{wit_parser::*, TypeInfo, Types};
 
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
pchickey, line 8: Let's not add any Eq constraints

crates/test-helpers/macros/build.rs
```diff
@@ -168,14 +162,8 @@ fn main() {
...
-            let deps = fs::read_to_string(&dep_file).expect("failed to read dep file");
-            for dep in deps
-                .splitn(2, ":")
-                .skip(1)
-                .next()
-                .unwrap()
-                .split_whitespace()
-            {
+            let deps = fs::read_to_string(dep_file).expect("failed to read dep file");
+            for dep in deps.split_once(':').unwrap().1.split_whitespace() {
```
pchickey, line 166: I don't think this transformation is correct - we want the split whitespace after the second `:`

pchickey (commented):
Some of these changes are fine but some I think decreased consistency and readability, and I think I found one place the transformation is incorrect. In general I don't find it worthwhile to go through PRs like this to see which parts I like and which parts I don't, because its mostly bike-shedding.

willemneal:
Thank you @pchickey! I know this kind of PR can be annoying, but hopefully after removing a lot of the unneeded changes it will be worth it!

pchickey (approved):

alexcrichton:
This is quite old at this point with lots of merge conflicts and a lot has moved around in the meantime, so I'm going to close this. If you'd like though feel free to reopen.

PR #421, opened 2022-11-08
Implement exporting interfaces from adapter modules

alexcrichton:
This commit rounds out the support for adapter modules in `wit-component` by implementing support for exported interfaces in addition to the prior support for imported interfaces. This commit means that a component-model-ignorant `*.wasm` can now be fully adapted, both for imports and exports, into a component. The main motivation for this is to bridge the legacy `wasi_snapshot_preview1` command/reactor modes into the component model, for example defining a new:

```
command: func(args: vec<string>)
```

and removing host imports for things such as env vars and arguments, instead implementing them in the adapter.

The support in this commit largely "simply" lifts the restriction of adapters not having exported interfaces. This is now allowed and everything is wired up as:

* All exported interfaces from adapter modules will get exported on the main component as well. The function-to-lift comes from the adapter instances themselves.
* Adapters specify one realloc function for imports and one for exports. These are now called `cabi_import_realloc` and `cabi_export_realloc`. There's nothing necessarily "canonical" about this besides the signature, but that's the name for now.
* Adapters can import functions from the main module through the `__main_module__` import string. This enables the `command` adapter export to import the `_start` function from the main module, for example.

Some various tests have been added to showcase this. The rough workflow for WASI is intended to look like:

1. The adapter defines a `command` export
2. The adapter, as part of the `command` export, receives args, env vars, initial descriptors, etc.
3. The adapter will configure in-memory one-time-allocated otherwise "static" state as part of the `command` "startup"
4. The adapter will import `_start` from the main module and then call it
5. The main module might call `args_get`, which goes back to the adapter, which reads state set up by the `command` entrypoint.

The realloc for this adapter is expected to use `memory.grow` to allocate memory. This "leakage" is expected to be ok because the memory lives for the entire lifetime of the component and otherwise does not need to be deallocated.


pchickey (approved):
Looks good. I appreciate writing the tests in wat instead of in the wasi test shim. There is still some work on the rust-guest bindgen to get that side working, but I'll follow up with that shortly!

PR #422, opened 2022-11-09
Add the ability to skip bindings in Rust

alexcrichton:
This commit adds a `skip` attribute and CLI argument for skipping generation of bindings for a set of named functions. This can be useful when the generated bindings can't be used for a reason such as they're to high-level, instead allowing applications to manually define functions which work with the raw inputs of the canonical ABI.


alexcrichton:
@pchickey you'll be interested in this, my intention is that for the `preview1.wasm` we'll use this to skip generating bindings for `command` (or something like this) to write the raw signature and work with that rather than using the generated bindings which would do different things for `Vec<T>` and such.

PR #423, opened 2022-11-09
gen-host-js: support direct ArrayBuffer for Uint8Array lists

guybedford:
This adds support for the JS lowering of `list<u8>` to support either a `Uint8Array` or a direct `ArrayBuffer` as well.

Discovered when piping a fetch response array buffer to a component input, which I think should be easily supportable here without additional wrapping being needed.

### DIFF
--- a/crates/gen-host-js/src/lib.rs
+++ b/crates/gen-host-js/src/lib.rs
@@ -2508,10 +2508,17 @@ impl Bindgen for FunctionBindgen<'_> {
                     "const ptr{tmp} = {realloc}(0, 0, {align}, len{tmp} * {size});"
                 );
                 // TODO: this is the wrong endianness
-                uwriteln!(
-                    self.src.js,
-                    "const src{tmp} = new Uint8Array(val{tmp}, val{tmp}.byteOffset, len{tmp} * {size});",
-                );
+                if matches!(element, Type::U8) {
+                    uwriteln!(
+                        self.src.js,
+                        "const src{tmp} = new Uint8Array(val{tmp}.buffer || val{tmp}, val{tmp}.byteOffset, len{tmp} * {size});",
+                    );
+                } else {
+                    uwriteln!(
+                        self.src.js,
+                        "const src{tmp} = new Uint8Array(val{tmp}.buffer, val{tmp}.byteOffset, len{tmp} * {size});",
+                    );
+                }
                 uwriteln!(
                     self.src.js,
                     "(new Uint8Array({memory}.buffer, ptr{tmp}, len{tmp} * {size})).set(src{tmp});",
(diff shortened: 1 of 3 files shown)


tests/runtime/lists/host.ts
```diff
@@ -11,7 +11,7 @@ async function run() {
     testwasi: helpers,
     imports: {
       emptyListParam(a) {
-        assert.deepStrictEqual(Array.from(a), []);
+        assert.deepStrictEqual(Array.from(a as Uint8Array), []);
```
guybedford, line 14: This is quite an important consideration actually - basically because "inputs" are from users, we have a more relaxed definition for _lowered input types_ in JavaScript than we do for _lifted output types_. I think we need to more carefully take this into account in the generation process, so that we could relax the input type to `ArrayBuffer | Uint8Array` but at the same type guarantee that the output type is always `Uint8Array`. Any pointers on setting up such a structure welcome, but there were no easy references to use and I didn't want to make timing assumptions.

alexcrichton (commented):
The generator isn't really set up for this, but would you be up for updating the type printing to print `Uint8Array | ArrayBuffer` only for values flowing into wasm, and `Uint8Array` for values coming out of wasm? Since `ArrayBuffer` is never returned it seems unfortunate to force the new `as` casts in the typescript tests.

tests/runtime/lists/host.ts
alexcrichton: I think this could be solved by threading through a `Mode` parameter or something like that throughout `print_ty`, where it's either `Mode::Lower` or `Mode::Lift` which would enable the ability to print different types.

guybedford: One problem with this approach I believe is that defined types won't get the distinction.

guybedford:
Yeah, it's not ideal but I've threaded the mode through to fix this. As mentioned this doesn't currently carry through to defined types unfortunately. Perhaps that is good enough for now though.

crates/gen-host-js/src/lib.rs
```diff
@@ -1265,6 +1280,12 @@ impl Instantiator<'_> {
     }
 }
 
+#[derive(Copy, Clone)]
+enum Mode {
+    Internal,
+    External,
```
alexcrichton, line 1286: Could this be renamed as "Lift" and "Lower" perhaps? I'm not sure what internal/external mean in this context myself.

alexcrichton (approved):
That's true, yeah, that defined types won't work as well. That's why the Rust bindings, for example, could generate multiple types for one input time. For example exports taking an argument and imports taking the same argument would take different values in Rust, one representing ownership and one representing borrowing. That's a lot of infrastructure to pull in here though which probably isn't necessary yet.

PR #425, opened 2022-11-10
host-gen-js: replace --load with threshold-based granular inlining

guybedford:
Replaces the `--load` flag with a single option to determine base64 inlining - `--base64-cutoff=bytesize`. Disabling inling is available via `--base64-cutoff=0` and you can enforce it with a larger cutoff, which seems better to be explicit about size limits. It applies granularly per-core-wasm source so that the small trampoline modules can be inlined where appropriate.

The default `--base64-cutoff` is set to 2500 for relatively small inlining only.

For `--load=custom`. `--instantiation` was able to fulfill this asm.js wrapping use case via a new optional `instantiateWasm` argument to `instantiate` which defaults to `WebAssembly.instantiate`.

### DIFF
--- a/crates/bindgen-core/src/lib.rs
+++ b/crates/bindgen-core/src/lib.rs
@@ -245,6 +245,13 @@ impl Files {
         }
     }
 
+    pub fn get_size(&mut self, name: &str) -> Option<usize> {
+        match self.files.get(name) {
+            Some(data) => Some(data.len()),
+            None => None,
+        }
+    }
+
     pub fn remove(&mut self, name: &str) -> Option<Vec<u8>> {
         return self.files.remove(name);
     }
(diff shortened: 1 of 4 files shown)


guybedford:
I was thinking some more about how a transition to ES modules might look and I think the `--load` flag actually doesn't make that much sense if we are making auto the default mode now. Instead you can disable base64 output via `--base64-cutoff=0` and you can enforce it with a larger cutoff, which seems better to be explicit about size limits.

Finally, for `--load=custom` I was able to replicate that with `--instantiation` rather to avoid the need for the flag entirely. What I did need to do though was add a new argument to the `instantiate` with an optional `instantiateWasm` which defaults to `WebAssembly.instantiate`.

The instantiate signature in `--instantiation` is therefore now `function instantiate (compileCore, imports, instantiateCore = WebAssembly.instantiate)`. The ordering might look a little odd, but was done for backwards compatibility. If backwards compatibility is not a concern we can move to any other ordering as suits, although I'm fine with the above since `instantiateCore` would be overridden only rarely.

Hopefully the above aligns with keeping unnecessary output complexity to a minimum.

alexcrichton (approved):

PR #430, opened 2022-11-15
Remove usage of `Interface::parse_file` in `wit-component` tests

alexcrichton:
This commit migrates the internals of `wit-component` as well as all tests over to using worlds from `wit-parser` instead of just interfaces. Lots of test files changed here but the goal was to have relatively few actual golden output changes. Code-wise the majority of the changes are supporting new features of worlds, such as import/exporting with a different name than is declared on the interface.


pchickey (approved):
Hell Yeah

PR #431, opened 2022-11-15
Expose `macro_export` in guest-rust-macro

AmitPr:
This PR exposes the option to pass `export_macro` as an argument, which allows one to do something like:
```rust
wit_bindgen_guest_rust::generate!({
    default: "./wit/http_service.wit",
    name: "http_component",
    macro_export,
});
```

For context, setting the `macro_export` option to `true` adds a `#[macro_export]` to the generated `export_{module_name}` macro.


alexcrichton:
Thanks!

PR #432, opened 2022-11-15
Migrate all tests to using world-file-syntax

alexcrichton:
This commit migrates all of `tests/codegen/*.wit` and `tests/runtime/**/*.wit` to using world files rather than per-file interfaces. After this there's no more in-tree usage of `Interface::parse_file` and I'll remove that as a follow-up.

Most of the changes here are mechanical changes to the tests, but various code generators also received test fixes to support new features that are now tested with worlds. For example generators inconsistently tested import/export/default interfaces and now the codegen test worlds all exercise all of these features.


pchickey (approved):
Very much mechanical so I spot-checked and trust CI with the rest.

Issue #395, opened 2022-10-25
Spin out host generators into their own repositories


alexcrichton:
This was discussed at the wit-bindgen meeting last Friday, and after some futher discussion with @pchickey I've come up with a proposal which I personally feel is the direction that wit-bindgen should go in. I wanted to open an issue though to get feedback from others, however.

To recap, the core issue at hand here is that the Wasmtime repository would like to vendor an implementation of `wasi-preview2`. This would necessitate Wasmtime depending on `wit-bindgen` for host bindings generation, however, but `wit-bindgen` depends back on the Wasmtime repository. This circular dependency between git repositories cannot be resolved and isn't a maintainable way to implement this code.

After much delibration and possible ways to go about this, I would like to propose that all host generators be "split out" of the `wit-bindgen` repository into their own repositories, in time possibly doing the same thing for guests. Concretely what I'm envisioning is:

* Move out `wit-component` and `wit-parser` into the `wasm-tools` repository. This sort of low-level tooling is the "base" of the dependency tree and would live next to all our other low level tools. We could create commands such as `wasm-tools wit-validate` or similar to validate the structure of `*.wit` files or generate the markdown documentation that's there today. No host/guest generator would live in the `wasm-tools` repository. Tests here would solely be about name resolution and syntactical correctness of `*.wit` files plus the `*.wat` tests that `wit-component` has today.
* Move the host wasmtime rust generator to the Wasmtime repository. This would depend on the crates from `wasm-tools` and would be tested with raw `*.wat` inputs, unlike today where it's tested with compiled guest inputs.
* Move the host wasmtime python generator to the `wasmtime-py` repository (exact location and naming TBD).
* Move the host JS generator to its own repository with naming TBD. The idea here is that this might make more sense as a hub for component tooling compiled to wasm to use in JS, for example `wit-bindgen` as-is today could be compiled to a `*.wasm` and published to NPM from this location.
* Leave the guest generators inside of the `wit-bindgen` repository, for now. While these may eventually move to their own repository it seems like it may make the most sense to leave them in one location now for discoverability and in general there's no pressing need for splitting them out. Guest generators would be tested with a host generator.
* Finally use the published-to-crates-io version of the rust guest generator to test the wasi-preview2 implementation that will live inside of the Wasmtime repository.

[...]

PR #436, opened 2022-11-18
Remove `wit-bindgen-gen-host-wasmtime-py`

alexcrichton:
This commit removes the host Python generation from this project. The support here has all moved upstream to
https://github.com/bytecodealliance/wasmtime-py and is now supported through:

    python -m wasmtime.bindgen my-component.wasm --out-dir ./my-component

Some documentation is in the upstream `README.md` about how to use. Necessary tests have been mirrored to the upstream repository to the best of my ability, and this is intended to represent an inflection point where the growth of the Python generator is no longer tied up with all the other languages in this repository. This is a step further along the path of #395.


alexcrichton:
The upstream wasmtime-py PR is https://github.com/bytecodealliance/wasmtime-py/pull/104 and once this merges I'll work on migrating issues from this repository to that repository.

pchickey (approved):
They grow up so fast!!

PR #440, opened 2022-11-23
gen-guest-c: define OK constant for scalar enum results

guybedford:
Defines a new `WORLD_RESULT_OK` constant in the case where there is scalar enum result flattening.

It may make sense to have dedicated helper functions for the particular enum types being returned, but this should cover the use case for now at least. The result structs are also still being generated although unused.

Also fixes some whitespace handling.

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -21,7 +21,6 @@ struct C {
     return_pointer_area_align: usize,
     names: Ns,
     needs_string: bool,
-    result_enums: bool,
     world: String,
 }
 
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):
Could this have a test as well?

guybedford:
Added a test, also found in further testing that there is a risk of bad casts here, so instead of emitting this as a generic constant, it is now a per-enum constant at the exact correct value instead of an implicit `-1`. There wasn't an easy way to work out if it is always used or not so it will always be output now for all enums in the result case as `#define WORLD_RESULT_OK_ENUM_NAME 255` or whatever value applies.

alexcrichton (approved):

alexcrichton:
Er actually, I'm not sure if this works? If an enum has precisely 255 values (or 256? not sure) then I don't think the placeholder works since it'll collide with the last variant?

guybedford:
@alexcrichton currently `-1` is being returned for the enum OK slot, so I had assumed that the lost slot was already being reserved?

guybedford:
I did a quick test against a 255 entry enum, and it does correctly expand the type without colliding with the `max - 1` value.

PR #443, opened 2022-11-25
gen-guest-c: c strings pass by value

guybedford:
This is an exploratory PR to see if it might be worth considering a value struct pattern for strings and possibly other two value struct types like lists, optionals and results when passed as parameters.

In theory this avoids memory operations, but I'm not sure how much benefit there really is. See the test changes for what this means in the APIs, it does feel slightly cleaner to me though.

Feedback very welcome, I can explore this path further / drop it as makes sense.

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -1394,23 +1394,13 @@ impl InterfaceGenerator<'_> {
 
     fn free(&mut self, ty: &Type, expr: &str) {
         if self.is_string(ty) {
-            self.src.c_helpers("free(");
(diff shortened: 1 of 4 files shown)


alexcrichton (commented):
I don't personally feel like I have a good enough grasp on C idioms to say this is the right way or it's not. All my historical experience points to "nothing bigger than a pointer is passed by-value" but I don't exactly have a wealth of experience to draw from.

If this is merged, however, then I think the same treatment should be applied to lists since they're more-or-less the same in this respect to strings.

guybedford:
Yes, it turns out C is not smart enough to do any flattening optimizations and the generated Wasm is actually identical.

PR #444, opened 2022-11-25
bug: multiple ret areas

guybedford:
This demonstrates a bug, where for an imported function with a big enough record and result return, multiple ret areas end up being defined in that function.

Test case included demonstrating the issue, with a fix for the C generator case, but not the Rust host yet.

Guidance would be appreciated if this seems like the right sort of track to fix or not.


alexcrichton:
Ah I had forgotten that return areas were requested twice when merging https://github.com/bytecodealliance/wit-bindgen/pull/326. I think the fix here will be to avoid generating the return area until after the function has completed bindgen. Essentially this should be the same as for not-imports which is to save off how big the area is requested to be and have it generated later.

Issue #449, opened 2022-12-08
gen-guest-c: remove result enum flattening


guybedford:
The current result enum flattening should probably be removed, as per the discussion in https://github.com/WebAssembly/component-model/pull/139. The primary reason being that this overload in theory can overflow with the enum value space.

PR #450, opened 2022-12-09
gen-guest-c: scalar enum result -> bool return, ok & err ptrs

guybedford:
This updates the result error enum handling to return a boolean instead of a flattened enum, with two possible return pointers - one for the success value and one for the error value.

It does seem slightly higher bandwidth in comparison to what we had before, in that there are three values and three writes instead of two values and two writes now.

I'm going to test out this diff in the js-compute-runtime bindings work, and see how noisy it looks.

The alternative is to completely unflatten the result and just have the double pointer situation from result -> ok / err pointer writes... still somewhat on the fence between the two approaches.

Fixes https://github.com/bytecodealliance/wit-bindgen/issues/449.

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -62,7 +62,7 @@ struct CSig {
 enum Scalar {
     Void,
     OptionBool(Type),
-    ResultEnum { err: TypeId, max_err: usize },
+    Result(bool),
     Type(Type),
 }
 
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
I think it'd also be reasonable to take a union out-ptr (C union) for the return payload as well rather than two, but either way works IMO.

guybedford:
I've updated the PR in the latest commit to avoid double writing to both the err and ok return pointers using a singular store function approach.

alexcrichton (approved):

PR #454, opened 2022-12-24
gen-guest-rust, gen-gueste-c: Support multiple `RetArea`s in a single function.

sunfishcode:
Add a suffix to the generated `RetArea`/`ret_area` declarations so that there are multiple occurrences in the same scope, they don't collide.


sunfishcode:
This may be obsoleted by the fix discussed in https://github.com/bytecodealliance/wit-bindgen/pull/444/.

alexcrichton:
If ok yeah I think it would be best to implement the solution from https://github.com/bytecodealliance/wit-bindgen/pull/444#issuecomment-1329288046 for this.

PR #457, opened 2023-01-19
Update the wit-{component,parser} crates

alexcrichton:
This commit updates `wit-bindgen` for all the changes made in https://github.com/bytecodealliance/wasm-tools/pull/867, namely the implementation of `use`. The syntax of WIT has changed to support this and additionally the internal structures of WIT have all radically changed.

This change updates all the various generators here, tests with various syntaxes, and all `*.wit` files.


alexcrichton:
Note that I'm still working my way through this. The Rust guest and JS host generators are all fully updated and working as per the tests now. I plan on tackling C & Java next.

alexcrichton:
Ok I think almost everything is ready here except for the java generator. 

@dicej would you be up for helping out getting that working? The generator generates code successfully currently but it doesn't compile. I fear that this is diving into an area of Java I know very little about. I did change the structure of the generated code so I know why it's failing to compile, namely previously many tests used `default export` and now they have both `export foo: some.interface` as well as `export test-imports: func()` so tests are testing more than they used to. I'm not sure how to get the right Java incantation with classes and such to update the `wasm.java` to have the whole test. I realize this is a bit to just dump on you, so if you'd like I'd be happy to pair up for a moment to see if you can look over my shoulder and point out what to you may be an obvious fix.

alexcrichton:
Ok with the help of @dicej I think this is ready go go now!

crates/gen-guest-teavm-java/tests/codegen.rs
```diff
@@ -4,18 +4,21 @@ use std::path::Path;
 use std::process::Command;
 
 macro_rules! codegen_test {
-    ($name:ident $test:tt) => {
+    // TODO: should get this test working
+    (use_across_interfaces $($x:tt)*) => {};
```
alexcrichton, line 8: Oh @dicej I also meant to point this out to you, as your assistance would be appreciated in getting this test working. This has to do with importing things from different paths which I'm less certain how it works in Java. Again happy to have you walk me through it over Zoom or similar or can provide more details myself as well.

dicej: Sure, I'll take a look -- might need to wait until Monday, though.

peterhuene (approved):
Looks great 👍 One tiny nit commit.

PR #458, opened 2023-01-23
fix cross-interface symbol resolution in Java guest generator

dicej:


dicej:
@alexcrichton I'm not sure this is 100% correct for all the ways symbol resolution and aliasing can happen, but at least all the current tests are passing.  Feel free to ping me if future tests reveal any issues.

alexcrichton:
Looks reasonable enough to me, thanks!

PR #460, opened 2023-01-24
gen-host-js: fix ts compilation bug in newer ts versions

guybedford:
This fixes a TS compilation bug that was coming up. An alternative could be to check in the package lock to ensure consistent tsx versioning.

### DIFF
--- a/crates/gen-host-js/tests/helpers.ts
+++ b/crates/gen-host-js/tests/helpers.ts
@@ -2,6 +2,7 @@
 import { readFile } from 'node:fs/promises';
 // @ts-ignore
 import { argv, stdout, stderr } from 'node:process';
+// @ts-ignore
 import { Buffer } from 'node:buffer';
 
 // This is a helper function used from `host.ts` test in the `tests/runtime/*`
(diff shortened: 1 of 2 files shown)


guybedford:
The error with the TypeScript change in the second CI run became:

```
Error: ../../target/runtime-tests/js/variants_rust/helpers.ts(20,16): error TS2580: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
	../../target/runtime-tests/js/variants_rust/helpers.ts(23,16): error TS2580: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
```

While the `@types/node` package is definitely present in the package.json. This implies the older npx version with Node.js 16 isn't respecting the local versions properly. So I've just reverted to not using npx instead.

guybedford:
Actually it seems the `--project` argument is causing `@types/node` not to be picked up, which may be an argument for moving the package.json to the base level. I've worked around that issue for now.

PR #463, opened 2023-01-25
Remove the JS host generator from this repository

alexcrichton:
This PR finishes the final remaining bits of https://github.com/bytecodealliance/wit-bindgen/issues/395 by removing the JS host generator from this repository. The generator now lives at https://github.com/bytecodealliance/js-component-tools with an NPM package and everything.

The major remaining question for this repository was how testing was going to work. My original intention was to depend on the JS host to run tests here, as is done today, but after some discussion on https://github.com/bytecodealliance/js-component-tools/pull/16 the js-component-tools repo is still going to depend on this repository for now for testing, so that wouldn't work. Instead I've brought back the Wasmtime version of all the tests and removed the JS host tests. This means that all tests are now run with Wasmtime's component runtime. This does mean that tests build more slowly and I've refactored the infrastructure to have less crates so it may be slower, but at least they're all still running!


alexcrichton:
Ah and I should also mention, I've removed the demo from this repository as well. While it's a neat demo I don't think it makes sense to keep around any more with so many generators spun out. I think each repo can have its own version of a demo if necessary.

One thing I do plan on doing shortly after htis though is to update the README and add prebuilt binaries to CI.

guybedford:
Instead of entirely removing the demo, which is incredibly useful for those new to the project to see how things fit together, perhaps we can move the demo to its own dedicated repo at this point, that brings together the various workflows as much as possible?

alexcrichton:
That seems reasonable to me, but to break the dependence of this repo on the JS generator it must be moved out of this repository.

pchickey (approved):

PR #465, opened 2023-01-26
Fixes for `cabi_realloc` when new_size = 0

### LINKED ISSUE #462: gen-guest-c: `cabi_realloc` doesn't tollerate `realloc(..., 0) == NULL`
> realloc(3p)
> ...
> RETURN VALUE
> ... If size is 0, either:
> *  A null pointer
> *  A pointer to the allocated space shall be returned

While dlmalloc does the latter (returning a unique pointer), some other allocators - including emmalloc as a compile-time option to wasi-libc - do the former (returning NULL).

`gen-guest-c` currently generates this wrapper:

```c
__attribute__((weak, export_name(\"cabi_realloc\")))
void *cabi_realloc(void *ptr, size_t orig_size, size_t org_align, size_t new_size) {
    void *ret = realloc(ptr, new_size);
    if (!ret) abort();
    return ret;
}
```

This will always trap for these other allocators on `new_size = 0`.

### LINKED ISSUE #464: guest-rust: UB in `cabi_realloc`
In `cabi_realloc`:

https://github.com/bytecodealliance/wit-bindgen/blob/98c2b1e4cb10d3757114b7524d112511a4c7459e/crates/guest-rust/src/lib.rs#L28-L37

If `old_len > 0` and `new_len == 0`, you'll end up calling `alloc::realloc(..., 0)`.

> https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#method.realloc
>
> ... undefined behavior can result if the caller does not ensure all of the following:
> ...
> - new_size must be greater than zero.

lann:
* gen-guest-c: Fix cabi_realloc behavior when new_size is zero (Fixes #462)
* guest-rust: Add an assert to cabi_realloc (Fixes #464)

### DIFF
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -32,6 +32,7 @@ pub mod rt {
             layout = Layout::from_size_align_unchecked(new_len, align);
             alloc::alloc(layout)
         } else {
+            debug_assert_ne!(new_len, 0, "non-zero old_len requires non-zero new_len!");
             layout = Layout::from_size_align_unchecked(old_len, align);
             alloc::realloc(old_ptr, layout, new_len)
         };
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -418,14 +418,15 @@ impl C {
         // Note that these intrinsics are declared as `weak` so they can be
         // overridden from some other symbol.
         self.src.c_fns(
-            "
-                __attribute__((weak, export_name(\"cabi_realloc\")))
-                void *cabi_realloc(void *ptr, size_t orig_size, size_t org_align, size_t new_size) {
+            r#"
+                __attribute__((weak, export_name("cabi_realloc")))
+                void *cabi_realloc(void *ptr, size_t old_size, size_t align, size_t new_size) {
+                    if (new_size == 0) return (void*) align;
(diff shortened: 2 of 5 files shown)


crates/guest-rust/src/lib.rs
```diff
@@ -32,6 +32,7 @@ pub mod rt {
             layout = Layout::from_size_align_unchecked(new_len, align);
             alloc::alloc(layout)
         } else {
+            assert_ne!(new_len, 0, "non-zero old_len requires non-zero new_len!");
```
lann, line 35: It isn't obvious to me whether this should be `debug_assert_ne`.

alexcrichton (approved):
Looks good to me, could a test be added for the 0-sized allocation as well?

crates/guest-rust/src/lib.rs
alexcrichton: I thought that this was ok, but actually on second thought now that I'm playing around with some modules locally, I think it would be best to keep asserts out of this function. So yeah can you change this to a `debug_assert!`?

lann:
> Looks good to me, could a test be added for the 0-sized allocation as well?

Sure. I haven't worked with these runtime tests yet so it might take a while.

lann:
One COVID later, I've added a test to exercise cabi_realloc with new_size = 0

alexcrichton (approved):

PR #470, opened 2023-01-29
Fix rust generator guest  import error in `no_std` build

mambisi:
error snapshot
```
    |
100 |     pub fn generate_native_address(seed: &[u8]) -> Vec<u8> {
    |                                                    ^^^ not found in this scope
    |
help: consider importing one of these items
    |
3   |     use alloc::vec::Vec;
    |
3   |     use crate::Vec;
    |
3   |     use rune_std::vec::Vec;
    |
3   |     use wit_bindgen_guest_rust::rt::vec::Vec;
    |
```
adds
```
#[allow(unused_imports)]
use wit_bindgen_guest_rust::rt::{{alloc, vec::Vec, string::String}};
```
in the beginning of the mod generation, remove it from ending

### DIFF
--- a/crates/gen-guest-rust/src/lib.rs
+++ b/crates/gen-guest-rust/src/lib.rs
@@ -368,12 +368,6 @@ impl InterfaceGenerator<'_> {
         self.src.push_str("#[allow(clippy::all)]\n");
         let params = self.print_signature(func, param_mode, &sig);
         self.src.push_str("{\n");
-        self.src.push_str(
-            "
-                #[allow(unused_imports)]
-                use wit_bindgen_guest_rust::rt::{{alloc, vec::Vec, string::String}};
-            ",
-        );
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I think this breaks some existing tests, however, so that'll need fixing as well.

Can you additionally add tests for this to the `tests/codegen.rs` file?

alexcrichton:
Sorry but for fixing the tests I don't mean adding imports to the test. The generated code shouldn't require the code around the macro to import anything as the generated code should be entirely standalone.

mambisi:
Got it

sunfishcode:
Thanks for the PR! This ended up getting fixed in #488. For `no_std` use cases, wit-bindgen now has a `--std-feature` flag in which it emits code that uses `#[feature = "std"]` guarding things that need std.

PR #321, opened 2022-09-22
feat(go): add a TinyGo guest generator (WIP, do not merge!)

Mossaka:
I started this work as a hackathon project. It currently can only work with integer functions for both import/export. It's not nearly half done yet, but I want to put this out to those who are interested in. Any feedback on the go user application's experience would be much welcomed!

This is also a place to track things to be done:
- [x] type numbers
- [x] update to use the world generator trait
- [x] type record
- [x] type flags
- [x] type tuple
- [x] type string
- [ ] type varient
- [ ] type option
- [ ] type union
- [ ] type enum
- [ ] type alias
- [x] type list
- [ ] type builtin
- [ ] "go:generate" and executable that hides "init" function
- [ ] ~type resource~

Signed-off-by: Jiaxiao Zhou <jiazho@microsoft.com>

PR #471, opened 2023-01-31
feat: add guest binding generator for TinyGo/Go

Mossaka:
This PR contains changes related to the implementation of a `wit-bindgen-gen-guest-go` crate, which based on the [TinyGo](https://tinygo.org/)/Go compiler. The `codegen.rs` was added and all the "tests/codegen" tests have been passed, except a few that include floating numbers. The crate takes a dependency on the C bindgen crate. 

## Why choosing TinyGo?

Fermyon has an execllent [guide](https://www.fermyon.com/wasm-languages/webassembly-language-support) that tracks a list of programming languages that compile to wasm. [The Go in WebAssembly](https://www.fermyon.com/wasm-languages/go-lang) mentions currently only TinyGo compiler supports `wasm32-wasi` as a build target. 

I expect to rework on this crate to switch to the mainline Go compiler when the upstream Go supports wasm/wasi.

Go's [proposal](https://github.com/golang/go/issues/58141) to add wasi target 

## Why depending on C APIs instead of Wasm ABI?

I chose to implement the Go bindings for the C APIs rather than the Wasm ABIs without a compelling reason. This project started as a one-week hackathon project in September 2022, and I wanted to choose something simple to get started. The CGo and the generated C bindings seemed like a good option as they covered lower-level pointer arithmetic. At the end of the week, I had a codegen that can generate numbers for Go #321. However, I later realized that the implementation of converting Go types to C types and vice versa was not as straightforward as I had initially thought. The optimizations that made the C APIs more idiomatic to C users made the type conversions more challenging.

If there is a strong need to implement the Go bindings using Wasm ABIs in the future, I am open to refactoring the code to accommodate this change.

## What is the issue with floating numbers?

It is a known issue in TinyGo and has been fixed in the dev branch. See here for [details](https://github.com/tinygo-org/tinygo/issues/3193). I am expecting the next release (`0.27`) will fix this issue.  

The CI failures are all due to this issue. https://github.com/bytecodealliance/wit-bindgen/actions/runs/4051619869/jobs/6970134812

TinyGo has released [v0.27.0](https://github.com/tinygo-org/tinygo/releases/tag/v0.27.0) which fixed the floating numbers issue. If you are using the latest TinyGo version then you should not run into the issue described above. 

## What's next?

- [x] I have not added any guest Go applications in the `tests/runtime` folder. This will be the next step. 
- [ ] I'd also love to work on improving go bindings ergonomics. 
- [ ] Use `Ns` to create local variables to improve readable on the genereated Go code.


alexcrichton:
Thanks for this!

Reading over this, I notice that this isn't using the `Bindgen` trait from the `wit-parser` crate, which makes sense from using the C bindings rather than dealing with ABI details. One major consequence, however, is that it doesn't appear that memory management is baked in here? I see calls to `C.CString`, for example, without paired calls to `C.free`. Similarly memory allocated by lowering into memory doesn't appear to be deallocated anywhere. Given the `-gc=leaking` flag to TinyGo, however, is that expected that all memory is leaked?

One other thing I noticed is that lists don't have what appears to be much special treatment, instead just creating a slice. I don't think that this would work for aggregates like `list<string>` or `list<list<u8>>` or similar?

Mossaka:
@alexcrichton thanks for your comment!

> is that it doesn't appear that memory management is baked in here?

That's right. I forgot to invoke `free` after allocating memory in C. Will work on this.

> I don't think that this would work for aggregates like list<string> or list<list<u8>> or similar?

Good point!

Mossaka:
@alexcrichton I am not usre if my memory management is solid, but I have covered the cases where function arguments and variables are allocated as C types. Next step -> I will fix nested lists and probably nested `Result`s too.

alexcrichton:
Hm well so honestly I don't know whether this is right or not. There's a ton of special-casing already and looks like you've updated quite a few locations to call free, but I can't really check to verify that they're all correct. Note that this is one of the main purposes of the `Bindgen` trait from `wit-parser` as it moves most of the complexity to one or two locations rather than having to call free in lots of places.

Overall I would recommend getting `runtime` tests working, especially the `lists` one. Even the runtime tests, though, are unlikely sufficient as they're not very broad in their coverage. Additionally there likely aren't that many protections against if something goes wrong. For example if a use-after-free happens it may just-so-happen to work. Similarly if something doesn't get deallocated nothing currently fails. This is handled in Rust with a custom allocator and such, but I'm not sure that the equivalent can be done for TinyGo.

Note that there's also special casing for whether something is lifted or lowered into an import. For example if you get a string from outside the wasm module then that's an owned memory allocation that needs to be free'd. If you return a string from an exported function then it must be an owned allocation that `post_return` deallocates. Finally if you pass a string to an imported function it probably doesn't need any allocation at all since the raw pointer/length can be passed along to the import call. I'm not sure whether this is all handled correctly in the implementation.

Mossaka:
I have successfully updated the implementation for complex type lists, making it much more robust. I have run the lists runtime tests, and all have passed except those using `float32` and `float64` types. This issue is a [known issue](https://github.com/tinygo-org/tinygo/pull/3210) in TinyGo, but has already been resolved and will be fixed in the next release.

Next, I will focus on the `Result` type, which is currently being implemented in a similar manner to the previous implementation of lists. I anticipate that results may not be functioning properly with complex types.

Moving forward, my plan is to enable all runtime tests incrementally and fix any bugs that arise. I am confident in my progress thus far, as the list implementation is now functioning properly.

Mossaka:
I have updated the PR to include some runtime tests. Here is a list of all runtime tests currently sitting in the main branch. Checked ones include a Go guest implementation and are passing. 

- [x] flavorful/
- [x] lists/
- [x] many_arguments/
- [x] numbers/
- [x] records/
- [ ] results/
- [x] smoke/
- [ ] strings/
- [ ] unions/
- [x] variants/

alexcrichton (approved):
Thanks again for this @Mossaka! Was there anything else you wanted to add or is this good to go?

Mossaka:
> Thanks again for this @Mossaka! Was there anything else you wanted to add or is this good to go?

I'd prefer to run the runtime tests a few more time. I've observed some flakiness in the tests.

alexcrichton:
Thanks again for this @Mossaka! I'm going to go ahead and merge this and it can continue to be iterated on in-tree.

Issue #472, opened 2023-01-31
gen-guest-c: failed on option of empty record


Mossaka:
The C generator will fail for the following wit file:

```fsharp
interface option-result {
	record empty {}
        type o2 = option<empty>
	o2-arg: func(x: o2)
	o2-result: func() -> o2
}

default world my-world {
	import imports: self.option-result
	export exports: self.option-result
}
```

The error says that 
```
/Users/mossaka/Developer/bca/wit-bindgen/target/codegen-tests/guest-c-option-result/my_world/my_world.c:44:17: error: no member named 'val' in 'imports_o2_t'
          *ret = option.val;
                 ~~~~~~ ^
        /Users/mossaka/Developer/bca/wit-bindgen/target/codegen-tests/guest-c-option-result/my_world/my_world.c:71:8: error: no member named 'val' in 'exports_o2_t'
          ret0.val = val;
          ~~~~ ^
        2 errors generated.
        ', crates/test-helpers/src/lib.rs:45:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I expect the cause of the issue is in this line https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/gen-guest-c/src/lib.rs#L2374, where it forgets to check whether the inner type of `OptionBool` is empty type.

PR #473, opened 2023-01-31
fix: added empty checks to option and result types

Mossaka:
It appears that [this issue](https://github.com/bytecodealliance/wit-bindgen/issues/472) was caused by the missing of empty-type checks for inner types of `Option` and `Result` in various places. This PR adds the necessary checks to the Return functions and a more comprehensive `option-result.wit` file for testing nested options and more complicated compositions of results.

One thing before merging in this PR is that I have commented out "self.check_all_retptrs_written();". This is because when the empty-check is enforced, the retptrs len is no longer matching the nunber of `self.store_in_retptr()` being invoked. I am not sure if this is the right thing to do, but so far, I've been able to pass the tests. @alexcrichton @guybedford 


Mossaka:
@dicej it looks like Java CI is failing the tests after the new wit file I added in this PR. If you have time, would you like to take a look at it? It looks like there is a type conversion from `Tuple0` to `Empty` which is incompatible. This means there is a mismatch between the expected data type and the actual data type being assigned.

dicej:
> @dicej it looks like Java CI is failing the tests after the new wit file I added in this PR. If you have time, would you like to take a look at it? It looks like there is a type conversion from `Tuple0` to `Empty` which is incompatible. This means there is a mismatch between the expected data type and the actual data type being assigned.

Thanks for letting me know.  I'll definitely take a look at it, but it might not happen today.  Feel free to disable the offending Java test for the time being.

alexcrichton (approved):
Thanks!

PR #484, opened 2023-02-06
Implement record, variant, etc. for the markdown backend.

sunfishcode:
Implement basic markdown printing for record, variant, union, enum, and flags types.


alexcrichton:
I believe the intention of the type printing at this location was not to render the full structure of the type but instead print named items, and the various previously-`unreachable!()` cases should all be required to have names. I think it might be better if these had their names instead of their structures printed? I don't really remember what the `skip_names` field and various other structures going on here are though

sunfishcode:
`skip_names` turned out to only be used in one place, and I think the output looks ok without skipping the name, so I've now removed `skip_names` and made this code just assert that records etc. are named.

alexcrichton (approved):

PR #486, opened 2023-02-06
Add a `--check` option to wit-bindgen.

sunfishcode:
Add a `--check` option to wit-bindgen for testing whether generated files are already up to date. With this flag, if the generated files differ from what would be generated, the wit-bindgen command exits with an error.

### DIFF
--- a/src/bin/wit-bindgen.rs
+++ b/src/bin/wit-bindgen.rs
@@ -94,9 +94,8 @@ fn main() -> Result<()> {
         println!("Generating {:?}", dst);
 
         if opt.check {
-            let prev = std::fs::read_to_string(&dst)
-                .with_context(|| format!("failed to read {:?}", dst))?;
-            if prev.as_bytes() != contents {
+            let prev = std::fs::read(&dst).with_context(|| format!("failed to read {:?}", dst))?;
+            if prev != contents {
                 bail!("not up to date: {}", dst.display());
             }
             continue;
--- a/src/bin/wit-bindgen.rs
+++ b/src/bin/wit-bindgen.rs
@@ -64,6 +64,11 @@ struct Common {
     /// it's `foo.bar` which is the world named `bar` within document `foo`.
     #[clap(short, long)]
     world: Option<String>,
+
+    /// Indicates that no files are written and instead files are checked if
+    /// they're up-to-date with the source files.
+    #[clap(long)]
+    check: bool,
 }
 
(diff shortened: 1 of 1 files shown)


pchickey (approved):

Mossaka:
This is great to have! I wonder if it makes sense to embed codegen metadata into the generated file. Something like the output of of `wit-bindgen --version` being embedded to the top of the generated files as docs.

src/bin/wit-bindgen.rs
```diff
@@ -87,6 +92,16 @@ fn main() -> Result<()> {
             None => name.into(),
         };
         println!("Generating {:?}", dst);
+
+        if opt.check {
+            let prev = std::fs::read_to_string(&dst)
```
alexcrichton, line 97: I think this'll want to be `std::fs::read`, minus the `_to_string`, to handle some of the binary outputs of some generators,

sunfishcode: Fixed!

alexcrichton:
Hm actually reading over this again, one thing that this'll probably have to handle is crlf stuff. For example if you check in a file to git and then check it out on Windows the line endings probably all changed, which means that the equality check here should probably ignore line endings.

sunfishcode:
Wit-bindgen also does generate binary files. Rather than complicate this tool with knowledge of which outputs are text or binary, could we just say that we'll always set git metadata to check out files with NL-style line endings even on Windows? All Windows tools these days can deal with that.

alexcrichton:
It's true, yeah, but could this perhaps use `str::from_utf8` and do an exact equality check on failure, and a line-equality check otherwise? If possible it seems best to handle this in one place instead of having all users update `.gitattributes` slowly as they hit this issue.

sunfishcode:
I'm not comfortable with "if it validates as UTF-8 then ignore CRLF differences". We could also check for unexpected control codes, but it still feels like it'd be leaving an obscure trap door in the code.

If we take the .gitattributes approach, most wit-bindgen users wouldn't be affected. It'd only be the people writing `wit-bindgen --check` rules to run in CI. I've also now added logic to print a targeted error message when the contents appear text and differ only in line endings; what do you think?

alexcrichton:
I don't think it would be too hard to update generators to have a flag when something is pushed in `Files` for "is this text" which would avoid this sort of detection logic. That being said this isn't worth belaguering too much.

PR #488, opened 2023-02-07
Teach guest Rust how to qualify `Vec` and `String` for `no_std`.

sunfishcode:
The Rust generator emits `use` declarations for `Vec` and `String` inside function bodies which usually works, but doesn't work when When `Vec` and `String` appear as return types. This teaches the Rust guest generator how to fully qualify `Vec` and `String` so that they always work, even in `no_std` mode when they aren't in scope by default.


pchickey (approved):

alexcrichton:
Could some tests be added for this as well?

sunfishcode:
Tests added!

Issue #288, opened 2022-07-21
suggestion: add DO NOT EDIT comment to generated bindings files


ricochet:
Generated bindings files should indicate that they are generated with an auto-generated DO NOT EDIT comment at top of the file. This is not needed for Rust, but any language like c where the bindings are part of the project should have this comment. 

```c
// Code generated by wit-bindgen DO NOT EDIT.
// Version: 
// Revision: 
// Build Date: 
// Built By: 
``` 

Many editors recognize this string and warn when editing the file. Many tools include build-level information for triage. This might be especially useful for wit-bindgen while breaking changes are still frequent.

sunfishcode:
:+1:  to adding comments to the generated bindings.

I'm curious about the need ffor Version/Revision/Build Date/Built By fields. At first glance, it seems like version-control systems do a good job of tracking this information already.

ricochet:
I included this snippet from what I've seen when the generator is a CLI and not managed as a dependency, e.g. not in go.mod/go.sum, Cargo.toml, etc.

go-enum is an example: https://github.com/abice/go-enum/blob/8268238fd705b39d430e7d8682b527af8452b6ab/generator/enum.tmpl#L1-L6

So if wit-bindgen were run as a CLI because it was provided as a globally installed tool in a build container, the version used might not be captured in git.

To be honest, most of those fields seemed redundant to me, but I figured I'd drop the superset of what I've seen and we can pick from that.

philpax:
This has been fixed by #490; it appears GitHub didn't auto-close this issue.

PR #490, opened 2023-02-09
add bindgen version to file header

### LINKED ISSUE #487: Embed the version of wit-bindgen in generated files
Following up on @Mossaka's idea in #486, we should add a comment to generated source files indicating the wit-bindgen version, for debugging purposes.

Mossaka:
Generated source files now include a header comment indicating that the file is generated by a specific version of `wit-bindgen`. This will close #487 and #288


pchickey (approved):

crates/bindgen-core/src/lib.rs
```diff
@@ -8,6 +8,12 @@ mod ns;
 
 pub use ns::Ns;
 
+/// Helper for getting the version of the bindgen.
+/// If CARGO_VERSION_INFO is set, use it, otherwise use CARGO_PKG_VERSION.
+pub fn version() -> &'static str {
+    option_env!("CARGO_VERSION_INFO").unwrap_or(env!("CARGO_PKG_VERSION"))
```
alexcrichton, line 14: If this is moved here the `build.rs` is going to need to move in tandem since that's what defines the `CARGO_VERSION_INFO` env var

alexcrichton (approved):

crates/bindgen-core/src/lib.rs
Mossaka: That's a really good point!

alexcrichton (commented):
Can you verify that `cargo run -- --version` still shows the git sha? I think that the build script's detection of `.git` may need to be updated

Mossaka:
> Can you verify that `cargo run -- --version` still shows the git sha? I think that the build script's detection of `.git` may need to be updated

good catch! I just updated the path to `.git`

crates/bindgen-core/build.rs
```diff
@@ -7,7 +7,7 @@ fn main() {
 }
 
 fn commit_info() {
-    if !Path::new(".git").exists() {
+    if !Path::new("../../.git").exists() {
```
alexcrichton, line 10: Sorry but I think that this check may not work out well. This won't play well when this is `cargo vendor`'d somewhere else, for example, since it may check for a `.git` directory that's not its own. Additionally as part of the registry this would be reaching outside of its registry location and looking for `.git`.

While this works in this git repository itself and it probably won't cause issues in other locations I'm wary to have this as it seems like an avenue to cause issues.

I don't know of a way to fix this so in retrospect I think it's fine to just print a version in the generated source code and leave off the date/commit/etc.

Mossaka: That sounds fine to me. Do you still want to print date/commit info in the CLI though?

alexcrichton: For the CLI, yeah.

Although as I think on this more in relation to https://github.com/bytecodealliance/wit-bindgen/pull/498 instead of pulling a version from `bindgen-core` could each generator use its own version? I think we'll get to a future soon where each generator has its own independent track of versions.

Mossaka: That's a good point! I will revert it back and use crate's version instead of pulling from bindgen-core.

Mossaka: Okay now I reverted back to use "CARGO_PKG_VERSION" in the file header. This does not include git hash and date, but at least gives the version information.

alexcrichton (approved):

PR #492, opened 2023-02-11
fix Windows CI

dicej:
The same Java and Maven versions that work on Linux and Mac have suddenly starting failing on Windows, so we now manually install a known good version for that platform.


dicej:
Whelp, nevermind.  It worked on my branch, but apparently there's some kind of non-determinism happening.  We might need to just disable the Java tests on Windows if I can't get to the bottom of this.

PR #498, opened 2023-02-13
Prepare CI/automation for releasing

alexcrichton:
This commit prepares this repository for starting the publication process to crates.io and binary releases. Now that we're intending to enter an era of more stability around WIT and the component model binary format it seems prudent to have releases of this repository to assist others with integrating the tooling into their workflows. At this time this includes:

* Tags will be published when necessary to represent versions of `wit-bindgen`. All crates are managed separately with different versions and publication of one crate doesn't necessarily imply one of the other. All publishes will update the CLI version though.
* A script has been added to make managing versions a bit easier in this repository. (`ci/publish.rs`)
* Tags will have a GitHub release made with prebuilt binaries of the `wit-bindgen` binary. 
* The Rust guest embedding is now named `wit-bindgen` as a crate name instead of `wit-bindgen-gen-guest-rust`.

Once this is sorted out I plan on doing a trial run with an 0.3.0 release of all the fiddly bits here.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ permissions:
 
 jobs:
   build:
-    name: Build wasm-tools
+    name: Build wit-bindgen
     runs-on: ${{ matrix.os }}
     if: github.event_name != 'push' || startsWith(github.ref, 'refs/tags/wit-bindgen-cli-')
     strategy:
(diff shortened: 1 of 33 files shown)


pchickey:
I don't think we can cut a release until we no longer depend on a patch for the git version of wasmtime

alexcrichton:
I think that's only used for dev-dependencies (tests), though? I can try to switch that to the 6.0.0 release branch though and see if that's sufficient to use when it's released.

alexcrichton:
Ok I'm gonna attempt to turn these cranks and see what happens.

PR #505, opened 2023-02-14
Standardize on world selection throughout tooling

alexcrichton:
This updates to use `Resolve::select_world` which was added to `wit-parser` from #494.


pchickey (approved):
Love to see it

PR #509, opened 2023-02-16
Rust: do not print `-> ()` for signatures that have no results.

peterhuene:
This PR fixes the printing of an unnecessary `-> ()` for function signatures that have no results.

While we ignore clippy lints in the generated code, rust-analyzer will blindly copy the unnecessary return specification when auto-implementing missing trait members. This might cause clippy warnings in user code as a result.

### DIFF
--- a/crates/gen-rust-lib/src/lib.rs
+++ b/crates/gen-rust-lib/src/lib.rs
@@ -95,8 +95,7 @@ pub trait RustGenerator<'a> {
         sig: &FnSig,
     ) -> Vec<String> {
         let params = self.print_docs_and_params(func, param_mode, &sig);
-        self.push_str(" -> ");
-        self.print_result_params(&func.results, TypeMode::Owned);
+        self.print_results(&func.results, TypeMode::Owned);
         params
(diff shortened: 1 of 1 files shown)


peterhuene:
I'm currently implementing a source generator for `cargo-component` that creates a `src/lib.rs` with the boiler plate impls for all exports of a target world automatically when creating a new component and specifying the target to use.

I just happen to notice this when I got as far as spitting out the impls of the bindings traits, sans member impls; I then asked to rust-analyzer to generate the member impls for me and out popped the unnecessary return specifiers.

One might consider that a rust-analyzer bug, but I think that the garbage-in / garbage-out principle applies here and we should just not generate these from wit-bindgen.

pchickey (approved):

pchickey:
Thanks! This very literal copying from rust-analyzer gets worse, unfortunately, when the generated traits are async, because it will copy out the expansion of the #[async_trait] macro. That will probably require some sort of fix upstream, but this is a fine thing for us to fix locally.

PR #514, opened 2023-02-17
Fix using Rust types in owned and borrowed contexts

### LINKED ISSUE #513: gen-guest-rust: cannot reuse shared records between import and export interfaces
The following wit file:

```
interface types {
    record a {
        b: string    
    }
}

interface foo {
    use self.types.{a}

    x: func(param: a)
}

interface bar {
    use self.types.{a}

    x: func(param: a)
    y: func() -> a
}

default world demo {
    import foo: self.foo
    export bar: self.bar
}
```

Fails to compile with the following error:

```rust
error[E0422]: cannot find struct, variant or union type `AParam` in this scope
 --> src/lib.rs:3:1
  |
3 | / wit_bindgen::generate!({
4 | |     world: "demo",
5 | | });
  | |__^ not found in this scope
  |
  = help: consider importing this struct:
          crate::types::AParam
error[E0412]: cannot find type `A` in module `super::types`
 --> src/lib.rs:3:1
  |
3 | / wit_bindgen::generate!({
4 | |     world: "demo",
5 | | });
  | |__^ not found in `super::types`
  |
  = help: consider importing this type alias:
          crate::foo::A
  ```
My findings:

The generated Rust type for `record a` gets split into two types and looks like this:

```rust
pub mod types {
    pub struct AParam<'a> {
        pub b: &'a str,
    }
    ...
    pub struct AResult {
        pub b: wit_bindgen::rt::string::String,
    }
    ...
}
```

However, The `bar` module attempts to import `record a` as `super::types::A;`:

```rust
pub mod bar {
    pub type A = super::types::A;
    ...
}
```

Conditions to encounter this error:

- one `import` interface uses a record (with a non-primitive field) as a function parameter
- an `export` interface uses the same record as either parameter or return type

Tested with:

- v0.3.0
- commit: 462527127f0d9de935f01e298a7e1c975e124a62

alexcrichton:
This pulls in type analysis code from Wasmtime which has a more modern notion of param/result as owned/borrowed and additionally removes the need for a `default_param_mode` which caused #513 in the first place.

Closes #513

### DIFF
--- a/crates/gen-guest-c/src/lib.rs
+++ b/crates/gen-guest-c/src/lib.rs
@@ -1115,21 +1115,7 @@ impl InterfaceGenerator<'_> {
                 let ty = &self.resolve.types[*id];
                 match &ty.name {
                     Some(name) => {
-                        match ty.owner {
-                            TypeOwner::Interface(owner) => {
-                                self.src.print(
-                                    stype,
-                                    &self.gen.interface_names[&owner].to_snake_case(),
-                                );
-                                self.src.print(stype, "_");
-                            }
-                            TypeOwner::World(owner) => {
-                                self.src
-                                    .print(stype, &self.resolve.worlds[owner].name.to_snake_case());
-                                self.src.print(stype, "_");
-                            }
-                            TypeOwner::None => {}
-                        }
+                        self.print_owner_namespace(stype, *id);
 
                         self.src.print(stype, &name.to_snake_case());
                         self.src.print(stype, "_t");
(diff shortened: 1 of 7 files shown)


Mossaka:
I have already told Joel about the macos CI failure on maven. Feel free to merge it!

crates/gen-guest-c/src/lib.rs
```diff
@@ -1457,6 +1462,9 @@ impl InterfaceGenerator<'_> {
                 self.src.h_helpers(&self.gen.world.to_snake_case());
                 self.src.h_helpers("_");
             }
+            Type::Id(id) => {
```
Mossaka, line 1465: will this change the behavior of option flattening?

Mossaka (approved):
lgtm!

crates/gen-guest-c/src/lib.rs
alexcrichton: Unfortunately no, this was just to fix some compile issues that cropped up with the new test case.

PR #517, opened 2023-02-21
Use a crates.io-based version of Wasmtime for tests

alexcrichton:
Update from the git dependency to a crates.io-based dependency.


pchickey (approved):
🚀

Issue #495, opened 2023-02-11
CI tests are flaky due to spurious Maven failures


dicej:
CI started failing for various PRs a few days ago, unrelated to any code changes.  I've been unable to reproduce it locally.

Using the latest OpenJDK and latest Maven releases initially seemed to fix it, but that seems to have been a coincidence since it still fails most of the time.

If I can't figure out how to make Maven happy, I'll switch the Java tests to download and run a TeaVM "fat" jar and run it that way, bypassing Maven.

dicej:
For reference, here is an example of the output of a failing test:

```
---- numbers::run stdout ----
wasi adapter = "D:\\a\\wit-bindgen\\wit-bindgen\\target\\debug\\build\\test-artifacts-17eef5ab52e6f25e\\out\\wasm32-unknown-unknown/release/wasi_snapshot_preview1.wasm"
rust core module = "D:\\a\\wit-bindgen\\wit-bindgen\\target\\debug\\build\\test-artifacts-17eef5ab52e6f25e\\out\\wasm32-wasi/debug\\numbers.wasm"
rust component "D:\\a\\wit-bindgen\\wit-bindgen\\target\\runtime-tests\\numbers\\rust.wasm"
"D:/a/wit-bindgen/wit-bindgen/wasi-sdk-16.0\\bin/clang" "--sysroot" "D:/a/wit-bindgen/wit-bindgen/wasi-sdk-16.0\\share/wasi-sysroot" "./tests/runtime\\numbers\\wasm.c" "D:\\a\\wit-bindgen\\wit-bindgen\\target\\runtime-tests\\numbers\\numbers.c" "D:\\a\\wit-bindgen\\wit-bindgen\\target\\runtime-tests\\numbers\\numbers_component_type.o" "-I" "D:\\a\\wit-bindgen\\wit-bindgen\\target\\runtime-tests\\numbers" "-Wall" "-Wextra" "-Werror" "-Wno-unused-parameter" "-mexec-model=reactor" "-g" "-o" "D:\\a\\wit-bindgen\\wit-bindgen\\target\\runtime-tests\\numbers\\c-wasm.wasm"
"cmd" "/c" "mvn" "prepare-package"
status: exit code: 1
stdout: ------------------------------------------
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------------------------< foo:foo >-------------------------------
[INFO] Building foo 1.0-SNAPSHOT
[INFO]   from pom.xml
[...]

PR #520, opened 2023-02-23
remove use of Maven in Java tests

dicej:
Per #495, Maven has proven quite unreliable for the past few weeks, so now we download and execute the relevant TeaVM jars directly, which should be more reliable.

This also fixes the lift-lower-foreign.wit codegen test for Java.


Mossaka (approved):
lgtm!

PR #527, opened 2023-03-04
feat(go): Add variants runtime test for go bindgen

### LINKED ISSUE #528: Go: Enable variants tests

Mossaka:
This will close #528


Mossaka:
It is weird that it works on my local macbook and passed the windows CI, but not Linux and MacOS runners. I will make this PR a draft one and continue to find the root cause.

PR #530, opened 2023-03-06
Fix bindings generation with unused types in Rust

alexcrichton:
This commit fixes an issue where if a type was imported into an interface but wasn't actually used anywhere then bindings weren't correctly generated because inference about whether it was owned or borrowed didn't run. The fix here is to skip generating bindings for unused types since they're not actually needed anywhere anyway. If necessary in the future bindings can be forcibly generated but for now this is hopefully largely "just" fixing an edge case.

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -166,7 +166,8 @@ impl Types {
             }
             TypeDefKind::Unknown => unreachable!(),
         }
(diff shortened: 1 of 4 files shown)


peterhuene (approved):

alexcrichton:
cc @Mossaka I'm adding an ignore for TinyGo for the new test I'm adding since it failed in CI above

PR #533, opened 2023-03-07
rust: export funcs explicitly invoke static constructors at most once

pchickey:
Before executing any other code, use this function (provided by wasm-ld)
to run all static constructors. When wasm-ld sees this explicit invocation
of ctors, it will not generate a wrapper aorund each exported func which
executes ctors first, and therefore we make sure that cabi_realloc doesnt
invoke ctors when called as an export func.
See https://github.com/bytecodealliance/preview2-prototyping/issues/99
for more details.

### DIFF
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -21,6 +21,7 @@ pub mod rt {
     /// permitted by the Component Model when inside realloc.
     ///
     /// We intend to remove this once rust 1.69.0 stabilizes.
+    #[cfg(target_arch = "wasm32")]
     pub fn run_ctors_once() {
         static mut RUN: bool = false;
         unsafe {
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -11,6 +11,35 @@ pub use bitflags;
 
 #[doc(hidden)]
 pub mod rt {
+
+    /// Provide a hook for generated export functions to run static
+    /// constructors at most once. wit-bindgen-rust generates a call to this
+    /// function at the start of all component export functions. Importantly,
+    /// it is not called as part of `cabi_realloc`, which is a *core* export
+    /// func, but may not execute ctors, because the environment ctor in
+    /// wasi-libc (before rust 1.69.0) calls an import func, which is not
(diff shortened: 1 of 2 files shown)


pchickey:
https://github.com/bytecodealliance/preview2-prototyping/pull/101 demonstrates that this is a fix that prevents running constructors which call import functions from cabi_realloc

alexcrichton (approved):

alexcrichton:
Hm actually is this correct in the face of multiple invocations? IIRC wasi-libc has no guards against repeatedly invoking this so this may run the risk of allocating the whole environment on each export call?

Should this instead be moved to `crates/guest-rust/src/lib.rs` and have a static bool guarding it so it's only invoked once?

sunfishcode:
This is calling `__wasm_call_ctors` which is synthesized by wasm-ld, and it does not have guards against being called twice.

pchickey:
Right, the present situation is that the ctors get called at the start of every export func, and this PR backs them that to all export funcs except cabi_realloc. Right now the ctors that exist in wasi-libc are tolerant of that behavior.

In the Rust bindgen, we can add a once cell to the guest-rust crate and make sure ctors get invoked only once.

In the C bindgen, its much more difficult to add that logic, since there is no global singleton (like a crate static can provide) to make sure constructors only get invoked once. So, I'm actually considering dropping this patch from the C bindgen, rather than provide a fix that would be somehow incorrect, and leaving it as a problem for tool conventions and the canonical abi to fix.

pchickey:
Constructors run at most once, I rebased the C code generator changes out, and I added a note that this is just a temporary fix until rust 1.69.0 stabilizes, at which point wasi-libc ctors should no longer call import funcs (right?)

alexcrichton (approved):

alexcrichton:
Oh the Windows failure reminds me, could the contents of the whole fuction added here be wrapped in `#[cfg(target_arch = "wasm32")]`? Technically it's only safe on single-threaded platforms like wasm anyway and shouldn't do anything anywhere else otherwise.

pchickey:
Yes the target arch guard got dropped, by accident, in my refactor. I added it back in as a guard around the presence of the rt::run_ctors_once function.

Issue #524, opened 2023-02-27
C Optional Flattening in Foreign


Mossaka:
For the following interfaces 

```Go
interface a {
	record t2 { a: u32 }
	type t7 = option<t2>
}
  
interface the-interface {
	use self.a.{ t7 as u7, t2 as u2 }
	f2: func(a: u2) -> u2
	f7: func(a: u7) -> u7
}
  
default world foo {
	use self.a.{ t7 as u7 }

	export f7: func(a: u7) -> u7

	import the-import: self.the-interface
	export the-export: self.the-interface
}

```

The generated C header file looks like this

```C
typedef struct {
  uint32_t a;
} a_t2_t;

typedef struct {
  bool is_some;
  a_t2_t val;
} a_t7_t;

typedef a_t7_t the_import_u7_t;

typedef a_t2_t the_import_u2_t;

typedef a_t7_t foo_u7_t;

typedef a_t7_t the_export_u7_t;

typedef a_t2_t the_export_u2_t;

// Imported Functions from `the-import`
void the_import_f2(the_import_u2_t *a, the_import_u2_t *ret);
bool the_import_f7(the_import_u7_t *a, a_t2_t *ret);
```

It looks like the optional flattening will use the renamed type `the_import_u2_t` but instead uses the original type name `a_t2_t`. I am not sure what's the right behaviour here. Would love to hear your ideas. @alexcrichton @guybedford

alexcrichton:
This isn't necessarily intentional and is fine to change. Option stuff came way before interfaces and use.

As a side note, one idea I just had is to add an option to the C generator to disable all flattening like this and generate a more "raw" interface. It might be worth exploring what that would look like maintainability wise

guybedford:
I tend to agree having a first-class direct interface would be a good idea, for this generator especially.

Another way to look at the distinction between raw and usable bindgen might be defining usable bindgen in terms of wrappings of raw bindgen. Then raw bindgen with direct low-level options as the base layer. At least that was my hope for the JS generation as well.

Mossaka:
I don't know if this much effort of making a first-class direct interface for C is worth it considering the Go bindgen depending on C bindgen is a temperory solutation that eventually will be removed in favor of building a Wasm ABI bindgen for Go. But I can see value in it especially if we want to have more generators based on the C bindgen.

Mossaka:
Close this one as #534 removes flattening on Go bindgen

PR #534, opened 2023-03-07
feat: no_sig_flattening option

guybedford:
Was exploring https://github.com/bytecodealliance/wit-bindgen/issues/524 today, and was considering if we should disable all flattening by default as well.

This PR implements a `no_sig_flattening` option for the C generation as suggested.

I did explore entirely removing the flattening based on the argument that one might want a more "low level" interface, but at the end of the day even the C interface is a high level interface! And if anything, the flattening is actually slightly more optimal from a bindings perspective (while trivial, it does avoid unnecessary allocations of stack objects in lifted interfaces).

So I think having this option may well just be the best approach here.

This is just a draft initial exploration of the work, not ready to merge.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -784,23 +784,25 @@ impl InterfaceGenerator<'_> {
         // construct optional adapters from maybe pointers to real optional
         // structs internally
         let mut optional_adapters = String::from("");
-        for (i, (_, param)) in c_sig.params.iter().enumerate() {
-            let ty = &func.params[i].1;
-            if let Type::Id(id) = ty {
-                if let TypeDefKind::Option(option_ty) = &self.resolve.types[*id].kind {
-                    let ty = self.type_string(ty);
-                    uwrite!(
-                        optional_adapters,
-                        "{ty} {param};
-                        {param}.is_some = maybe_{param} != NULL;"
-                    );
-                    if !is_empty_type(self.resolve, option_ty) {
-                        uwriteln!(
+        if !self.gen.opts.no_sig_flattening {
+            for (i, (_, param)) in c_sig.params.iter().enumerate() {
+                let ty = &func.params[i].1;
+                if let Type::Id(id) = ty {
+                    if let TypeDefKind::Option(option_ty) = &self.resolve.types[*id].kind {
+                        let ty = self.type_string(ty);
+                        uwrite!(
                             optional_adapters,
-                            "if (maybe_{param}) {{
-                                {param}.val = *maybe_{param};
-                            }}",
+                            "{ty} {param};
+                            {param}.is_some = maybe_{param} != NULL;"
                         );
+                        if !is_empty_type(self.resolve, option_ty) {
+                            uwriteln!(
+                                optional_adapters,
+                                "if (maybe_{param}) {{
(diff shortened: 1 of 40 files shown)


crates/c/src/lib.rs
```diff
@@ -35,6 +35,9 @@ pub struct Opts {
     /// Set component string encoding
     #[cfg_attr(feature = "clap", arg(long, default_value_t = StringEncoding::default()))]
     pub string_encoding: StringEncoding,
+    // Skip optional null pointer and boolean result argument signature flattening
+    #[cfg_attr(feature = "clap", arg(long, default_value_t = false))]
+    pub no_sig_flattening: bool,
```
Mossaka, line 40: Would it be more precise and descriptive to name this flag as `no_option_ty_flattening`?

crates/c/src/lib.rs
```diff
@@ -466,22 +475,26 @@ impl Return {
...
 
             // Unpack a result as a boolean return type, with two
             // return pointers for ok and err values
             TypeDefKind::Result(r) => {
-                if let Some(ok) = r.ok {
-                    self.retptrs.push(ok);
-                }
-                if let Some(err) = r.err {
-                    self.retptrs.push(err);
+                if sig_flattening {
```
Mossaka, line 488: Nice!

Mossaka:
For `no_sig_flattening = true`, can we remove all the option flattening for both arguments and results?

I am still seeing the optional arguments being flattened even if the flag is set to true. 

```wit
interface option-result {
	type o1 = option<u32>
    o1-arg: func(x: o1)
    o1-result: func() -> o1

    record empty {}
    type o2 = option<empty>
    o2-arg: func(x: o2)
    o2-result: func() -> o2
}

default world my-world {
	import imports: self.option-result
	export exports: self.option-result
}
  
```
```C
// Imported Functions from `imports`
void imports_o1_arg(uint32_t *maybe_x);
void imports_o1_result(imports_o1_t *ret);
void imports_o2_arg(imports_empty_t *maybe_x);
void imports_o2_result(imports_o2_t *ret);

// Exported Functions from `exports`
void exports_o1_arg(uint32_t *maybe_x);
void exports_o1_result(exports_o1_t *ret);
void exports_o2_arg(exports_empty_t *maybe_x);
void exports_o2_result(exports_o2_t *ret);
```

## Expected

```C
```

Mossaka:
I pushed a commit to fix the above issue.

Mossaka:
All the runtime and codegen tests for go bindgen are passing for me.

alexcrichton:
@Mossaka would you be up for rebasing this and handling my comment above?

Mossaka:
> @Mossaka would you be up for rebasing this and handling my comment above?

Yes, WIP I am trying to enable all the codegen tests for go bindgen. 

What comment were you referring to? I didn't see you have any comment above.

Mossaka:
I've enable all the codegen go tests except issue 544 and issue 551. I intend to follow up with another PR to resolve these two tests. For now, I dis-abled them in the codegen tests, and thus this PR is ready to review!

guybedford (commented):
LGTM!

pchickey (approved):

PR #542, opened 2023-03-20
wit-bindgen-rust: link component type for imports.

peterhuene:
This PR forces a reference to the component type in import sub-modules.

It fixes an issue when building bindings to a different crate and only referencing an import binding.

As there was no reference to `__link_section`, the generated core module from the LLVM toolchain was missing the component type custom section.

The fix simply performs what we're already doing for exports in the `export!` macro, but adds the reference to each generated import submodule.

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -375,7 +375,7 @@ impl InterfaceGenerator<'_> {
                     #[used]
                     #[doc(hidden)]
                     #[cfg(target_arch = \"wasm32\")]
-                    static __FORCE_SECTION_REF: fn() = crate::__link_section;
+                    static __FORCE_SECTION_REF: fn() = super::__link_section;
 
                     {module}
                 }}
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -230,12 +230,7 @@ impl WorldGenerator for RustWasm {
                         #[used]
                         #[doc(hidden)]
                         #[cfg(target_arch = \"wasm32\")]
-                        static __FORCE_SECTION_REF: fn() = __force_section_ref;
-                        #[doc(hidden)]
-                        #[cfg(target_arch = \"wasm32\")]
-                        fn __force_section_ref() {{
-                            {prefix}__link_section()
-                        }}
+                        static __FORCE_SECTION_REF: fn() = {prefix}__link_section;
                     }});
                 ",
                 prefix = self.opts.macro_call_prefix.as_deref().unwrap_or("")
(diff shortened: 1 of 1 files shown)


peterhuene:
Unfortunately I don't know of a good way to test this fix that is local to the wit-bindgen, as we manually encode the component type information custom section and don't rely on wasm-ld to do so (afaik).

crates/rust/src/lib.rs
```diff
@@ -377,9 +377,19 @@ impl InterfaceGenerator<'_> {
             "
                 #[allow(clippy::all)]
                 pub mod {snake} {{
+                    #[used]
+                    #[doc(hidden)]
+                    #[cfg(target_arch = \"wasm32\")]
+                    static __FORCE_SECTION_REF: fn() = __force_section_ref;
```
peterhuene, line 383: Actually, I think we can simplify this to simply `crate::__force_section_ref`.  I'll fix.

crates/rust/src/lib.rs
```diff
@@ -230,12 +230,7 @@ impl WorldGenerator for RustWasm {
                         #[used]
                         #[doc(hidden)]
                         #[cfg(target_arch = \"wasm32\")]
-                        static __FORCE_SECTION_REF: fn() = __force_section_ref;
-                        #[doc(hidden)]
-                        #[cfg(target_arch = \"wasm32\")]
-                        fn __force_section_ref() {{
```
peterhuene, line 236: I've tested the removal of this symbol and it seems like just referencing `__force_section_ref` directly from the `#[used]` static is enough to get the component type custom section linked in.

crates/rust/src/lib.rs
```diff
@@ -377,9 +372,14 @@ impl InterfaceGenerator<'_> {
             "
                 #[allow(clippy::all)]
                 pub mod {snake} {{
+                    #[used]
+                    #[doc(hidden)]
+                    #[cfg(target_arch = \"wasm32\")]
+                    static __FORCE_SECTION_REF: fn() = crate::__link_section;
```
alexcrichton, line 378: Should this use a relative path of some form instead of the crate root to handle `generate!` in submodules?

alexcrichton (approved):

alexcrichton:
Also would it make sense to keep the current one for exports too in the case that something has exports but no imports?

crates/rust/src/lib.rs
peterhuene: Oh good point, yeah, `super` should do, I think.

peterhuene:
> Also would it make sense to keep the current one for exports too in the case that something has exports but no imports?

I'm not sure I understand. Do you want me to revert the change to the export macro for some reason? Exports without imports should still find the `__link_section` symbol like it did before.

alexcrichton:
Oh oops I misread the diff! What's here all looks good to me 👍

PR #545, opened 2023-03-27
Update how types are generated in C

### LINKED ISSUE #544: type/function redefinition when same tuple/list type appear in both typedef and funcdef
```wit
// ./wit/issue.wit
default world issue {
    union union-list-u8 {
        list<u8>,
    }

    record record-tuple-u8 {
        field: tuple<u8>,
    }
    
    variant variant-list-tuple-bool {
        field(list<tuple<bool>>),
    }

    import func-using-list-u8-in-args: func(body: list<u8>) -> ()
    export func-using-tuple-u8-in-return: func() -> tuple<u8>
    export func-using-list-tuple-bool-in-return: func() -> list<tuple<bool>>
}
```

just a simple WIT file, and using `wit-bindgen c --out-dir=gen ./wit`, it generates: 

<details><summary>Generated .h file</summary>

    // Generated by `wit-bindgen` 0.4.0. DO NOT EDIT!
    #ifndef __BINDINGS_ISSUE_H
    #define __BINDINGS_ISSUE_H
    #ifdef __cplusplus
    extern "C" {
    #endif
    
    #include <stdlib.h>
    #include <stdint.h>
    #include <stdbool.h>
    
    typedef struct {
      bool f0;
    } issue_tuple1_bool_t;
    
    typedef struct {
      issue_tuple1_bool_t *ptr;
      size_t len;
    } issue_list_tuple1_bool_t;
    
    typedef struct {
      uint8_t tag;
      union {
        issue_list_tuple1_bool_t field;
      } val;
    } issue_variant_list_tuple_bool_t;
    
    #define ISSUE_VARIANT_LIST_TUPLE_BOOL_FIELD 0
    
    typedef struct {
      uint8_t *ptr;
      size_t len;
    } issue_list_u8_t;
    
    typedef struct {
      uint8_t tag;
      union {
        issue_list_u8_t f0;
      } val;
    } issue_union_list_u8_t;
    
    typedef struct {
      uint8_t f0;
    } issue_tuple1_u8_t;
    
    typedef struct {
      issue_tuple1_u8_t field;
    } issue_record_tuple_u8_t;
    
    typedef struct {
      uint8_t *ptr;
      size_t len;
    } issue_list_u8_t;
    
    typedef struct {
      bool f0;
    } issue_tuple1_bool_t;
    
    typedef struct {
      issue_tuple1_bool_t *ptr;
      size_t len;
    } issue_list_tuple1_bool_t;
    
    typedef struct {
      uint8_t f0;
    } issue_tuple1_u8_t;
    
    // Imported Functions from `issue`
    void issue_func_using_list_u8_in_args(issue_list_u8_t *body);
    
    // Exported Functions from `issue`
    void issue_func_using_tuple_u8_in_return(issue_tuple1_u8_t *ret);
    void issue_func_using_list_tuple_bool_in_return(issue_list_tuple1_bool_t *ret);
    
    // Helper Functions
    
    void issue_list_tuple1_bool_free(issue_list_tuple1_bool_t *ptr);
    void issue_variant_list_tuple_bool_free(issue_variant_list_tuple_bool_t *ptr);
    void issue_list_u8_free(issue_list_u8_t *ptr);
    void issue_union_list_u8_free(issue_union_list_u8_t *ptr);
    void issue_list_u8_free(issue_list_u8_t *ptr);
    void issue_list_tuple1_bool_free(issue_list_tuple1_bool_t *ptr);
    
    #ifdef __cplusplus
    }
    #endif
    #endif
</details>

You can see it defines thoose type/function twice: 

- type `issue_list_u8_t` and function `issue_list_u8_free`
- type `issue_tuple1_bool_t`
- type `issue_list_tuple1_bool_t` and function `issue_list_tuple1_bool_free`
- type `issue_tuple1_u8_t`

Then it makes compiling fail.

I meet this issue when trying the tinygo generator, and the golang wrap type is defined twice too, so it seems to be a common behavior, is this by design?

And I tried type alias, this can fix redefinition issue,  so I should alias each complex type instead of using them directly?

alexcrichton:
Prior to this commit each interface used in C would generate a separate type per import and per export. For example if a `world` both imported and exported an interface then distinct, but defined the same way, types would be generated. Furthermore this same logic was extended to types-in-worlds which ended up not working well and causing #544.

This commit overhauls how types are generated in C. All used interfaces now generate one set of types, regardless of whether the interface is used in an import or an export. Additionally this enables fixing #544 by generating types for worlds only once at the end instead of incrementally throughout.

Along the way this updates many names of generated types to be more appropriate to their scoping constraints. For example types in interfaces are now namespaced by the name of the interface. Functions continue to be namespaced by the name of the import/export and disregard the name of the interface. Ambient types like `record<T, E>` continue to be prefixed by the name of the world that bindings are generated for.

Closes #544

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -23,7 +23,36 @@ struct C {
     needs_string: bool,
     world: String,
     sizes: SizeAlign,
+
+    // Known names for interfaces as they're seen in imports and exports.
+    //
+    // This is subsequently used to generate a namespace for each type that's
+    // used, but only in the case that the interface itself doesn't already have
+    // an original name.
     interface_names: HashMap<InterfaceId, String>,
+
+    // Interfaces who have had their types printed.
+    //
+    // This is used to guard against printing the types for an interface twice.
+    // The same interface can be both imported and exported in which case only
+    // one set of types is generated and all bindings for both imports and
+    // exports use that set of types.
+    interfaces_with_types_printed: HashSet<InterfaceId>,
+
+    // Type definitions for the given `TypeId`. This is printed topologically
+    // at the end.
+    types: HashMap<TypeId, wit_bindgen_core::Source>,
+
+    // The set of types that are considered public (aka need to be in the
+    // header file) which are anonymous and we're effectively monomorphizing.
+    // This is discovered lazily when printing type names.
+    public_anonymous_types: BTreeSet<TypeId>,
+
+    // This is similar to `public_anonymous_types` where it's discovered
+    // lazily, but the set here are for private types only used in the
+    // implementation of functions. These types go in the implementation file,
+    // not the header file.
+    private_anonymous_types: BTreeSet<TypeId>,
 }
 
 #[derive(Default, Debug, Clone)]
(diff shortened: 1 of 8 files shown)


Mossaka:
> cc @Mossaka while I haven't run the Go tests locally (haven't gotten around to installing the right toolchains) this is almost surely going to break Go bindings generation since it's relying on the C bindings generator.

Ack, I'll take a look at it later today

Mossaka:
In the meantime, feel free to disable Go bindgen tests.

pchickey (approved):

PR #547, opened 2023-03-28
Don't generate `*Param` and `*Result` types in Rust by default

### LINKED ISSUE #535: Offer option to disable generation/use of borrowed types for Rust
G'day again!

Under certain scenarios, `wit-bindgen` (and Wasmtime, which uses similar machinery) will use borrowed types instead of owned types for arrays and strings. This works fine when it's for single-level owned types, like `String -> &str` or `Vec<u8> -> &[u8]`.

Unfortunately, having nested owned types - like `Vec<String>` - results in the borrowed equivalent being `&[&str]`, which is quite tedious to work with, especially with converting it back and forth to the owned representation.

In our project, I've had to define [a macro](https://github.com/AmbientRun/Ambient/blob/e96a252355b823b1a4ea8a834c8ac7a3dce15c17/crates/wasm/src/shared/borrowed_types.rs) to generate intermediate types to convert the owned representation of [our value variant type](https://github.com/AmbientRun/Ambient/blob/e96a252355b823b1a4ea8a834c8ac7a3dce15c17/crates/wasm/wit/component.wit#L8-L68) (as seen in #526) to the borrowed representation. This requires allocations, which feels like it defeats the purpose of borrowing to begin with.

To alleviate this, one of the following solutions would work for me (in order of bluntness):

1) Disabling the generation of borrowed types outright
2) Disabling the generation of borrowed types for a specific type
3) Limiting the generation of borrowed types to a single level, so that `Vec<String> -> &[String]` and not `&[&str]` or `Vec<&str>`.

Our current solution works fine, but it needs to be updated on both the host and the guest when new cases are added to the variant, which ends up being a bit of a maintenance hazard.

alexcrichton:
Previously when a single WIT type was used in a "borrowed" position, or a parameter to an import, and an "owned" position then the type would have two copies of its definition generated if it internally contained a list. This was intended to signify that for import parameters ownership wasn't necessary so borrowed values could be passed in.

In practice though I don't think this was the right default to have. This surprisingly generates two types when one might expect one and otherwise if a type if received from an export and expected to be passed to an import then it's not easy to do so since it must be converted to the borrowed type.

This commit moves the preexisting behavior of generating duplicate types behind a new `duplicate_if_necessary` option. The default behavior is to no longer generate two types, but instead just one. Imports now generate their types with a `&` in front if it's otherwise owned internally to signify that ownership isn't required.

Closes #535

### DIFF
--- a/crates/rust-lib/src/lib.rs
+++ b/crates/rust-lib/src/lib.rs
@@ -16,6 +16,16 @@ pub trait RustGenerator<'a> {
     fn resolve(&self) -> &'a Resolve;
     fn path_to_interface(&self, interface: InterfaceId) -> Option<String>;
 
+    /// This, if enabled, will possibly cause types to get duplicate copies to
+    /// get generated of each other. For example a record containing a string
+    /// used both in the import and export context would get one variant
+    /// generated for both.
+    ///
+    /// If this is disabled then the import context would require the same type
+    /// used for the export context, which has an owned string that might not
+    /// otherwise be necessary.
+    fn duplicate_if_necessary(&self) -> bool;
+
     /// Return true iff the generator should qualify uses of `std` features with
     /// `#[cfg(feature = "std")]` in its output.
     fn std_feature(&self) -> bool {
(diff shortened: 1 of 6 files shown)


pchickey (approved):
My experience so far is that this is a better default. Thanks!

philpax:
Fantastic, thank you - I'm looking forward to trying this out!

PR #552, opened 2023-03-29
Relax an assertion in Rust type emission

### LINKED ISSUE #551: Assertion failure in `print_tyid` when using a list of lists in both param/return form and with a `type` definition
I was testing out #547 with our bindings, and encountered this assert:
https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/rust-lib/src/lib.rs#L245

This is the minimal repro:
```wit
default world bindings {
    import component: self.component
}

interface component {
    type value = list<tuple<>>
    type entity = list<tuple<u32, value>>

    add-components: func(entity: u64, data: entity)
    query-eval: func(q: u64) -> list<tuple<u64, list<value>>>
}
```

Note that the `type entity` is necessary; substituting its definition into `add-components` allows for generation to pass.

This is the resulting callstack:
```
% RUST_BACKTRACE=1 wit-bindgen rust wit -w main
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `"'a"`,
 right: `"'_"`', wit-bindgen-f98ccff846621481/58c130a/crates/rust-lib/src/lib.rs:245:21
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:64:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
   4: wit_bindgen_rust_lib::RustGenerator::print_tyid
   5: wit_bindgen_rust_lib::RustGenerator::print_ty
   6: wit_bindgen_rust_lib::RustGenerator::print_tyid
   7: wit_bindgen_rust_lib::RustGenerator::print_ty
   8: wit_bindgen_rust_lib::RustGenerator::print_rust_slice
   9: wit_bindgen_rust_lib::RustGenerator::print_type_list
  10: <wit_bindgen_rust::RustWasm as wit_bindgen_core::WorldGenerator>::import_interface
  11: wit_bindgen_core::WorldGenerator::generate
  12: wit_bindgen::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

(with the latest `wit-bindgen-cli` installed through `cargo install --git https://github.com/bytecodealliance/wit-bindgen.git wit-bindgen-cli`)

alexcrichton:
I forgot a context where the assertion would trip, so implement the code necessary to avoid the need for the assertion.

Closes #551

### DIFF
--- a/crates/rust-lib/src/lib.rs
+++ b/crates/rust-lib/src/lib.rs
@@ -242,8 +242,11 @@ pub trait RustGenerator<'a> {
             // the API level, ownership isn't required.
             if info.has_list && lt.is_none() {
                 if let TypeMode::AllBorrowed(lt) | TypeMode::LeafBorrowed(lt) = mode {
-                    assert_eq!(lt, "'_");
                     self.push_str("&");
+                    if lt != "'_" {
+                        self.push_str(lt);
+                        self.push_str(" ");
+                    }
                 }
             }
             let name = self.type_path(id, lt.is_none());
--- a/tests/codegen/issue551.wit
+++ b/tests/codegen/issue551.wit
@@ -1,0 +1,11 @@
+default world bindings {
+    import component: self.component
+}
+
+interface component {
+    type value = list<tuple<>>
+    type entity = list<tuple<u32, value>>
+
+    add-components: func(entity: u64, data: entity)
+    query-eval: func(q: u64) -> list<tuple<u64, list<value>>>
+}


pchickey (approved):

philpax:
I can confirm that this works! Thanks for the quick fix 🚀

PR #560, opened 2023-04-27
docs: link to non-command wasi_snapshot_preview1

danbev:
This commit updates the `non-command` link to
`wasi_snapshot_preview1.reactor.wasm`.

The motivation for this is that the current link results in a 404 Not Found.


alexcrichton:
Thanks!

PR #580, opened 2023-05-18
Update wit-bindgen for upcoming WIT changes

alexcrichton:
This PR is based on https://github.com/bytecodealliance/wasm-tools/pull/1027 and  https://github.com/bytecodealliance/wasmtime/pull/6390 and updates the C and Rust guest generators in this repository for the new syntax. Additionally the Wasmtime host test harness is updated with the new syntax and the new constructs.

This commit disables the TeaVM Java generator (cc @dicej) and the TinyGo generator (cc @Mossaka) entirely since from the work on Rust and C I think that these will need relatively substantial updates related to namespacing. I'm happy to help out but I'm hoping that y'all are able to help out with the update as well.

PR #585, opened 2023-05-26
update go generator to re-enable codegen-tests

Mossaka:
This fixes the breaking change introduced in https://github.com/bytecodealliance/wit-bindgen/pull/580 and re-enable all the codegen-tests

I realized that I still need to re-enable all the runtime tests that were disabled in #545 . Will work on that. In the meantime this PR can merge without being blocked.


alexcrichton:
Thanks!

PR #587, opened 2023-05-26
fix(go): fix go test files to re-enable runtime tests

Mossaka:
This re-enabled all the existing runtime tests for go generator which are turned off by #545 

Unfortunately the types and function names became much more verbose than before :(


alexcrichton:
Oops it appears I merged https://github.com/bytecodealliance/wit-bindgen/pull/588 before this, but that includes this so I'm going to close this. (let me know if that's in error though)

PR #598, opened 2023-06-27
provide more control over type ownership

dicej:
This is the guest version of https://github.com/bytecodealliance/wasmtime/pull/6648.

This replaces the duplicate_if_necessary parameter to the Rust binding generator with a new ownership parameter which provides finer-grained control over whether and how generated types own their fields.

The default is `Ownership::Owning`, which means types own their fields regardless of how they are used in functions. These types are passed by reference when used as parameters to guest-exported functions. Note that this also affects how unnamed types (e.g. `list<list<string>>`) are passed: using a reference only at the top level (e.g. `&[Vec<String>]` instead of `&[&[&str]]`, which is more difficult to construct when using non-'static data).

The other option is `Ownership::Borrowing`, which includes a `duplicate_if_necessary` field, providing the same code generation strategy as was used prior to this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1862,6 +1862,7 @@ dependencies = [
  "syn 2.0.15",
  "wit-bindgen-core",
  "wit-bindgen-rust",
+ "wit-bindgen-rust-lib",
  "wit-component",
 ]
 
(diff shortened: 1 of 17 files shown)


dicej:
CI and I just aren't getting along today.  I can't repro the Linux failure.  I'll see if I can repro the Windows one tomorrow.

alexcrichton (approved):

PR #602, opened 2023-06-30
fix symbol conflicts for post return

xxchan:
follow-up of #593


xxchan:
cc @alexcrichton

alexcrichton:
Thanks!

PR #606, opened 2023-07-06
Alias anon types in tiny-go

rylev:
This fixes a bug where previously when an anonymous type had already been defined, we simply replaced the definition. However, this did not work as other parts of the codegen could already be referring to the anon type by that previous name. 

This change keeps track of all the names used for a type. Then in the codegen for an anonymous type, it first looks to see if the type was already defined, and if so, it generates aliases for the previously used name. 

I imagine we'll want to add a test that would have caught this. Do we want to add a new test or to change an existing test to exercise this? I think it should be as simple some `use`s in the wit file.

### DIFF
--- a/crates/c/tests/codegen.rs
+++ b/crates/c/tests/codegen.rs
@@ -34,7 +34,8 @@ macro_rules! codegen_test {
 test_helpers::codegen_tests!();
 
 fn verify(dir: &Path, name: &str) {
-    let path = PathBuf::from(env::var_os("WASI_SDK_PATH").unwrap());
+    let path =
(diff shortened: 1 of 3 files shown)


Mossaka (approved):
lgtm!

Mossaka:
> . Do we want to add a new test or to change an existing test to exercise this?

Let's add a new codegen test. @rylev

rylev:
@Mossaka I extended an existing test because that actually seemed easier. I confirmed that the test fails without the fix.

Mossaka (approved):
LGTM!

Issue #607, opened 2023-07-11
Conflicting anonymous types in C generator


rylev:
When creating C types to represent anonymous types used in an interface, the C generator creates the same C type name for what should be two distinct types. Here's a wit file that fails to generate valid C code:

```wit
package local:demo

interface interface1 {
    variant error {
        some-error
    } 
    my-other-func: func() -> result<_, error>
}

interface interface2 {
    variant error {
        other-error
    } 
    my-other-func: func() -> result<_, error>
}

world my-world {
    import interface1
    import interface2
}
```

The two functions return two result types that happen to both be `result<_, error>` but the `error` in each interface is a different error type and thus they should not be treated as the same. However, the C generator generates a type with the name `my_world_result_void_error_t`. 

I believe this is the same issue for all anonymous types not just types in the `error` position of `result` .It seems to me the solution would be to namespace types used inside of anonymous types so that they are unambiguous.

PR #608, opened 2023-07-12
C: Namespace types inside of annoymous types

rylev:
This fixes the issue in C for #607, but unfortunately does not yet fix the issue in the Go generator as the Go generator uses bespoke logic for rebuilding C type names. I wanted to get this up first to initial feedback, but my plan is to fix the issue in the Go generator tomorrow.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -705,7 +705,11 @@ impl C {
     }
 
     fn owner_namespace(&self, resolve: &Resolve, id: TypeId) -> String {
-        owner_namespace(resolve, id, &self.interface_names, &self.world)
+        owner_namespace(resolve, id, &self.interface_names).unwrap_or_else(|| {
+            // Namespace everything else under the "default" world being
+            // generated to avoid putting too much into the root namespace in C.
+            self.world.to_snake_case()
+        })
     }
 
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks! Seems reasonable enough to me, but I find it easiest to review updated tests. I'd be surprised if none of `tests/runtime/*/wasm.c` need updating from this, and my guess is that testing in CI failed on an earlier crate before it got there?

rylev:
@alexcrichton - it doesn't seem like it, but that's because the runtime tests don't exercise the issue. We've currently only made the change for annoymous types that have optional inner types, but this is still an issue for annoymous types that don't have optional inner types like `option`. This world breaks the C generator:

```wit
package local:demo

interface interface1 { 
    record my-record {
        field: u32
    }
    optional: func() -> option<my-record>
}

interface interface2 {
    record my-record {
        other-field: u32
    }
    other-optional: func() -> option<my-record>
}

world my-world {
    import interface1
    import interface2
}
```

The interfaces look identical if you only look at the top-level types, as the C generator does, but they're not actually equivalent because the inner types `my-record` are not equivalent. I'll push a fix for this to this PR.

alexcrichton:
Ah ok makes sense! This all seems good to me but I think it may need a rebase now, after that though with a green CI I'm happy to merge

rylev:
@alexcrichton done!

PR #611, opened 2023-07-12
Ensure that structs and unions in C are never empty.

sunfishcode:
Empty structs and unions are a non-standard extension in C, and have a different layout between C and C++, so avoid emitting them. If a generated struct or union would have zero members, give it a `u8 __empty;` member so that it's not actually empty, so that it gets a consistent layout no matter how it's being compiled.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -514,13 +514,19 @@ impl C {
                     union {
                 ",
                 );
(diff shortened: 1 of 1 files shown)


sunfishcode:
This approach of inserting `__empty` fields isn't working well, so I'm going to take a different approach here.

PR #613, opened 2023-07-12
Omit structs and unions that would be empty.

sunfishcode:
Empty structs and unions are a non-standard extension in C, and have a different layout between C and C++, so avoid emitting them. If a generated struct or union would have zero members, avoid emitting it.


alexcrichton:
I believe this is resolved by https://github.com/WebAssembly/component-model/pull/218, so closing

PR #619, opened 2023-07-14
Rust: add option for path to wit-bindgen runtime.

peterhuene:
This PR adds an optional path to the `wit-bindgen` runtime to use for Rust code generation.

This will allow a different crate from `wit-bindgen` to generate bindings while also re-exporting `wit-bindgen::rt` at a known path for the generated code to use.

### DIFF
--- a/crates/rust-lib/src/lib.rs
+++ b/crates/rust-lib/src/lib.rs
@@ -65,11 +65,11 @@ pub trait RustGenerator<'a> {
         true
     }
 
-    /// Return the fully-qualified name for the `Vec` type to use.
-    fn vec_name(&self) -> &'static str;
+    /// Pushes the name of the `Vec` type to use.
+    fn push_vec_name(&mut self);
 
-    /// Return the fully-qualified name for the `String` type to use.
-    fn string_name(&self) -> &'static str;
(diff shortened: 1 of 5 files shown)


pchickey (approved):
Thanks! can you add a test which invokes this with e.g. `pub mod my_rt { pub use wit_bindgen::rt::*;}`?

peterhuene:
I've added a `runtime_path` option to the `generate!` macro so we can set it via a test, which I've also added.

peterhuene:
@alexcrichton thoughts on doing a release that might include this change? 

This will enable `cargo-component` to migrate to using a `wit-bindgen::generate!`-style macro ahead of the resource changes and we've also rev'd the `wasm-tools` dependencies in main, which would be nice to have for downstream crates to sync on.

I realize that we'll want a release in the (near?) future to support resources, but an interim release would definitely help me out.

alexcrichton:
Sounds good to me! I'll send a version bump to this repo which will be good to pull in the wasm-tools update regardless.

alexcrichton:
Posted a bump at https://github.com/bytecodealliance/wit-bindgen/pull/620

PR #621, opened 2023-07-17
Consistently add generated preamble

rylev:
This adds the generated preamble "Generated by `wit-bindgen` {version}. DO NOT EDIT!" to the generated Go files, and refactors all code generators to use the same text.


rylev:
@alexcrichton I'm not sure what you mean. The version is already being passed from the leaf crate so they should be able to keep independent versions.

https://github.com/bytecodealliance/wit-bindgen/blob/217baaeb72e7bb7be6dc9fd491c6f1e0f208af74/crates/core/src/lib.rs#L346

alexcrichton:
Wow today is just not my day...

Ignore me! I think there's errors in CI though?

rylev:
@alexcrichton fixed!

crates/go/src/lib.rs
```diff
@@ -242,35 +242,33 @@ impl WorldGenerator for TinyGo {
...
-        let mut header = Source::default();
+        // prepend package and imports header
+        let src = mem::take(&mut self.src);
+        wit_bindgen_core::generated_preamble(&mut self.src, env!("CARGO_PKG_VERSION"));
         let snake = self.world.to_snake_case();
         // add package
-        header.push_str("package ");
-        header.push_str(&snake);
-        header.push_str("\n\n");
+        self.src.push_str("package ");
```
Mossaka, line 250: Thanks for refactoring!

crates/go/tests/codegen.rs
```diff
@@ -43,8 +43,10 @@ fn verify(dir: &Path, name: &str) {
         .open(&main)
         .expect("failed to open file");
     let mut reader = BufReader::new(file);
-    reader.read_until(b'\n', &mut Vec::new()).unwrap();
     let mut buf = Vec::new();
+    reader.read_until(b'\n', &mut buf).unwrap();
+    // Skip over `package $WORLD` line
+    reader.read_until(b'\n', &mut Vec::new()).unwrap();
```
Mossaka, line 49: Could you please help me to understand what this line is doing?

tests/runtime/main.rs
```diff
@@ -264,6 +264,7 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
         };
 
         if !output.status.success() {
+            println!("dir: {}", out_dir.display());
```
Mossaka, line 267: Is this added for debugging purposes?

Mossaka (approved):
Thanks for making this change and LGTM!

crates/go/tests/codegen.rs
rylev: The comment above is trying to explain. There's also a much larger comment before the entire code block that explains in more detail. Essentially, tinygo can't compile the generated code as is because it requires a main package and function. This code just munges the generated a bit to put it into compileable form. This need to change since the first line of the generated code is no longer the package name but the new comment.

tests/runtime/main.rs
rylev: Well yes, but everything in that code block is for debugging purposes so this just makes the output more verbose. I can reverse it if you want, but then I'd just open a new PR adding it again 😀

crates/go/tests/codegen.rs
Mossaka: > This need to change since the first line of the generated code is no longer the package name but the new comment.

Ah that is what I was missing. Thank you!

tests/runtime/main.rs
Mossaka: I am fine with checking this in!

PR #627, opened 2023-07-21
Adjustments to Rust guest code generation

alexcrichton:
This commit refactors the exact shape of the Rust guest code generated by `wit-bindgen`. The original problem here was that `rustfmt` would fail on the generated output due to extreme nesting and extreme line lengths presumably causing `rustfmt`'s heuristics to give up. While the issue has been fixed for the input WIT world this is a bit of a game of whack-a-mole so there's not a great regression test suite to have here.

In general the changes here are:

* Where possible move library functionality into the guest crate which generated code calls into. For example removing branches with `cfg(debug_assertions)`.
* Trying to de-indent code by less using nested Rust expressions and moreso using `let foo = ...` and propagate `foo` as the result of an operation.
* Enum lifting operations are now helper functions to avoid generating a lift-per-callsite which can be quite a lot for errors.
* Various other minor adjustments as aliasing variant names as a local `V{N}` instead of `super::super::super::your::here::Type` or restructuring variant lifts/lowers to have less nesting and/or be easier for `rustfmt` to format.

I think there's still a good bit to improve here. For example the generated code has a lot of blocks such as:

    for ... {
        let base = ...;
        {
            // use `base`
        }
    }

where the body of the `for` loop has an unnecessary block. I've left fixing this for a future refactoring though.

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -1327,6 +1327,31 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
 
     fn type_enum(&mut self, id: TypeId, name: &str, enum_: &Enum, docs: &Docs) {
         self.print_typedef_enum(id, name, enum_, docs, &[], Box::new(|_| String::new()));
+
+        let name = name.to_upper_camel_case();
+        let mut cases = String::new();
+        let repr = int_repr(enum_.tag());
+        for (i, case) in enum_.cases.iter().enumerate() {
+            let case = case.name.to_upper_camel_case();
+            cases.push_str(&format!("{i} => {name}::{case},\n"));
+        }
+        uwriteln!(
+            self.src,
+            r#"
+                impl {name} {{
+                    pub(crate) unsafe fn _lift(val: {repr}) -> {name} {{
+                        if !cfg!(debug_assertions) {{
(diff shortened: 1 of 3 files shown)


crates/rust/src/lib.rs
```diff
@@ -884,6 +884,31 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
...
+        uwriteln!(
+            self.src,
+            r#"
+                impl {name} {{
+                    pub(crate) unsafe fn _lift(val: {repr}) -> {name} {{
+                        if !cfg!(debug_assertions) {{
+                            return ::core::mem::transmute(val);
+                        }}
+
+                        match val {{
```
cpetig, line 904: this feels like conditionally dead code, I would prefer either an else (which would indent more) or `#[cfg(debug_assertions)]`.

alexcrichton: True it is conditionally dead code! Personally though I prefer to avoid conditional compilation where possible because that enables rustc to typecheck both implementations and avoids lints such as unused imports as well. The actual codegen-time impact here in theory is quite negligible.

cpetig:
I am surprised no-one else reviews this MR, I really like this change and above point was my only minor question about it.

sunfishcode (approved):

PR #637, opened 2023-08-07
Remove reliance on `wit_parser::abi`

alexcrichton:
This commit is the dual of bytecodealliance/wasm-tools#1149 where the code is being moved from the `wit-parser` crate into this repository where it's mainly developed. This commit adds a `wit_bindgen_core::abi` module with the same contents as before with small refactorings to avoid inherent methods on `Resolve`.

Note that this temporarily switches to a git dependency for the wasm-tools family of crates which will be reverted back to a crates.io dependency one they're published.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1206,8 +1206,8 @@ name = "test-helpers"
 version = "0.0.0"
 dependencies = [
  "codegen-macro",
- "wasm-encoder 0.31.1",
- "wat",
+ "wasm-encoder 0.31.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wat 1.0.69 (registry+https://github.com/rust-lang/crates.io-index)",
  "wit-bindgen-core",
  "wit-component",
  "wit-parser 0.9.2",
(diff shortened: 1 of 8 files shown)


guybedford:
Nice! I assume the interface is mostly unchanged in the move?

alexcrichton:
Yeah this is intended to be a copy/paste, and the only real difference is that instead of `self.resolve.foo(...)` it's now `abi::foo(self.resolve, ...)`

cpetig:
Also a "Nice!" from me. The post-return change in wasm-tools looks worth its own PR as well.

guybedford:
I was able to update jco to this branch and can confirm it all works as expected.

guybedford (approved):

peterhuene:
With this now merged, any chance we can do a wasm-tools and wit-bindgen release?

Without one, `cargo-component` has to generate projects with the same `patch.crates-io` in it for the wasm-tools dependencies to get things building (currently it references wit-bindgen by git for the resources support), which isn't great.

alexcrichton:
Sounds good to me! I'll work on the bump PRs later today if no one else has beaten me to it by then.

Wow and I forgot to send this message until I'm ready to do the bumps, sorry about that!

PR #638, opened 2023-08-07
fix resource import/export module names

### LINKED ISSUE #635: Rust: exporting a resource results in `exports::` prefix in exported name
For the following wit in a `wit` subdir:

```wit
package test:export-resource

world test {
    export foo: interface {
        resource x {
            constructor()
        }
    }
}
```

and the following Rust code:

```rust
wit_bindgen::generate!({
    path: "wit",
    exports: {
        "foo/x": MyX
    }
});

use exports::foo::X;

pub struct MyX;

impl X for MyX {
    fn new() -> Self {
        Self
    }
}
```

The produced module fails to componentize:

```
$ wasm-tools component new foo.wasm
error: failed to encode a component from module

Caused by:
    0: import from `[export]exports::foo` does not correspond to exported interface
```

I believe [this line](https://github.com/bytecodealliance/wit-bindgen/blob/bf318b8ae4703586ae236a232f62813efc946df3/crates/rust/src/lib.rs#L873) is adding the `exports::` prefix to the interface name.

It appears we might be relying on that for getting the proper path to the interface from the bindings source, but when generating the `#[link(wasm_import_module = "..."]` or `#[export_name = "..."]` decorators the `exports::` prefix should not be present.

dicej:
Both the C and Rust generators were generating these names incorrectly, at least in some cases.  I believe this addresses that, although I've had a hard time finding definitive documentation about how these names should be constructed.

Fixes #635

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -518,11 +518,12 @@ impl C {
             let namespace = self.owner_namespace(resolve, *id);
             let name = resolve.types[*id].name.as_deref().unwrap();
             let snake = name.to_snake_case();
+            let module = self.owner_wasm_namespace(resolve, *id);
 
             for s in [&mut *h_str, &mut *c_str] {
                 uwriteln!(
                     s,
-                    "\n// Functions for working with resource `{namespace}::{name}`"
+                    "\n// Functions for working with resource `{module}/{name}`"
                 );
             }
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
> I've had a hard time finding definitive documentation about how these names should be constructed

Ah yeah this is where there's no actual documentation on this. The component model doesn't define the operation of taking a core wasm module and converting it to a component, so that's left entirely to `wit-component` which is scant on documentation. This is something we've talked about though for upstreaming and/or codifying in the component model itself.

alexcrichton (approved):

Issue #636, opened 2023-07-31
Resource bindings use incorrect names for drop functions.


peterhuene:
It appears that wit-bindgen is generating resource drop imports of `[resource-drop-borrow]` and `[resource-drop-own]` instead of `[resource-drop]` that is currently expected by `wit-component`'s validator.

cc: @dicej @alexcrichton

alexcrichton:
Fixed in #642

PR #642, opened 2023-08-11
changed to correct name for drop functions

silesmo:
Changes resource drop function names from `resource-drop-{own|borrow}` to `resource-drop`
Solves #636

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -556,7 +556,7 @@ impl C {
 
                 uwriteln!(
                     c_str,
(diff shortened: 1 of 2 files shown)


cpetig:
Looks correct to me, as per Alex' explanation on https://github.com/bytecodealliance/wit-bindgen/issues/636

peterhuene:
I think Alex might be out this week, so I'll approve based off of what he described earlier.

peterhuene (approved):

peterhuene:
@silesmo thanks for the fix!

PR #652, opened 2023-08-24
Fix Rust example in README

### LINKED ISSUE #649: Example in readme does not work
wit/host.wit
```
package example:host

world host {
  import print: func(msg: string)

  export run: func()
}
```

src/lib.rs
```rust
wit_bindgen::generate!("host");
```

Cargo.toml
```toml
[lib]
crate-type = ["cdylib"]

[dependencies]
wit-bindgen = { git = "https://github.com/bytecodealliance/wit-bindgen", version = "0.10.0" }
```

Error message
```plaintext
error: export parameter required for `world`
 --> wasi-wit/src/lib.rs:1:1
  |
1 | wit_bindgen::generate!("host");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |

```

alexcrichton:
Closes #649


pchickey (approved):

LLFourn:
FWIW there's another example of the old invocation in:

https://github.com/bytecodealliance/wit-bindgen/blob/749c01697bb3b11daeae4225789e14b765dcf839/crates/wasi_snapshot_preview1/src/lib.rs#L21

alexcrichton:
Thanks @LLFourn but that's actually slightly different where because there are no exports that doesn't actually need to change, although there's still the question of whether the old syntax should be supported at all since almost everything has at least one export.

In any case I'm going to go ahead and merge this and we can continue to iterate afterwards.

PR #655, opened 2023-08-25
Refactor how resources are generated in Rust

alexcrichton:
This commit is aimed at refactoring the internals of resources in Rust to achieve a few goals:

* Abstract the underlying representation of `Box<T>` into a library-defined representation rather than codegen-defined. This helps generate less code and makes it easier to read/test/understand in isolation.

* Reduce the amount of resource-specific logic throughout the code generator by trying to stick to "iterate over this list of functions" with minor handling here and there for when the functions turn out to be resource-related methods/etc.

* Move resource-related bits together in the code generator, for example `type_resource` now generates impls/etc for resources as one central location.

For now both imported and exported resources are using a new `Resource<T>` type where `T` contains the implementation of the intrinsics for the component model. In the future I'd like to change resources to internally use `RefCell<T>` to enable methods to take `&mut self` instead of `&self` which is part of the motivation for this change, but that's not done yet.

Additionally some of the refactorings in the code generator reflect the new structure of generated code. For example the type for a resource export may be imported twice now instead of just once -- once for the type itself (`pub use`'d under a well-known name) and once for export implementations (under a private `_*Impl` name).

### DIFF
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -2,6 +2,12 @@
 
 extern crate alloc;
 
+use alloc::boxed::Box;
+use core::fmt;
+use core::marker;
+use core::mem::ManuallyDrop;
+use core::ops::{Deref, DerefMut};
+
 #[cfg(feature = "macros")]
 pub use wit_bindgen_rust_macro::*;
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
cc @dicej

crates/guest-rust/src/lib.rs
```diff
@@ -163,3 +171,120 @@ pub mod rt {
...
+    }
+}
+
+impl<T: RustResource> Deref for Resource<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        unsafe {
+            let rep = T::rep(self.handle);
+            &*(rep as *mut T)
```
juntyr, line 262:
```suggestion
            &*(rep as *const T)
```

dicej (approved):

Issue #657, opened 2023-08-26
Possibly outdated README.md


gibbz00:
https://github.com/bytecodealliance/wit-bindgen/blob/86401dc72aa872b02a44c7e56064d22d4385b1cb/README.md?plain=1#L232 might be outdated since https://github.com/bytecodealliance/wit-bindgen/pull/651 was merged 😊

PR #658, opened 2023-08-27
docs(README): changed rust example Host -> Guest

kjuulh:
Updated the rust example in the readme. 

I haven't tested the other languages mentioned in the readme, but they're probably also outdated.

fyi: #657


alexcrichton:
Thanks!

PR #664, opened 2023-09-11
C: Add `extern` to header files for provided functions

### LINKED ISSUE #663: Add the extern keyword to imported functions in C generated header files
The wit-bindgen c generates a header file with stub functions both for imported and exported functions. To help the c developer know which functions require an implementation can you place the `extern` keyword to the function definition of the imported functions?

alexcrichton:
Help syntactically show which functions are implemented by users and which are imported from the environment. I don't think `extern` actually does anything here and most of the functions actually have shims in the C files generated, but I also don't believe that `extern` harms anything either.

Closes #663

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -569,7 +569,7 @@ impl C {
 
                 uwriteln!(
                     h_str,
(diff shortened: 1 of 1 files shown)


Mossaka (approved):
lgtm!

PR #665, opened 2023-09-11
Update string representation to not `char*`

### LINKED ISSUE #662: C string bindings generate char*, change to uint8_t*
`wit-bindgen c ` generates a `struct` to deal with strings. However the suggestion is that the structure are going to be UTF8, most c developers would assume that a `char*` pointer would point to an ASCII string, they are likely to use c's `strlen`  or `strcmp`, etc, family of functions on these pointers, but of course these won't work. 

To avoid c developer confusion, can we change this to generate uint8_t* types ?

Original:
````
typedef struct {
  char*ptr;
  size_t len;
} greeter_string_t;
````

Desired:
````
typedef struct {
  uint8_t*ptr;
  size_t len;
} greeter_string_t;
````

alexcrichton:
Use `uint8_t*` to represent that it's not a normal `char*` since it's neither nul-terminated and it may also contain interior nul characters.

Closes #662

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -284,23 +284,35 @@ impl WorldGenerator for C {
                 StringEncoding::CompactUTF16 => unimplemented!(),
             };
             let ty = self.char_type();
+            let c_string_ty = match self.opts.string_encoding {
+                StringEncoding::UTF8 => "char",
+                StringEncoding::UTF16 => "char16_t",
+                StringEncoding::CompactUTF16 => panic!("Compact UTF16 unsupported"),
+            };
             uwrite!(
                 self.src.h_helpers,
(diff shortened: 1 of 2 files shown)


alexcrichton:
@Mossaka would you be able to assist with the Go failures here? I could add some exceptions to ignore the tests but that seems infeasible in this case because it'd break all usage of strings from Go

Mossaka:
Yup I'll take a look at today. Thanks for letting me know

alexcrichton:
Thanks! Not super urgent, but there may also be a simple fix too

Mossaka:
@alexcrichton I just added a commit, should fix the Go failures.

Mossaka (approved):

PR #669, opened 2023-09-14
Fix taking owned resource handles in Rust imports

### LINKED ISSUE #668: Rust: error in record with resource
An example for a wit file that cause the error:
```wit
package test:test

interface test-import {
  resource resource-a {
    constructor(id: u32)
  }

  record record-a {
    resource-a: resource-a,
    resources: list<resource-a>,
  }

  resource resource-b {
    make: static func(record-a: record-a)
  }
}

world test-world {
  import test-import

}
```

With this rust code:
```rust
wit_bindgen::generate!({
    world: "test-world",
});
```

I get the following error:
```bash
error[E0507]: cannot move out of `*resource_a0` which is behind a shared reference
 --> src/lib.rs:1:1
  |
1 | / wit_bindgen::generate!({
2 | |     world: "test-world",
3 | | });
  | |  ^
  | |  |
  | |__`*resource_a0` moved due to this method call
  |    move occurs because `*resource_a0` has type `ResourceA`, which does not implement the `Copy` trait
  |
note: `ResourceA::into_handle` takes ownership of the receiver `self`, which moves `*resource_a0`
 --> src/lib.rs:1:1
  |
1 | / wit_bindgen::generate!({
2 | |     world: "test-world",
3 | | });
  | |__^

For more information about this error, try `rustc --explain E0507`.
error: could not compile `wasm-component-test` (lib) due to previous error
```

The combination that cause the error is a field of a resource and a another field with a list of the same type resource. If I remove one of the field, it compiles with no error.

alexcrichton:
I went ahead and did a bit more refactoring at the logic here since I think the old conditions aren't as applicable any more (they haven't aged well)

Closes #668

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -42,6 +42,9 @@ pub struct TypeInfo {
 
     /// Whether or not this type (transitively) has a borrow handle.
     pub has_borrow_handle: bool,
+
+    /// Whether or not this type (transitively) has an own handle.
+    pub has_own_handle: bool,
 }
 
 impl std::ops::BitOrAssign for TypeInfo {
@@ -52,6 +55,7 @@ impl std::ops::BitOrAssign for TypeInfo {
         self.has_list |= rhs.has_list;
         self.has_resource |= rhs.has_resource;
         self.has_borrow_handle |= rhs.has_borrow_handle;
+        self.has_own_handle |= rhs.has_own_handle;
     }
 }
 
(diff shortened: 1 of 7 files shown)


alexcrichton:
I'll also note that the exact way to get this all working in Rust is sort of a perpetual game of whack-a-mole. Short of setting up fuzzing to suss out issues I'm personally content to add individual test cases as they come up and constrain the logic to 'seems fine if it passes tests'

PR #670, opened 2023-09-18
Fix generated bindings for list-of-borrows

alexcrichton:
This commit fixes an issue for imported functions taking `list<borrow<T>>` arguments. Previously they were erroneously passed through as lists-of-pointers and now they're correctly mapped to list-of-handle-indexes in the ABI. This required rejiggering a bit how types are generated and what's considered owned when, namely requiring that all own handles use `Vec<T>` at the boundary since they need to take ownership.

### DIFF
--- a/crates/rust-lib/src/lib.rs
+++ b/crates/rust-lib/src/lib.rs
@@ -337,7 +337,9 @@ pub trait RustGenerator<'a> {
             // If the type recursively owns data and it's a
             // variant/record/list, then we need to place the
             // lifetime parameter on the type as well.
-            if (info.has_list || info.has_borrow_handle) && needs_generics(self.resolve(), &ty.kind)
+            if (info.has_list || info.has_borrow_handle)
+                && !info.has_own_handle
+                && needs_generics(self.resolve(), &ty.kind)
             {
                 self.print_generics(lt);
             }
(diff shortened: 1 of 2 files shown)


alexcrichton:
I'll note that testing-wise the existing codegen tests cover a good deal, and otherwise tests for resources are not present at the moment as this requires wasmtime support which is still nascent.

crates/rust-lib/src/lib.rs
```diff
@@ -470,6 +472,11 @@ pub trait RustGenerator<'a> {
         } else {
             mode
         };
+        // Lists with `own` handles must always be owned
+        let mode = match *ty {
+            Type::Id(id) if self.info(id).has_own_handle => TypeMode::Owned,
+            _ => mode,
+        };
```
sunfishcode, line 479: I'm curious why lists are special, necessitating special code in `print_list`. Wouldn't the same issue come up in records and variants?

alexcrichton: I know I sounds like a broken record but this was basically the mole that came up that I figured I'd whack.

The Rust generator tries to keep things borrowed where it can but that support has not aged well. This, AFAIK, is only here now to handle the non-"owning" mode, which is a non-default setting, where types may be duplicated to aggressively borrow things. Bare types already have special logic for handles and this is sort of copying said special logic to lists-of-handles as well since "this requries owned" is a transitive property through the list.

Not a great solution by any means for sure, and I'm hoping that we can basically just build a corpus of programs to keep working into the future.

sunfishcode (approved):

crates/rust-lib/src/lib.rs
sunfishcode: Makes sense, thanks.

sunfishcode:
I tested this by copying the bindings generated with this patch into https://github.com/bytecodealliance/wasmtime/pull/7029 and it fixes the bug I was seeing.

PR #675, opened 2023-09-22
Look through aliases when creating import params

rylev:
When creating params for imported functions, look through aliases so that they are treated exactly like their aliased type. 

This prevents aliased primitive types from being borrowed more than they need to be.

### DIFF
--- a/crates/rust-lib/src/lib.rs
+++ b/crates/rust-lib/src/lib.rs
@@ -304,6 +304,21 @@ pub trait RustGenerator<'a> {
         let info = self.info(id);
         let lt = self.lifetime_for(&info, mode);
         let ty = &self.resolve().types[id];
+
+        // Look through type aliases
+        let mut potential_alias = ty;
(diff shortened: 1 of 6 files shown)


rylev:
I believe the CI failure is not related to this PR. It seems that the last run on main also has the same error.

alexcrichton:
Thanks! Something about this feels a bit odd to me because the logic here seems like it's written to generally peel back aliases but at the same time it only works for primitives. While there's not really any harm in duplicating primitives definitions it does lose the original structure of the WIT. 

Would it be possible though to fix this via a different means? I think the original issue you're looking to fix is that `type x = string` means that `import foo(x: x)` takes a `&&str` parameter, right? (due to taking `&X`). If that's the case it may also be possible to fix by putting extra logic on the "print the leading `&`" bit of code

rylev:
I've not been able to find a solution for this that I'm fully happy with so closing for now.

PR #676, opened 2023-09-22
Rust: Don't borrow borrowed types

### LINKED ISSUE #673: wit-bindgen 0.12 Rust changes how explicitly borrowed param types are borrowed
Assume the following wit:

```wit
type my-param = string	
my-func: func(param: my-param)
```

If this is code-gened in wit-bindgen Rust 0.11 using `duplicate_if_necessary: true` you get the following signature:

```rust
fn my_func(param: MyParam<'_>)
```

When upgrading to 0.13 however, you get the following:

```rust
fn my_func(param: &MyParam<'_>)
```

Essentially the borrowed param is double borrowed. I would expect that this would not change. [This recent change](https://github.com/bytecodealliance/wit-bindgen/commit/f6004cd10bc7249439da20d9d56deb3808aac86b#diff-d91d041cd6a7e3e16c405fe2fbc3ab6ee6ead9972338e5ac3ebcbd7001741ca1L308) seems to be the culprit.

rylev:
Fixes #673

This partially reverts a change now live in v0.12 that was introduced in #669. If a type has a lifetime associated with it, treat it as already borrowed and don't borrow again.

### DIFF
--- a/crates/rust/src/interface.rs
+++ b/crates/rust/src/interface.rs
@@ -670,10 +670,13 @@ impl InterfaceGenerator<'_> {
             // it is actually required that we take ownership since Rust is
             // losing access to those handles.
             //
+            // We also skip borrowing if the type has a lifetime associated with
+            // in which case we treated it as already borrowed.
+            //
             // Check here if the type has the right shape and if we're in the
             // right mode, and if those conditions are met a lifetime is
             // printed.
-            if info.has_list && !info.has_own_handle {
+            if info.has_list && !info.has_own_handle && lt.is_none() {
                 if let TypeMode::AllBorrowed(lt) | TypeMode::HandlesBorrowed(lt) = mode {
                     self.push_str("&");
                     if lt != "'_" {
--- a/tests/runtime/ownership/borrowing-duplicate-if-necessary.rs
+++ b/tests/runtime/ownership/borrowing-duplicate-if-necessary.rs
@@ -24,7 +24,7 @@ impl Guest for Exports {
             lists::foo(value)
         );
 
-        thing_in::bar(&thing_in::Thing {
+        thing_in::bar(thing_in::Thing {
             name: "thing 1",
             value: &["some value", "another value"],
         });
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thinking through this a bit now, one consequence of this is that the generation of this:

```wit
package foo:bar

world x {
  resource z

  record x {
    y: list<u32>,
    z: borrow<z>,
  }

  import the-fn: func(x: x)
}
```

changes. Before this PR this generates:

```rust
pub struct X<'a> {
    pub y: wit_bindgen::rt::vec::Vec<u32>,
    pub z: &'a Z,
}

pub fn the_fn(x: &X<'_>) { ... }
```

whereas after this PR it generates

```rust
pub struct X<'a> {
    pub y: wit_bindgen::rt::vec::Vec<u32>,
    pub z: &'a Z,
}

pub fn the_fn(x: X<'_>) { ... }
```

that means that this PR requires ownership of the `Vec` field whereas previously it wasn't required, and I think this was one thing I was trying to fix implicitly in #673 as well.

rylev:
@alexcrichton sorry for the delay in getting back to you.

[I've updated the code I'm working on from wit-bindgen 11 to 12 ](https://github.com/fermyon/spin/commit/56666be796bf888a333b8d0e45f20d3684d6f0bb)so we can see what the code ends up looking like. I think it's hard to argue that the generated code here is more idiomatic. Any thoughts on how we might keep the implicitness you desire while reducing the amount of double/triple borrows seen in the code I've linked to?

alexcrichton:
Well hey if we've already got double and triple borrows may as well go for the quadruple and legendary quintuple borrow while we're at it!

But no my main concern here is the runtime implications of the generated signatures as opposed to feeling that the current bindings are idiomatic (those clearly aren't). I think we should probably go back to what was there prior, like with this PR, but there probably needs to be affordances for resources to handle the case above I mention. I'm a bit frustrated with how I can't seem to ever settle on a strategy of types generated here (it feels like constant whack-a-mole), but such is the experience of developing a bindings generator!

If you've run out of time to work on this I can also poke around and see if I can't restore the bindings from before too

rylev:
@alexcrichton I've rebased - I believe this is what we've agreed upon (i.e., the original changes). Perhaps we can revisit this with more targeted improvements in the future?

alexcrichton:
Sounds good!

PR #677, opened 2023-09-25
Fix the `exports` key for packages-with-versions

alexcrichton:
Previously one location looking up `exports` would account for the version and another wouldn't which caused the `ERROR` case to leak through by accident. These are now kept in sync to ensure that the right errors get surfaced.

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -296,24 +296,14 @@ impl WorldGenerator for RustWasm {
         id: InterfaceId,
         _files: &mut Files,
     ) -> Result<()> {
-        let (pkg, inner_name) = match name {
-            WorldKey::Name(name) => (None, name),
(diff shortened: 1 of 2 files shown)


dicej:
Noticing that the package name is no longer part of the key; would this cause an issue when exporting two interfaces with the same name but different packages, e.g. `foo:bar/x` and `some:other/x`?

alexcrichton:
Oh the package name should still be there, that's a mistake if I left it out by accident. My intention was that previously you'd specify `a:b/c/resource` but now you have to specify `a:b/c@0.0.0/resource` instead

dicej:
Ok, I see now.  It's `Resolve::id_of` that's doing the work of providing a fully-qualified identifier now.  Thanks for clarifying.

dicej (approved):

PR #678, opened 2023-09-25
feat(rust): Adds the ability to specify custom derive attributes

### LINKED ISSUE #554: Add support for Serialize, Deserialize, and Default
The structures that we get from Rust when doing wit-bindgen do not have the `Default` or `Serialize` or `Deserialize` derivations on them. This means that as soon as I attempt to marshal one of these data types somewhere without using a wit function call, I have to manually create my own "mirror types" that have the same structure, but with the serialization attributes on them. Then I have to manually implement either `From` to convert the wit-bindgen type to my type prior to serialization. 

I understand that this kind of activity typically shows up in anti-corruption layers and in an ideal world, the "outside" model and the inside (wit-bindgen) model would never overlap. However, if I have a _lot_ of types and I need to marshal those in and out of my components without making bindgen function calls, I'm stuck redefining every type in my wit world.

Can we get this functionality? If it isn't something that you want on by default, it could be added as a flag to the macro like this:

```rust
wit_bindgen::generate!({
    path: "../wit",
    world: "bankaccount-aggregate",
    serializers: true
});
```

Support for the `serializers` flag could even be hidden behind a feature flag on the wit-bindgen crate.

thomastaylor312:
I also did a little bit of cleanup of clippy warnings while I was in these files.

Closes #554

### DIFF
--- a/crates/rust-lib/src/lib.rs
+++ b/crates/rust-lib/src/lib.rs
@@ -1,4 +1,5 @@
 use heck::*;
+use std::collections::BTreeSet;
 use std::fmt;
 use std::str::FromStr;
 use wit_bindgen_core::abi::{Bitcast, LiftLower, WasmType};
@@ -91,6 +92,9 @@ pub trait RustGenerator<'a> {
 
     fn is_exported_resource(&self, ty: TypeId) -> bool;
 
+    /// Return any additional derive attributes to add to the generated types
+    fn additional_derives(&self) -> &[String];
+
     fn mark_resource_owned(&mut self, resource: TypeId);
 
     fn push_str(&mut self, s: &str);
@@ -159,7 +163,7 @@ pub trait RustGenerator<'a> {
         param_mode: TypeMode,
         sig: &FnSig,
     ) -> Vec<String> {
-        let params = self.print_docs_and_params(func, param_mode, &sig);
+        let params = self.print_docs_and_params(func, param_mode, sig);
         if let FunctionKind::Constructor(_) = &func.kind {
             self.push_str(" -> Self")
         } else {
@@ -198,7 +202,7 @@ pub trait RustGenerator<'a> {
         } else {
             &func.name
         };
-        self.push_str(&to_rust_ident(&func_name));
+        self.push_str(&to_rust_ident(func_name));
         if let Some(generics) = &sig.generics {
             self.push_str(generics);
         }
@@ -563,7 +567,7 @@ pub trait RustGenerator<'a> {
         if self.uses_two_names(&info) {
             result.push((self.param_name(ty), TypeMode::AllBorrowed("'a")));
         }
-        return result;
+        result
     }
 
     fn print_typedef_record(
@@ -574,6 +578,9 @@ pub trait RustGenerator<'a> {
         derive_component: bool,
     ) {
         let info = self.info(id);
+        // We use a BTree set to make sure we don't have any duplicates and we have a stable order
+        let additional_derives: BTreeSet<String> =
+            self.additional_derives().iter().cloned().collect();
         for (name, mode) in self.modes_of(id) {
             let lt = self.lifetime_for(&info, mode);
             self.rustdoc(docs);
@@ -586,12 +593,17 @@ pub trait RustGenerator<'a> {
                 self.push_str("#[derive(wasmtime::component::Lower)]\n");
                 self.push_str("#[component(record)]\n");
             }
-
+            let mut derives = additional_derives.clone();
             if info.is_copy() {
                 self.push_str("#[repr(C)]\n");
-                self.push_str("#[derive(Copy, Clone)]\n");
+                derives.extend(["Copy", "Clone"].into_iter().map(|s| s.to_string()));
             } else if info.is_clone() {
-                self.push_str("#[derive(Clone)]\n");
+                derives.insert("Clone".to_string());
+            }
+            if !derives.is_empty() {
+                self.push_str("#[derive(");
+                self.push_str(&derives.into_iter().collect::<Vec<_>>().join(", "));
+                self.push_str(")]\n")
             }
             self.push_str(&format!("pub struct {}", name));
             self.print_generics(lt);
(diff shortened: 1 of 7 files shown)


thomastaylor312:
I am pretty sure the test failure is a flake or not related to my code as I am seeing it in other PRs as well

crates/rust-lib/src/lib.rs
```diff
@@ -159,7 +160,7 @@ pub trait RustGenerator<'a> {
         param_mode: TypeMode,
         sig: &FnSig,
     ) -> Vec<String> {
-        let params = self.print_docs_and_params(func, param_mode, &sig);
+        let params = self.print_docs_and_params(func, param_mode, sig);
```
alexcrichton, line 163: No worries for this PR, but for future contributions I'd recommend splitting out clippy changes into a separate commit or possibly PR. Makes the "meat" elsewhere a bit easier to review that way.

crates/rust-lib/src/lib.rs
```diff
@@ -572,8 +573,11 @@ pub trait RustGenerator<'a> {
         record: &Record,
         docs: &Docs,
         derive_component: bool,
+        additional_derives: Vec<String>,
```
alexcrichton, line 576: It's a bit unfortunate to pass this around in so many places. In theory this `rust-lib` crate shouldn't even exist any longer. This was only split out as it was shared between Wasmtime and a Rust guest, but Wasmtime no longer lives here nor uses this so it's an unnecessary abstraction nowadays. Additionally that's a large enough refactor there's no need to do it in this PR.

In the meantime though a possible alternative design would be a trait method on `RustGenerator` along the lines of `fn additional_derives(&self) -> &[String]`. I think that should avoid most of the passing around and clean that up slightly. Mind taking this on and I'll as a follow up work on merging this crate into the `rust` crate?

crates/rust-lib/src/lib.rs
```diff
@@ -586,12 +590,20 @@ pub trait RustGenerator<'a> {
...
+                derives.extend(["Copy", "Clone"].into_iter().map(|s| s.to_string()));
             } else if info.is_clone() {
-                self.push_str("#[derive(Clone)]\n");
+                derives.insert("Clone".to_string());
+            }
+            if !derives.is_empty() {
+                self.push_str("#[derive(");
+                // NOTE(thomastaylor312): Since this is code generation, I don't mind the iteration
+                // and allocation of a new vec here. If it ever matters, we can create a function
+                // that glues it together via iterator
```
alexcrichton, line 604: No worries on leaving a note here, very little of wit-bindgen is optimized for performance :)

crates/rust-lib/src/lib.rs
```diff
@@ -572,8 +573,11 @@ pub trait RustGenerator<'a> {
         record: &Record,
         docs: &Docs,
         derive_component: bool,
+        additional_derives: Vec<String>,
     ) {
         let info = self.info(id);
+        // We use a hash set to make sure we don't have any duplicates
+        let additional_derives: HashSet<String> = additional_derives.into_iter().collect();
```
alexcrichton, line 580: One important caveat to be aware of about `HashSet<T>` is that its iteration order is not defined and not deterministic between runs of a process. This means, for example, that different code is generated from the same WIT source as derives might be shuffled around.

Mind using a `BTreeSet` (or `IndexSet`) here instead to preserve order?

crates/rust-lib/src/lib.rs
```diff
@@ -853,6 +878,7 @@ pub trait RustGenerator<'a> {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
```
alexcrichton, line 881: This project doesn't use clippy, so mind leaving this out?

alexcrichton (approved):
Thanks for this! No worries on the failed test, that's [known to be flaky](https://github.com/bytecodealliance/wit-bindgen/pull/680) and should be resolved on `main` soon through ignoring it.

At a high level one concern I have about this is that it won't work for `Resource` types. That won't implement Serialize/Deserialize and doesn't currently implement PartialEq/Hash, although it perhaps could. This means that implementing traits uniformly for all types may not be sufficient and eventually this may need a form of "ignore this type" or "only derive for this type". That all being said I think it's fine to deal with those situations as they come up later, for now this is suitable and works well so I'd say we should land it and then iterate on further syntaxes later if necessary.

crates/rust-lib/src/lib.rs
thomastaylor312: Yep, I'll shift it around. Honestly was having a hard time keeping track of where everything was at during the beginning. Now that I have a better grasp, I can do this quick

crates/rust-lib/src/lib.rs
thomastaylor312: Oh right, I should have though about the ordering issue. I'll switch that around

crates/rust-lib/src/lib.rs
thomastaylor312: Yeah I can leave it out

thomastaylor312:
@alexcrichton Should be ready to go for a rereview

In regards to this:

> At a high level one concern I have about this is that it won't work for Resource types. That won't implement Serialize/Deserialize and doesn't currently implement PartialEq/Hash, although it perhaps could. This means that implementing traits uniformly for all types may not be sufficient and eventually this may need a form of "ignore this type" or "only derive for this type". That all being said I think it's fine to deal with those situations as they come up later, for now this is suitable and works well so I'd say we should land it and then iterate on further syntaxes later if necessary.

I completely agree with the path forward here. I was curious how this would interact with resources, but it definitely seemed easiest to add this to struct/enum types for now and then we can see if there is a way to make it more specific in the future

crates/rust-lib/src/lib.rs
thomastaylor312: This should be all resolved, but leaving the comment unresolved until you have a chance to take a look at it

alexcrichton:
Looks great to me, thanks!

One option is to use the `TypeInfo` bits to avoid these extra derives for types-with-resources, but that's still a bit specific and may not work out in all cases. In any case this looks good for now 👍

sehz:
does this work work in `cargo component` ?

thomastaylor312:
@sehz I was actually going to go open an issue over there in cargo component

PR #689, opened 2023-10-05
Add resource-own-in-other-interface test and fix C

dicej:
This tests a case of aliasing a resource across interfaces which we hadn't covered before, and fixes the C generator to handle it.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -1039,8 +1039,7 @@ impl C {
                                 }
                                 _ => {}
                             }
-                            self.public_anonymous_types.insert(*id);
-                            self.private_anonymous_types.remove(id);
+                            self.visit_anonymous_types(resolve, &Type::Id(*id));
                             dst.push_str(&ns);
                             dst.push_str("_");
                             push_ty_name(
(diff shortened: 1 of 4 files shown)


dicej:
I'll be honest and say I'm kind of playing bop-a-mole with these test cases, as I have not yet thought deeply about how aliases and resources should interact in the C generator.  There are probably other broken corner cases lurking.

crates/c/src/lib.rs
```diff
@@ -650,22 +650,21 @@ impl C {
...
         loop {
-            if let TypeDefKind::Type(Type::Id(id)) = &resolve.types[*target].kind {
+            let def = &resolve.types[target];
+            if let TypeDefKind::Type(Type::Id(id)) = &def.kind {
                 let target_handle = match handle {
                     Handle::Borrow(_) => Handle::Borrow(*id),
                     Handle::Own(_) => Handle::Own(*id),
                 };
-                if let Some(ty) = handles.get(&target_handle) {
+                if let (true, Some(ty)) = (def.owner == owner, handles.get(&target_handle)) {
```
dicej, line 664: The basic idea here is that we shouldn't generate `typedef` aliases for handles that alias across interfaces, since the target interface might not have been imported at all.  We might want to expand this to check if the target interface _is_ imported and do something different.  Or maybe take a step back and consider a broader re-design.

dicej: Actually, I just reverted this part and the test still passes, so nevermind :).  I'll push an update.

alexcrichton (approved):
Oof. Sounds good for now, and I'll see if I can explore a better fix for C in the future which hopefully feels less whack-a-mole, but honestly that's largely what wit-bindgen feels like most of the time...

PR #691, opened 2023-10-06
test(C): test all generated C header files for C++ compatibility, too

tschneidereit:
This can't land yet, because it causes a bunch of tests to fail. One will be addressed by #690, but the others all involve empty types, and will probably require changes to the tests themselves.


JakeChampion:
Here is another C++ incompatibility I've come across -- https://github.com/bytecodealliance/wit-bindgen/issues/725

PR #693, opened 2023-10-09
Multi version support 

rylev:
A continuation of https://github.com/bytecodealliance/wasmtime/pull/7172 and https://github.com/bytecodealliance/wasmtime/pull/7196 and borrows liberally from those PRs. This allows the use of different versions of the same logical interface. 

This only fixes Rust so I'm marking it as a draft until tomorrow when I can fix the other languages.


alexcrichton (approved):
Thanks!

alexcrichton:
I see that at least the C test is failing, and it's ok to add ignores to other generators for this new test if you'd like

rylev:
@alexcrichton I added ignores to C, Go, and Java, and improved the test a bit to catch one additional issue I wasn't yet handling. This should be ready to merge. I can try to look into the other languages as well.

PR #695, opened 2023-10-10
Support multiple versions of the same interface in C

rylev:
This is a continuation of #693 adding multi-version support for C.


alexcrichton:
Thanks!

PR #699, opened 2023-10-12
feat(rust): implement `with` macro argument

### LINKED ISSUE #694: Rust: Support `with` option for `bindgen!` macro.
This is already supported in `wasmtime`'s generate macro.

rvolosatovs:
Closes #694 

Introduce `with` Rust macro argument supported by Wasmtime bindgen.
Big part of this is adapted from Wasmtime source code.
The implementation is very simple and has minimal changes to existing functionality - when `with` is specified and matching an interface name, submodule generation is completely skipped for that interface and all occurrences use the remapped path directly.

I have tested this locally by remapping the `wasi:io/streams` import. I am honestly not sure if the export logic is valid - Wasmtime bindgen handles the two cases differently.

### DIFF
--- a/crates/rust-macro/src/lib.rs
+++ b/crates/rust-macro/src/lib.rs
@@ -1,5 +1,6 @@
 use proc_macro2::{Span, TokenStream};
 use quote::ToTokens;
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use syn::parse::{Error, Parse, ParseStream, Result};
 use syn::punctuated::Punctuated;
@@ -79,6 +80,7 @@ impl Parse for Config {
                             .map(|p| p.into_token_stream().to_string())
                             .collect()
                     }
+                    Opt::With(with) => opts.with.extend(with),
                 }
             }
         } else {
@@ -166,6 +168,7 @@ mod kw {
     syn::custom_keyword!(stubs);
     syn::custom_keyword!(export_prefix);
     syn::custom_keyword!(additional_derives);
+    syn::custom_keyword!(with);
 }
 
 #[derive(Clone)]
@@ -225,6 +228,7 @@ enum Opt {
     ExportPrefix(syn::LitStr),
     // Parse as paths so we can take the concrete types/macro names rather than raw strings
     AdditionalDerives(Vec<syn::Path>),
+    With(HashMap<String, String>),
 }
 
 impl Parse for Opt {
@@ -322,6 +326,14 @@ impl Parse for Opt {
             syn::bracketed!(contents in input);
             let list = Punctuated::<_, Token![,]>::parse_terminated(&contents)?;
             Ok(Opt::AdditionalDerives(list.iter().cloned().collect()))
+        } else if l.peek(kw::with) {
+            input.parse::<kw::with>()?;
+            input.parse::<Token![:]>()?;
+            let contents;
+            let _lbrace = braced!(contents in input);
+            let fields: Punctuated<_, Token![,]> =
+                contents.parse_terminated(with_field_parse, Token![,])?;
+            Ok(Opt::With(HashMap::from_iter(fields.into_iter())))
         } else {
             Err(l.error())
         }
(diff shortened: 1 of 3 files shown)


fibonacci1729:
With this PR we are simply using the remapped name in place of the rust path to the module generated for the interface.  However i'm not sure this is going to work quite as expected, for example, given the WIT:

```
package component:example

interface api {
    record message {
        text: string
    }
}

interface xyz {
    use api.{message}
    f: func(m: message)
}

world guest {
    export xyz
}
```

Using the remapping of `component:example/api` to `app:api`, this currently generates (omitting most cruft for brevity):

```rust
pub mod component {
  pub mod example {
    pub mod xyz {
      // ...
      pub type Message = app::api::Message;
      // ...
    }
  }
}
```

However, in this particular example `app::api` refers to a module at the level of the macro invocation, I would expect the path in the generated code to be `super::super::super::app::api::Message`.

As far as I can tell, `wasmtime` anchors remappings to the root of the generated code and adds an alias using the `with_name_counter`, e.g using the above example:

```rust
use app::api as __with_name_0;

pub mod component {
  pub mod example {
    pub mod xyz {
      // ...
      pub type Message = super::super::__with_name_0::Message;
      // ...
    }
  }
}
```

As far as I can tell the point of doing it this way is 1) to keep the paths a little leaner and 2) to allow the dev using `with` in the macro call to avoid having to understand the internal generated structure and having to qualify the remap paths with the necessary `super`'s, i.e.:

```
with: { 
  "component:example/api": super::super::super::app::api 
}
```

Let me know if this make sense to you @rvolosatovs or if i'm misunderstanding some assumptions around using the macro.

rvolosatovs:
> With this PR we are simply using the remapped name in place of the rust path to the module generated for the interface.  However i'm not sure this is going to work quite as expected, for example, given the WIT:
> 
> ```
> package component:example
> 
> interface api {
>     record message {
>         text: string
>     }
> }
> 
> interface xyz {
>     use api.{message}
>     f: func(m: message)
> }
> 
> world guest {
>     export xyz
> }
> ```
> 
> Using the remapping of `component:example/api` to `app:api`, this currently generates (omitting most cruft for brevity):
> 
> ```rust
> pub mod component {
>   pub mod example {
>     pub mod xyz {
>       // ...
>       pub type Message = app::api::Message;
>       // ...
>     }
>   }
> }
> ```
> 
> However, in this particular example `app::api` refers to a module at the level of the macro invocation, I would expect the path in the generated code to be `super::super::super::app::api::Message`.
> 
> As far as I can tell, `wasmtime` anchors remappings to the root of the generated code and adds an alias using the `with_name_counter`, e.g using the above example:
> 
> ```rust
> use app::api as __with_name_0;
> 
> pub mod component {
>   pub mod example {
>     pub mod xyz {
>       // ...
>       pub type Message = super::super::__with_name_0::Message;
>       // ...
>     }
>   }
> }
> ```
> 
> As far as I can tell the point of doing it this way is 1) to keep the paths a little leaner and 2) to allow the dev using `with` in the macro call to avoid having to understand the internal generated structure and having to qualify the remap paths with the necessary `super`'s, i.e.:
> 
> ```
> with: { 
>   "component:example/api": super::super::super::app::api 
> }
> ```
> 
> Let me know if this make sense to you @rvolosatovs or if i'm misunderstanding some assumptions around using the macro.
> 

That makes sense, in fact originally I ported the feature with the counter and the alias, however there is one key difference here - Wasmtime bindgen generates the `__with...` aliases at the root, whereas in this crate we get the alias generated in the submodule.

Since this fulfilled my use case I decided to omit that part for MVP, since making it work with the relative paths is quite a bit trickier and would result in a much bigger change.

For now this essentially assumes absolute paths.

Since you were originally planning to open a PR anyway, would you be interested in contributing to this PR?

fibonacci1729:
I think what you have is great as an incremental addition and I believe that this addresses my use case as well. Could you perhaps leave a comment/TODO mentioning that `with` currently only supports absolute paths (we can land that later so as to not block progress elsewhere by trying to solve for relative paths pre-emptively)?

Otherwise this LGTM but i'll defer to @alexcrichton for final approval.

fibonacci1729 (approved):

alexcrichton:
Thanks! Agreed the implementation looks ok, but I also agree with @fibonacci1729 that this issue should be fixed with relative-vs-not paths.

I'd personally prefer to not land this as an interim state because it runs a risk of causing implementations to use the module-relative paths right now where in the future that code would break, likely confusingly, when the implementation switched to be relative to the root.

Would @rvolosatovs or @fibonacci1729 one of y'all be up for implementing the relative-to-the-macro-invocation logic?

fibonacci1729:
PR: https://github.com/rvolosatovs/wit-bindgen/pull/1

alexcrichton:
Looks good to me! 

There are some test failures though I think. Additionally this should have some tests, but I can write those up later today too.

rvolosatovs:
@fibonacci1729 I've rebased to get rid of the merge commit and amended your commit with a `cargo fmt --all`. Also added you as collaborator on my fork, I believe you should be able to push to the branch directly now if necessary

alexcrichton (approved):

PR #704, opened 2023-10-18
Add Wasi to runtime tests for csharp

yowl:
This PR add WASI to the runtime tests in preparation for adding c# tests.  Also adds the c# feature code to hopefully minimize the changes required when the tests are actually added.  WIll have to switch to the reactor adaptor as well, but that hopefully will be a small change.


alexcrichton:
Thanks! And yeah now that WASI exists it should be ok to remove the testwasi crate since it's no longer needed.

tests/runtime/main.rs
```diff
@@ -453,5 +489,225 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
...
+        <configuration>
+            <config>
+                <add key="globalPackagesFolder" value=".packages" />
+            </config>
+            <packageSources>
+            <!--To inherit the global NuGet package sources remove the <clear/> line below -->
+            <clear />
+            <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+            <!--<add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />-->
+            <add key="dotnet-experimental" value="C:\github\runtimelab\artifacts\packages\Debug\Shipping" />
```
jsturtevant, line 538: this looks like it is the project for the llvm approach?  Since we want to cover both will we need to create a sperate solution file for the mono?

tests/runtime/main.rs
```diff
@@ -97,13 +130,15 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
...
     let mut go = Vec::new();
+    let mut c_sharp = Vec::new();
     for file in dir.read_dir()? {
         let path = file?.path();
         match path.extension().and_then(|s| s.to_str()) {
             Some("c") => c.push(path),
             Some("java") => java.push(path),
             Some("rs") => rust.push(path),
             Some("go") => go.push(path),
+            Some("cs") => c_sharp.push(path),
```
jsturtevant, line 141: how should we distingiush between cs llvm vs mono should that be here or a feature flag?

yowl:
> Thanks! And yeah now that WASI exists it should be ok to remove the testwasi crate since it's no longer needed.

So would I delete https://github.com/bytecodealliance/wit-bindgen/blob/a868860fb4ac4ccf5b5a41f5ff9752d9c150dc99/crates/test-rust-wasm/artifacts/build.rs#L10-L19 and add the reactor wasm to the repo ?

tests/runtime/main.rs
yowl: Good point, yes, we will need a different project file for mono, so I guess that would mean, as far as running the tests, that we should have a `csharp-naot` feature and a `csharp-mono` feature.

tests/runtime/main.rs
yowl: I think just one `cs` but 2 features.

yowl:
@alexcrichton Shall I wait for wasmtme 14 or let this one be reviewed?

alexcrichton (approved):
Nah seems reasonable to me to go ahead and land!

jsturtevant:
still learning here but LGTM

PR #711, opened 2023-10-23
Add C# feature

yowl:
This PR adds the C# feature for the codegen (not complete).  With this it is possible to run 
```
cargo run c-sharp  --string-encoding utf8 --out-dir testing-csharp tests/codegen/floats.wit
```

Only one feature is needed here, `csharp`, because there is no compilation, just c# code gen.

Runtime tests are not enabled yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,7 @@ jobs:
     - run: cargo build --no-default-features --features c
     - run: cargo build --no-default-features --features teavm-java
     - run: cargo build --no-default-features --features go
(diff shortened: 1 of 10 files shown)


yowl:
Hi, requested a review on this one as it touches non -c# files.   Thanks

alexcrichton (approved):
Integration here all looks good to me. I know next-to-nothing about C# myself though so I'm probably not the best to review the guts of `lib.rs` of the new crate. Happy to have this merged regardless though if you feel it's ready!

yowl:
Thanks, I'll go ahead as its a blocker for other contributors.

PR #715, opened 2023-10-25
Resource runtime test

Mossaka:
This commit adds a resources runtime test for rust as this repo currently contains zero resources tests. 

CC @dicej

### DIFF
--- a/tests/runtime/resources/wasm.rs
+++ b/tests/runtime/resources/wasm.rs
@@ -65,7 +65,7 @@ impl exports::exports::GuestX for ComponentX {
         *self.val.borrow_mut() = a;
     }
     fn add(x: OwnX, a: i32) -> OwnX {
-        // not working
+        x.set_a(x.get_a() + a);
         x
     }
 }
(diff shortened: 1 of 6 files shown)


tests/runtime/resources/wasm.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+    fn get_a(&self) -> i32 {
+        *self.val.borrow()
+    }
+    fn set_a(&self, a: i32) {
+        *self.val.borrow_mut() = a;
+    }
+    fn add(x: OwnX, a: i32) -> OwnX {
+        // not working
```
dicej, line 68: Why isn't this working?  What kind of error are you getting?

dicej (approved):
The test looks reasonable to me.  Per CI, looks like some code will need to be updated to work with Wasmtime 14 and its WASI implementation.

tests/runtime/resources/wasm.rs
Mossaka: It turns out the issue I encountered was just a lacking of understanding of the `OwnX` type. Now the implementaiton is added.

alexcrichton (approved):

PR #718, opened 2023-10-27
C# enable the runtime numbers test

yowl:
This PR enables the first runtime test for c#.
It compiles using dotnet 8 and NativeAOT-LLVM which unfortunately requires clang from emscripten as well as the WASI SDK (see https://github.com/WebAssembly/wasi-sdk/issues/326).  So we download and install emscripten as well as dotnet 8.  Tha NativeAOT-LLVM compiler is only supported on windows so the yml has the windows build separated out.

Dotnet 8 is required as we are using `static abstract` interface methods.  This shouldn't be a major issue as dotnet 8 is coming out in a few weeks.

Also moved some of the casting to the `Instruction` `emit` cases.

Added 2 new features, `csharp-naot` and `csharp-mono` which bring in `csharp`.  `csharp-mono` has no implementation here for the runtime test.


.github/workflows/main.yml
```diff
@@ -47,6 +47,22 @@ jobs:
         echo "WASI_SDK_PATH=`pwd`/wasi-sdk-16.0" >> $GITHUB_ENV
       if : matrix.os == 'windows-latest'
 
+    - run: |
+        curl -O https://dotnet.microsoft.com/download/dotnet/scripts/v1/dotnet-install.ps1
```
jsturtevant, line 51:
```suggestion
        curl.exe -LO https://dotnet.microsoft.com/download/dotnet/scripts/v1/dotnet-install.ps1
```

tests/runtime/main.rs
```diff
@@ -115,15 +115,19 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
...
         match path.extension().and_then(|s| s.to_str()) {
             Some("c") => c.push(path),
             Some("java") => java.push(path),
             Some("rs") => rust.push(path),
             Some("go") => go.push(path),
-            Some("cs") => c_sharp.push(path),
+            Some("cs") => {
+                if cfg!(windows) {
+                    c_sharp.push(path);
+                }
```
jsturtevant, line 129: Question: When mono is added we will want to still add this for the linux case?  Since we aren't running the tests on linux right now, is this needed?

.github/workflows/main.yml
```diff
@@ -47,6 +47,22 @@ jobs:
...
+        powershell -File dotnet-install.ps1 -Channel 8.0.1xx -Verbose
+        echo DOTNET_ROOT=$LOCALAPPDATA'\Microsoft\dotnet' >> $GITHUB_ENV
+        export DOTNET_ROOT=$LOCALAPPDATA\\Microsoft\\dotnet
+        echo $LOCALAPPDATA'\Microsoft\dotnet' >> $GITHUB_PATH
+        echo $LOCALAPPDATA'\Microsoft\dotnet\tools' >> $GITHUB_PATH
+        $LOCALAPPDATA/Microsoft/dotnet/dotnet --info
+        echo nativeaot-llvm requires emscripten for its version of clang as wasi-sdk 20 does not work see https://github.com/WebAssembly/wasi-sdk/issues/326
+        curl -OL https://github.com/emscripten-core/emsdk/archive/refs/heads/main.zip
```
jsturtevant, line 59:
```suggestion
        curl.exe -OL https://github.com/emscripten-core/emsdk/archive/refs/heads/main.zip
```

tests/runtime/numbers/wasm.cs
```diff
@@ -0,0 +1,133 @@
...
+    public static void SetScalar(uint p0)
+    {
+        SCALAR = p0;
+    }
+
+    public static uint GetScalar()
+    {
+        return SCALAR;
+    }
+}
```
jsturtevant, line 133: nit: newline

tests/runtime/main.rs
yowl: The problem is the yml matrix in `build.yml` tries to run all the OSes and if it creates the c# files then it will fail to compile them.  Agree we will have to revisit this for Mono, not sure about MacOS, but for Linux we will want it.

tests/runtime/numbers/wasm.cs
yowl: Thanks

Issue #641, opened 2023-08-10
Add ability to consume an owned resource?


juntyr:
If I implement the export of a resource in Rust, and there is a method which takes an owned handle to this resource, the method implementation gets an `Own{my-resource-name}` passed by value. At the moment, this owned handle type only implements `Deref`, but not `DerefMut` or a helper function that consumes the resource handle and returns the underlying Rust type (the one we can pass to the `Own{my-resource-name}::new` method to construct the resource in the first place).

Even though the handle is owned and gives unique ownership, is there something in WIT or the component model that prevents the implementation of `DerefMut` or a consume method?

At the moment, I need to work around this deficiency by having the resource hold a `RefCell<Option<MyConsumableType>>`, which allows mutable access and consumption but is not infallible.

Thank you for your help! If these methods could be implemented, I'll gladly write the PR for them!

alexcrichton:
Thanks for the report, and apologies it took me a bit to get here! I believe it should be safe to implement `DerefMut` as you've mentioned, and a PR would be most welcome! Are you still up for sending a PR? If not no worries and I can prepare one.

juntyr:
No worries - and yes, I’ll work on one later today :)

juntyr:
@alexcrichton Could the component model also gain a method called, e.g. `[resource-consume]{name}`, which invalidates the resource handle (similar to drop) but does not drop the resource implementation data? That way, the `Own{name}` could get a `consume` function with an implementation similar to this (adapted from the `new` function template, i.e. meant to be inside a format string):

```rust
pub fn consume(own: Own{camel}) -> Rep{camel} {
    #[allow(unused_imports)]
    use {rt}::boxed::Box;

    unsafe {{
        #[cfg(target_arch = "wasm32")]
        #[link(wasm_import_module = "[export]{module}")]
        extern "C" {{
            #[link_name = "[resource-consume]{name}"]
            fn wit_import(_: i32) -> i32;
        }}

        #[cfg(not(target_arch = "wasm32"))]
        unsafe fn wit_import(_n: i32) -> i32 {{ unreachable!() }}

        // [resource-consume]{name} returns value address,
        //  like [resource-rep]{name}, but also invalidates the handle
        let rep_ptr = ::core::mem::transmute::<isize, *mut Rep{camel}>(
            wit_import(self.handle).try_into().unwrap()
        );

        *Box::from_raw(rep_ptr)
    }}
}}
```

alexcrichton:
[...]

PR #719, opened 2023-10-29
More rust runtime tests for resources

Mossaka:
This PR adds more rust runtime tests for Resources. These WITs and tests are moved from https://github.com/bytecodealliance/componentize-js/pull/65 and translated into Rust. 

A few things that I found difficult to do in Rust compared to Js or Python are
1. Consuming owned resource in the guest code. This is also described in #641. I had to use `Option::take` to consume the owned type underneath the Resource. 
2. Consuming `Resource<T>` in the host code.
3. Generated modules tend to have long import names. E.g. `use exports::test::resource_borrow_in_record::test::{Guest, GuestThing, OwnThing};` 
4. Question: How do I check if two `ResourceAny` are the same? 

Thanks @dicej , @alexcrichton for help!


Mossaka:
> Does == work?

What I want to test is a deep equal of two `ResourceAny`s

```rust
let res = instance
      .test_resource_alias_e1()
      .call_a(&mut *store, foo_e1)?;

// TODO: how do I test deep equal of ResourceAny type?
assert_eq!(
    res,
    vec![instance
        .test_resource_alias_e1()
        .x()
        .call_constructor(&mut *store, 42)?]
);

```

What I got is

```bash
thread 'resource_alias::run' panicked at 'assertion failed: `(left == right)`
  left: `[ResourceAny { idx: 0, ty: ResourceType { kind: Guest { store: StoreId(1), instance: 5062857728, id: DefinedResourceIndex(0) } }, own_state: Some(OwnState { store: StoreId(1) }) }]`,
 right: `[ResourceAny { idx: 1, ty: ResourceType { kind: Guest { store: StoreId(1), instance: 5062857728, id: DefinedResourceIndex(0) } }, own_state: Some(OwnState { store: StoreId(1) }) }]`'
```

Does it make sense to define `PartialEq` of `ResourceAny` by comparing their `own_state` equality?

alexcrichton:
Can you gist the guest and the WITs involved?

PR #728, opened 2023-11-02
Move wasmtime-wasi to dev-dependencies

yamt:
While it's using the majority of build time on my environment, it seems only used for tests.


yowl:
Thanks, looks fine for me.

alexcrichton (approved):
Thanks!

PR #736, opened 2023-11-06
Update to latest wasm-tools releases

alexcrichton:


Mossaka (approved):
Maybe we should get a dependabot for doing these bumps.

alexcrichton:
Perhaps yeah! I'm happy to do these manually for the time being though

Issue #713, opened 2023-10-24
C# Implementation Tracker


silesmo:
This is an issue to track the implementation progress of C# with the mono runtime as well as the Native ahead of time LLVM compiled implementation.

- [x] Worlds
- [x] Interfaces
- [x] Functions
- [x] Integers
- [x] Floating point numbers 
- [x] Strings
- [ ] Mono runtime tests - Dependent on (If we don't want to create 2 completely separate implementations to be able to support mono) https://github.com/dotnet/runtime/issues/86985 and https://github.com/dotnet/runtime/issues/86984
- [x] Native AoT LLVM runtime tests
- [x] Results - https://github.com/bytecodealliance/wit-bindgen/pull/807
- [x] Options 
- [x] List - https://github.com/bytecodealliance/wit-bindgen/pull/807
- [x] Enum
- [x] Records
- [x] Flags
- [x] Variants - https://github.com/bytecodealliance/wit-bindgen/pull/807
- [x] Tuples
- [x] Resources
  - [x] Handles
  - [x] Static Functions
  - [x] Methods
- [ ] #1074
- [ ] #1073
- [x] Runtime tests
- [x] Code Generation tests
  - [x] Enabled skipped tests in https://github.com/bytecodealliance/wit-bindgen/blob/5f903613cfb70364b89bbab92d62e85db16cf4e8/crates/csharp/tests/codegen.rs#L15
- bugs
  - [x] https://github.com/bytecodealliance/wit-bindgen/issues/777
  - [x] https://github.com/bytecodealliance/wit-bindgen/issues/778
  - [x] return pointer alignment - in some cases get error might be related to https://github.com/dotnet/runtimelab/pull/2531 (https://github.com/bytecodealliance/wit-bindgen/pull/807)
  - [x] #963

future work:
- [ ] Streams
- [ ] Integration with GC proposal. Related: https://github.com/WebAssembly/gc/issues/77, https://github.com/dotnet/runtime/issues/94420
- [ ] Exception handling https://github.com/WebAssembly/exception-handling/tree/master
- [ ] Shared everything dynamic linking https://github.com/WebAssembly/component-model/blob/9c43f51ebf775e8397469a744d97c8575a7b7386/design/mvp/examples/SharedEverythingDynamicLinking.md

Lower priority items:

- [ ] #1075
- [x] make options implementation use c# generics - Resolved in: https://github.com/bytecodealliance/wit-bindgen/pull/807
- [ ] #1076
[...]

PR #738, opened 2023-11-07
[csharp] Verify codegen

jsturtevant:
The code gen is running but isn't being verified.  This builds the source in the same way as the other languages. 

It fixes the stub code so it uses the stub functions not just creating a class so that the code generated can compile.  It also disables a few tests that are failing, to be fixed later

It uses the CI work enabled in #718, so will rebase after that is in.

Part of #713


jsturtevant:
>There is not enough space on the disk.

Seems like the builds for dotnet take up lots of space 🤔

yowl:
> > There is not enough space on the disk.
> 
> Seems like the builds for dotnet take up lots of space 🤔

Yes, its unfortunate.  Do we need to ask for more space?

jsturtevant:
The only option for getting more space is moving to the Enterprise or custom runners :-/  Interestingly it seems that some times we get a drive with 100GB+  and that's why it passed once.

I think we have a few options:

- Clean up the project after build - Trying this out first
- use the `C:` drive for build outputs - not technically supported but seems that other projects have done this (https://github.com/actions/runner-images/issues/1341#issuecomment-669204383)
- refactor the project builds so they happen on different jobs - I think this is probably the right choice longer time which will also speed up CI and can look at it later

tschneidereit:
I would very much recommend using `C:` for output. Enough projects are depending on that that I have a hard time seeing GH just break it. And even if they do, we can solve it then instead of preemptively trying to solve a problem we might never have :)

crates/csharp/tests/codegen.rs
```diff
@@ -77,6 +85,165 @@ macro_rules! codegen_test {
...
+
+fn aot_verify(dir: &Path, name: &str) {
+    let mut wasm_filename = dir.join(name);
+    wasm_filename.set_extension("wasm");
+
+    fs::write(
+        dir.join("nuget.config"),
+        r#"<?xml version="1.0" encoding="utf-8"?>
+    <configuration>
+        <config>
```
yowl, line 101: I guess this is copied from the runtime version.  I should have removed this really as I don't think there is much benefit and it may improve the CI time without it.

yowl: Its fine though and we can address them both at the same time.

yowl (approved):
Thanks, this will help a lot.

crates/csharp/tests/codegen.rs
jsturtevant: there a bit of duplicate code and this will become more challenging with a different project settings required for mono work.  Let's follow up with a clean up?

PR #747, opened 2023-11-08
Remove handling of empty types from C

alexcrichton:
This is no longer necessary because the component model now disallows empty types. Additionally update some tests to no longer exercise the usage of empty types.


Mossaka (approved):
Great! I will do the same for Go

Mossaka:
Just curious, is this in issue551 `type value = list<tuple<>>;` an empty type?

alexcrichton:
Indeed yeah! That's an empty type and it can be filled in.

PR #737, opened 2023-11-06
feat(go): add resources support

Mossaka:
This is a WIP PR to add `resources` support to `wit-bindgen-go`. 

Should close #696 and #615

TODO list
- [ ] Fix C's resource built-in functions imported module name for exported resources. https://github.com/bytecodealliance/wit-bindgen/pull/748 
- [x] enable `resources` runtime test
- [ ] enable `resources` codegen test
- [ ] enable `return_resource_from_export` codegen test
- [ ] enable `resource_local_alias_borrow` codegen test
- [ ] enable `resource_borrow_in_record` codegen test
- [ ] Fix C's import and export of the same resource issue. #723 
- [ ] test on `wasi-http/proxy` world. #735

PR #748, opened 2023-11-08
fix(c): added the missing [export] and make ResourceInfo pub

Mossaka:
This commit pulls from #737 to add the missing `[export]` to imported `resource.drop` function.


Mossaka:
Actually, I want to move `Direction` to core crate so all the generator can share this enum.

alexcrichton (approved):

PR #750, opened 2023-11-09
fix(c, go): pass the right direction to interface generator in `import_types`

### LINKED ISSUE #749: Mismatched result from the C resource generator
## WIT 1 

```wit
package foo:bar;

world x {
  resource y;

  import f: func(a: borrow<y>);
}
```

wit-bindgen c generates the following header
```c
typedef struct x_y_t x_y_t;

// Imported Functions from `x`
extern void x_f(x_y_t* a);
```

## WIT 2
```wit
package foo:bar;

interface something {
  resource y;
}

world x {
  use something.{y};

  import f: func(a: borrow<y>);
}
```

```c
typedef struct {
  int32_t __handle;
} x_borrow_y_t;

// Imported Functions from `x`
extern void x_f(x_borrow_y_t a);

// Helper Functions

void x_borrow_y_free(x_borrow_y_t *ptr);

// Functions for working with resource `foo:bar/something/y`
extern void x_y_drop_borrow(x_borrow_y_t);
```

## Expected Result

The generated C header file should look pretty much the same except naming and namespaces.

Mossaka:
Closes #749

### DIFF
--- a/tests/runtime/resource_borrow_simple/wasm.c
+++ b/tests/runtime/resource_borrow_simple/wasm.c
@@ -1,0 +1,10 @@
+#include <assert.h>
+#include <resource_borrow_simple.h>
+
+bool resource_borrow_simple_test_imports(void) {
+    resource_borrow_simple_borrow_r_t r = resource_borrow_simple_borrow_r_t {
+        .__handle = 0,
+    }
+    resource_borrow_simple_test(r);
+    resource_borrow_simple_r_drop_borrow(r);
+}
--- a/tests/runtime/resource_borrow_simple/wasm.c
+++ b/tests/runtime/resource_borrow_simple/wasm.c
@@ -1,10 +1,10 @@
 #include <assert.h>
 #include <resource_borrow_simple.h>
 
-bool resource_borrow_simple_test_imports(void) {
-    resource_borrow_simple_borrow_r_t r = resource_borrow_simple_borrow_r_t {
+void resource_borrow_simple_test_imports(void) {
+    resource_borrow_simple_borrow_r_t r = {
         .__handle = 0,
-    }
+    };
     resource_borrow_simple_test(r);
     resource_borrow_simple_r_drop_borrow(r);
 }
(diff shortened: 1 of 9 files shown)


tests/runtime/resource_borrow_simple/wasm.rs
```diff
@@ -0,0 +1,17 @@
...
+        world: Test,
+    }
+});
+
+pub struct Test {}
+
+impl Guest for Test {
+    fn test_imports() {
+        unsafe {
+            let r = R { handle: wit_bindgen::rt::Resource::from_handle(0) };
```
Mossaka, line 13: This doesn't feel good to use... If a resource doesn't have a way to construct a owned handle, how is the guest suppose to do? @alexcrichton

tests/runtime/resource_borrow_simple/wasm.c
```diff
@@ -0,0 +1,10 @@
+#include <assert.h>
+#include <resource_borrow_simple.h>
+
+bool resource_borrow_simple_test_imports(void) {
+    resource_borrow_simple_borrow_r_t r = resource_borrow_simple_borrow_r_t {
```
Mossaka, line 5: same question as the rust one.

Mossaka:
It looks like the generated C has the wrong import module name for `[resource-drop]`

```c
__attribute__((__import_module__("resource-borrow-simple"), __import_name__("[resource-drop]r")))
void __wasm_import_resource_borrow_simple_r_drop_borrow(int32_t);
```

The correct import module name should be `$root`

tests/runtime/resource_borrow_simple/wasm.rs
Mossaka: Nevermind, this isn't working

alexcrichton: The WIT will need to have a `constructor()` or some other means by which the guest can create a resource, because otherwise there's no valid way for the guest to get a resource.

Mossaka:
Close as this is fixed by #751

PR #751, opened 2023-11-09
Refactor generation of types in C

### LINKED ISSUE #629: Support importing and exporting the same resource in the C generator
While implementing https://github.com/bytecodealliance/wit-bindgen/pull/604, I noticed that the C generator will require significant refactoring to support importing and exporting the same interface containing one or more resources.

Currently, it panics on a `todo!` if you give it a WIT world which does that, and if you remove the `todo!` it will generate a single type and set of helper functions for each resource, appropriate for the import side but not the export side.  What we want instead is two separate types and sets of helper functions -- one for the import and one for the export.  This will presumably require duplicating structural types as well if they (transitively) contain handle fields, although it might be easiest to just duplicate all the types in the interface unconditionally.  In any case, this will likely require refactoring, e.g. moving various functions from the `C` type into the `InterfaceGenerator` type and generating helper functions (e.g. destructors) earlier when we still know whether we're importing or exporting the types (similar to how the Rust generator works).  It will also require namespacing exported types with e.g. `exports_` to distinguish them from their imported equivalents (again, inspired by the Rust generator).

### LINKED ISSUE #723: C: import and export of the same resource is not allowed
While I was working on adding runtime tests for the C bindgen, I found that there is a huge limitation at https://github.com/bytecodealliance/wit-bindgen/blob/5572e74449c1d8fcacebe29f927a01924e62a36f/crates/c/src/lib.rs#L186-L189

which prevents the `wit-bindgen`to generate C bindings for most resource related runtime tests. 

I am wondering how much efforts is needed to get this working in C? This work will be needed for Go bindgen to be properly tested to support resources, #696 .

### LINKED ISSUE #727: C: why do borrow/own functions have different prefixes?
```
package foo:bar;

interface hoge {
    resource fuga {
        constructor();
        op: func() -> result;
    }
}

world useless {
    import hoge;
}
```

wit-binden generates code like:
```
// Functions for working with resource `foo:bar/hoge/fuga`
useless_borrow_fuga_t foo_bar_hoge_borrow_fuga(useless_own_fuga_t);
void foo_bar_hoge_fuga_drop_own(useless_own_fuga_t);
extern void useless_fuga_drop_borrow(useless_borrow_fuga_t);
```
i wonder why borrow ones have prefixed with the world and own ones are prefixed with the package/interface.

### LINKED ISSUE #735: c generator outputs invalid typedef for wasi-http/proxy
When generating bindings for `wasi-http/proxywasi:http/proxy@0.2.0-rc-2023-11-05` I see a generated typedef of:
```c
typedef test_wasi_http_own_request_options_t test_wasi_http_own_request_options_t;
```

I tried to reduce the the definitions down to a smaller set of wit files but whenever I did that, the problem went away.

I've created a github repository which contains the exact wit definitions, wit-bindgen version used and the generated bindings - in the hope that they are helpful in figuring out what is happening.

https://github.com/JakeChampion/wasi-http-wit-bindgen-c

Note: It seems we don't have a consistent c binding generator, each time I run `wit-bindgen c --no-helpers  --world test-wasi-http . --out-dir out` in the above repository I get a different order of output in the built files.

### LINKED ISSUE #749: Mismatched result from the C resource generator

alexcrichton:
This commit is an attempt to redo how types are generated and modeled in the C bindings generator. There are a number of existing issues around the C generator related to resources for example, but there are other issues historically that plague the C generator.

This refactoring effectively entirely refactors the generation of types in C. Previously C worked by generating types throughout the bindings process and generating lots of bits at the end of the bindings generation process. This doesn't work for resources because the same resource imported and exported needs different bindings, unlike the one set of bindings generated prior.

The new scheme of generating types looks like:

* All work is now done immediately when a type is visited. This means that helper functions, destructors, etc, are all generated at type generation time. Nothing is deferred until the end of the bindings generation process.
* The split between anonymous and public types in the C generator is removed. Lots of weird infrastructure was present to manage this and the only benefit was that a few types would be omitted from the public header file because they didn't need to be there. Instead this commit takes the strategy of generating all the types all the time, and if they're not used that's no skin off anyone's back.
* When types are generated they now assume that all their dependencies have previously been generated. This makes generation of a reference to a type much easier because it's a simple "that name must be in this map" assertion. This is done by using `LiveTypes` to visit all types for a unit of generation. This topological sorting is then iterated over to emit all types.
* Resources now generate `own` and `borrow` handle types immediately when they're visited. These serve as the canonical versions of these types and then all other references to these handles are done via `typedef`s to ensure that the types all stay the same.
* Anonymous types such as tuples are now duplicated across interfaces. Previously the generator attempted to deduplicate everything into the world namespace but this doesn't work for resources which needs some means of duplicating in some cases. To solve this I went ahead and generated all anonymous types all the time which means that if a WIT files has `tuple<u32, u32>` in two interfaces that'll generate two types despite only needing one.

Overall the generated C bindings are a bit more verbose than before but the generation process is vastly simpler. My hope is that this'll make the C generator less buggy without losing any functionality.

Closes #629
Closes #723
Closes #727
Closes #735
Closes #749

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -2,7 +2,7 @@ mod component_type_object;
 
 use anyhow::Result;
 use heck::*;
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::HashMap;
 use std::fmt::Write;
 use std::mem;
 use wit_bindgen_core::abi::{self, AbiVariant, Bindgen, Bitcast, Instruction, LiftLower, WasmType};
@@ -10,13 +10,6 @@ use wit_bindgen_core::{
     uwrite, uwriteln, wit_parser::*, Direction, Files, InterfaceGenerator as _, Ns, WorldGenerator,
 };
 use wit_component::StringEncoding;
-
-#[derive(Default)]
-pub struct ResourceInfo {
-    pub direction: Direction,
-    pub borrow: Option<TypeId>,
-    pub own: Option<TypeId>,
-}
 
 #[derive(Default)]
 struct C {
@@ -30,37 +23,15 @@ struct C {
     world: String,
     sizes: SizeAlign,
 
-    // Known names for interfaces as they're seen in imports and exports.
-    //
-    // This is subsequently used to generate a namespace for each type that's
-    // used, but only in the case that the interface itself doesn't already have
-    // an original name.
-    interface_names: HashMap<InterfaceId, WorldKey>,
-
-    // Interfaces who have had their types printed.
-    //
-    // This is used to guard against printing the types for an interface twice.
-    // The same interface can be both imported and exported in which case only
-    // one set of types is generated and all bindings for both imports and
-    // exports use that set of types.
-    interfaces_with_types_printed: HashSet<InterfaceId>,
-
-    // Type definitions for the given `TypeId`. This is printed topologically
-    // at the end.
-    types: HashMap<TypeId, wit_bindgen_core::Source>,
-
+    dtor_funcs: HashMap<TypeId, String>,
+    type_names: HashMap<TypeId, String>,
     resources: HashMap<TypeId, ResourceInfo>,
+}
 
-    // The set of types that are considered public (aka need to be in the
-    // header file) which are anonymous and we're effectively monomorphizing.
-    // This is discovered lazily when printing type names.
-    public_anonymous_types: BTreeSet<TypeId>,
-
-    // This is similar to `public_anonymous_types` where it's discovered
-    // lazily, but the set here are for private types only used in the
-    // implementation of functions. These types go in the implementation file,
-    // not the header file.
-    private_anonymous_types: BTreeSet<TypeId>,
+struct ResourceInfo {
+    direction: Direction,
+    own: String,
+    borrow: String,
 }
 
 #[derive(Default, Debug, Clone)]
@@ -121,13 +92,9 @@ impl WorldGenerator for C {
         id: InterfaceId,
         _files: &mut Files,
     ) {
-        let prev = self.interface_names.insert(id, name.clone());
-        assert!(prev.is_none());
         let mut gen = self.interface(resolve, true);
-        gen.interface = Some(id);
-        if gen.gen.interfaces_with_types_printed.insert(id) {
-            gen.types(id);
-        }
+        gen.interface = Some((id, name));
+        gen.define_interface_types(id);
 
         for (i, (_name, func)) in resolve.interfaces[id].functions.iter().enumerate() {
             if i == 0 {
@@ -149,6 +116,7 @@ impl WorldGenerator for C {
     ) {
         let name = &resolve.worlds[world].name;
         let mut gen = self.interface(resolve, true);
+        gen.define_function_types(funcs);
 
         for (i, (_name, func)) in funcs.iter().enumerate() {
             if i == 0 {
@@ -167,22 +135,9 @@ impl WorldGenerator for C {
         id: InterfaceId,
         _files: &mut Files,
     ) -> Result<()> {
-        self.interface_names.insert(id, name.clone());
         let mut gen = self.interface(resolve, false);
-        gen.interface = Some(id);
-        if gen.gen.interfaces_with_types_printed.insert(id) {
-            gen.types(id);
-        } else {
-            let iface = &resolve.interfaces[id];
-            for id in iface.types.values() {
-                if let TypeDefKind::Resource = &resolve.types[*id].kind {
-                    // This will require a substantial refactor so we can
-                    // generate two sets of types and helper functions, one for
-                    // the imported types and one for the exported types.
-                    todo!("importing and exporting the same interface containing a resource not yet supported");
-                }
-            }
-        }
+        gen.interface = Some((id, name));
+        gen.define_interface_types(id);
 
         for (i, (_name, func)) in resolve.interfaces[id].functions.iter().enumerate() {
             if i == 0 {
@@ -205,6 +160,7 @@ impl WorldGenerator for C {
     ) -> Result<()> {
         let name = &resolve.worlds[world].name;
         let mut gen = self.interface(resolve, false);
+        gen.define_function_types(funcs);
 
         for (i, (_name, func)) in funcs.iter().enumerate() {
             if i == 0 {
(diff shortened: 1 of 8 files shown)


alexcrichton:
cc @Mossaka this is a pretty big change and going to affect Go a good deal

alexcrichton:
While I'm here working on the C generator I also folded in https://github.com/bytecodealliance/wit-bindgen/pull/691 and fixed up the remaining issues there.

Mossaka:
Thanks! Let me rebase my PR #737 to this branch

Mossaka:
Sorry been pulled away for other things. Will review this PR as the first thing tomorrow.

dicej (commented):
Looks great overall, and I love the simplifications.

I tried generating code for tests/codegen/import-and-export-resource.wit because I was curious what it would look like, and I'm a bit confused by the result:

```c
typedef struct my_resources_baz_own_x_t {
  int32_t __handle;
} my_resources_baz_own_x_t;

typedef struct my_resources_baz_borrow_x_t {
  int32_t __handle;
} my_resources_baz_borrow_x_t;

typedef my_resources_baz_own_x_t my_resources_baz_own_x_t;

typedef my_resources_baz_borrow_x_t my_resources_baz_borrow_x_t;

// Imported Functions from `my:resources/baz`
extern my_resources_baz_own_x_t my_resources_baz_constructor_x(resources_string_t *s);
extern void my_resources_baz_method_x_get(my_resources_baz_borrow_x_t self, resources_string_t *ret);

// Exported Functions from `my:resources/baz`
my_resources_baz_own_x_t exports_my_resources_baz_constructor_x(resources_string_t *s);
void exports_my_resources_baz_method_x_get(my_resources_baz_borrow_x_t self, resources_string_t *ret);
```
It seems to me there's a name collision between the imported and exported versions of the `x` resource handles.  Am I missing something?

alexcrichton:
>  Am I missing something?

Nope I think you've found a bug, thanks for catching! I'll work to fix.

alexcrichton:
Ok the latest commit should fix that issue and additionally add a new test which exercises this all

dicej (approved):

crates/c/src/lib.rs
```diff
@@ -441,6 +453,61 @@ impl C {
...
+    /// with the same `TypeId` internally within the `wit-parser`
+    /// representation. This means that duplicate types need to be generated for
+    /// exports, even if the same interface was already imported. If nothing
+    /// were done here though then the same type imported and exported wouldn't
+    /// generate anything new since preexisting types are skipped in
+    /// `define_live_types`.
+    ///
+    /// This function will trim the sets on `self` to only retain those types
+    /// which exports refer to that come from imports.
+    fn remove_types_redefined_by_exports(&mut self, resolve: &Resolve, world: WorldId) {
```
Mossaka, line 470: Alternatively, can we add a fourth parameter to `InterfaceGenerator` functions like `type_resource` named `direction: Direction`. Then we can generate types according to this direction, to avoid this specific logic of removing redefined types by exports?

crates/c/src/lib.rs
```diff
@@ -1159,90 +544,45 @@ pub fn push_ty_name(
...
-    if let Type::Id(i) = ty {
-        let namespace = owner_namespace(resolve, *i, interface_names);
-        if let Some(namespace) = namespace {
-            dst.push_str(&namespace);
-            dst.push_str("_");
-        }
-    }
-}
-
 pub fn owner_namespace(
```
Mossaka, line 586: Not used. Could you please remove this function?

crates/c/src/lib.rs
```diff
@@ -1961,19 +1616,40 @@ impl InterfaceGenerator<'_> {
...
+    }
+
+    fn owner_namespace(&self, id: TypeId) -> String {
+        let ty = &self.resolve.types[id];
+        match (ty.owner, self.interface) {
+            // If this type is owned by an interface, then we must be generating
+            // bindings for that interface to proceed.
+            (TypeOwner::Interface(a), Some((b, key))) => {
+                assert_eq!(a, b);
+                interface_identifier(key, self.resolve, !self.in_import)
```
Mossaka, line 1632:
```suggestion
            (TypeOwner::Interface(a), Some((b, key))) if a == b => {
```

crates/c/src/lib.rs
```diff
@@ -1632,6 +1062,231 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
 }
 
 impl InterfaceGenerator<'_> {
+    fn define_interface_types(&mut self, id: InterfaceId) {
+        let mut live = LiveTypes::default();
```
Mossaka, line 1066: Question: what are LiveTypes?

Mossaka (approved):

Mossaka:
One more issue that I found at https://github.com/bytecodealliance/wit-bindgen/pull/750#issuecomment-1803101910

Given this WIT package

```wit
package test:resource-borrow-simple;

world resource-borrow-simple {
  resource r;
  import test: func(r: borrow<r>);
  export test-imports: func();
}
```

The C function imports `[resource-drop]r` for module `resource-borrow-simple`, but the module name should be `$root`, no?

```c
__attribute__((__import_module__("resource-borrow-simple"), __import_name__("[resource-drop]r")))
extern void __wasm_import_resource_borrow_simple_r_drop(int32_t handle);
ccc

Mossaka:
This can be a follow-up task: add integration tests for all `runtime/resources*` for C gen.

crates/c/src/lib.rs
alexcrichton: It's a structure [in the `wit-parser` crate](https://docs.rs/wit-parser/latest/wit_parser/struct.LiveTypes.html) which is used to discover all types reachable from another construct.

PR #753, opened 2023-11-10
add `Resource::take` method to `wit_bindgen::rt`

dicej:
This allows the guest to take back ownership of an exported resource from the host; you can think of it as the reverse of `Resource::new`.  It's a bit awkward to do this for the time being;
https://github.com/WebAssembly/component-model/issues/238 will improve the situation if accepted.

### DIFF
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -172,6 +172,8 @@ pub mod rt {
     }
 }
 
+type RawRep<T> = Option<T>;
+
 /// A type which represents a component model resource, either imported or
 /// exported into this component.
 ///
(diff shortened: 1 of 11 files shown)


guybedford (approved):
This is required and has been tested to work for the WASI-Virt upgrade path.

crates/guest-rust/src/lib.rs
```diff
@@ -250,7 +250,18 @@ impl<T: WasmResource> Resource<T> {
     where
         T: RustResource,
     {
-        let _ = Box::from_raw(rep as *mut T);
+        let _ = Box::from_raw(rep as *mut Option<T>);
+    }
+
+    /// Takes back ownership of the object, dropping the resource handle.
+    pub fn take(resource: Self) -> T
```
alexcrichton, line 257: Could this be rename to `into_inner` perhaps? The `take` terminology seems derivitive of `Option<T>` which is ideally only an implementation detail here.

crates/guest-rust/src/lib.rs
```diff
@@ -250,7 +250,18 @@ impl<T: WasmResource> Resource<T> {
     where
         T: RustResource,
     {
-        let _ = Box::from_raw(rep as *mut T);
+        let _ = Box::from_raw(rep as *mut Option<T>);
```
alexcrichton, line 253: Would you be up for a small refactoring of this module? If so I think it'd be good to have something like 

```rust
type RawRep<T> = Option<T>;
```

or something like that where all of these casts and various "raw" methods would work with `RawRep<T>`. That way future changes to `RawRep<T>` would help trigger compilation errors where appropriate. Basically I think it'd be good to centralize the representation of `Option<T>` if possible.

crates/rust/src/bindgen.rs
```diff
@@ -443,7 +443,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                                     .as_deref()
                                     .unwrap()
                                     .to_upper_camel_case();
-                                format!("&*({op} as u32 as usize as *const {name})")
+                                format!("Option::as_ref(&*({op} as u32 as usize as *const Option<{name}>)).unwrap()")
```
alexcrichton, line 446: Mind adding a helper method to `Resource<T>` which is `#[doc(hidden)]` to call here? That way this could avoid duplicating knowledge of the internal representation across the code generator and the actual source

alexcrichton (approved):
Thanks! Looks reasonable to me and just a few cosmetic comments.

dicej:
@alexcrichton thanks for the feedback; I just pushed and update.  I wasn't sure about the scope of your `RawRep` suggestion; please let me know if there's more to it than what I did in the latest commit.

alexcrichton:
👍

PR #754, opened 2023-11-15
refactor(go): rewrite to adopt the new C generated types

Mossaka:
This PR builds on top of #751 to rewrite the Go generator to adopt the new C generated types. It also refactors a bunch of code and removes unnecessary code.

Note that there are no changes to the generated Go types, and thus no changes to the Go specific tests.


Mossaka:
Close in favor of #755

PR #756, opened 2023-11-16
fix(csharp): export the world stubs with correct namespace

jsturtevant:
This also re-enables a few of the tests disabled in #738


silesmo (approved):
Looks good!

PR #758, opened 2023-11-17
refactor(*): move anonymous type definition to core

Mossaka:
This PR moves the `define_anonymous_type` in C to the core crate, as I believe this is a common pattern in all wit binding languages generators. The benefit of this refactoring work is that language generator does not need to decide which types are anonymous and that is handled by the trait function itself. 

CC @alexcrichton

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -1346,7 +1346,39 @@ impl<'a> wit_bindgen_core::AnonymousTypeGenerator<'a> for InterfaceGenerator<'a>
         self.resolve
     }
 
-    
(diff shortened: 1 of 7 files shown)


Mossaka:
It seems like this PR needs a huge rebase. @alexcrichton any suggestions if we should proceed with this refactoring?

alexcrichton:
Oh dear I apologize I think github doesn't send out notifications for when a PR moves from draft to ready state, so I never saw this after it switched states. I think this is still reasonable to land though if you'd like to rebase!

Mossaka:
@alexcrichton okay I finally found some time to rebase this PR. it's ready for review.

alexcrichton (approved):

Issue #696, opened 2023-10-11
wit-bindgen go: support resources


pchickey:

Mossaka:
Duplicate of #615

Mossaka:
I've started working on adding runtime tests for rust and C to the repo. Once this is done I will start implementing resources in Go and write more tests!

Issue #615, opened 2023-07-12
Go: implement `resource` types


Mossaka:
Once https://github.com/bytecodealliance/wit-bindgen/pull/604 is merged in, the `resource` type should be supported in go-bindgen.

PR #759, opened 2023-11-17
feat(go): add resources support

Mossaka:
[Moved from PR #737]

## Introduction
This PR refactors the wit-bindgen-go to mainly support WIT Resources. 

Should close https://github.com/bytecodealliance/wit-bindgen/issues/696 and https://github.com/bytecodealliance/wit-bindgen/issues/615

## Resources
This PR has enabled WIT resources in the Go binding generator. The user interface for dealing with imported and exported resources look something like this

### Exported Resources
- Unlike other types that each interface generates a Go interface to implement, exported resource generates its own Go interface. Suppose we have a WIT that looks like
```WIT
export exports: interface {
    resource x {
      constructor(a: s32);
      get-a: func() -> s32;
      set-a: func(a: s32);
      add: static func(x: x, a: s32) -> x;
    }
}
```
Then the generated Go code contains an interface `ExportsX` waiting to be implemented. This interface has all the methods `GetA` and `SetA` in it except constructor and static functions, which belong to the WIT interface's generated Go interface. The Go code looks like:
```Go
type ExportsImpl struct {}

type MyX struct {
	a int32
}

func (e ExportsImpl) ConstructorX(a int32) ExportsX {
	return &MyX{a: a}
}

func (x *MyX) MethodXGetA() int32 {
	return x.a
}

func (x *MyX) MethodXSetA(a int32) {
	x.a = a
}

func (e ExportsImpl) StaticXAdd(x ExportsX, a int32) ExportsX {
	return &MyX{a: x.MethodXGetA() + a}
}
```
Note that if there are no methods in the exported resource, than the generated Go interface is just an empty interface. Any struct implementation would work.

### Imported Resources
Now moving on to the imported resources. Imported resources have a better Go semantics than exported ones. For example, given the following WIT resource
```wit
import imports: interface {
    resource y {
      constructor(a: s32);
      get-a: func() -> s32;
      set-a: func(a: s32);
      add: static func(y: y, a: s32) -> y;
    }
  }
```
You can use the Go APIs to interact with resource Y in the following way:
```Go
y := NewY(1)
if y.GetA() != 1 {
panic("y.GetA() != 1")
}
y.SetA(2)
if y.GetA() != 2 {
panic("y.GetA() != 2")
}

y2 := StaticYAdd(y, 3)
if y2.GetA() != 5 {
panic("y2.GetA() != 5")
}

y.SetA(5)

if y.GetA() != 5 {
panic("y.GetA() != 5")
}
```

Note that if the resource is empty (meaning it doesn't have any methods), then that resource's representation is just a `int32`. 

## More Improvement
- This PR refactors the `lib.rs` file into smaller files, each dealing with their own responsibility. For instance, there is a `bindgen.rs` file dealing with function generator, and `interface.rs` dealing with interface generator. 
- It adds a new `gofmt` flag that runs `gofmt` to generated Go files. 

## Breaking Changes
- It separated exported types from imported types. This is BREAKING change that the existing usage of exported types need to be prefixed with a `Export` to fix some minor bugs. More on this later with the package refactoring work.


Mossaka:
I think this PR is ready to review. I will write more in detail what this PR has changed.

crates/c/src/lib.rs
```diff
@@ -471,46 +468,51 @@ impl C {
...
-                }
-                WorldItem::Type(_) => unreachable!(),
+        let live_import_types = find_live_export_types(resolve, world);
+        self.dtor_funcs.retain(|k, _| live_import_types.contains(k));
+        self.type_names.retain(|k, _| live_import_types.contains(k));
+        self.resources.retain(|k, _| live_import_types.contains(k));
+    }
+}
+
+pub fn find_live_export_types(resolve: &Resolve, world: WorldId) -> HashSet<TypeId> {
```
alexcrichton, line 478: The name here isn't quite what this function is returning, perhaps something like `imported_types_used_by_exported_interfaces` or something like that?

alexcrichton (commented):
Thanks! I'm not reviewing the Go bits too too closely but all the other bits look good to me. There's some comments of mine from https://github.com/bytecodealliance/wit-bindgen/pull/757 that are applicable here as well, but otherwise this looks good to go

Mossaka:
Yea, I will resolve the comments in #757 and rebase onto it.

alexcrichton:
@Mossaka want to rebase this and I'll do a final pass?

Mossaka:
Rebased

crates/c/src/lib.rs
```diff
@@ -1094,6 +1090,24 @@ void __wasm_export_{ns}_{snake}_dtor({ns}_{snake}_t* arg) {{
     }
 }
 
+impl<'a> InterfaceGenerator<'a> {
+    fn reosurce_handle_names(&mut self, id: TypeId, name: &str) -> (String, String) {
```
alexcrichton, line 1094: I think this and the `match` just below may have snuck back in by accident?

Mossaka: umm you're right let me fix that

alexcrichton (approved):
Happy to merge as-is, I can follow-up with the cleanup I mentioned

Mossaka:
I will merge once the CI passes. Thanks for reviewing @alexcrichton

alexcrichton:
And thank you again for your work on this!

alexcrichton:
sorry wrong button

PR #762, opened 2023-11-21
refactor(c#): Consolidate cs project generation code

jsturtevant:
As pointed out in #738 there was quite a bit of duplicate csharp code for project generation. This consolidates it to one location.  Maybe in the future we can have a flag that would generate the project file for when generating?

I did try removing the nuget.config (https://github.com/bytecodealliance/wit-bindgen/pull/738#discussion_r1392547364) but I found those are needed for the naot compiler.


yowl (approved):
Looks good, thanks!

PR #767, opened 2023-11-27
Give the 'path' config meaning when used together with 'inline' in Rust macro

### LINKED ISSUE #742: [Rust]: Allow for world dependencies
Assume the following:

```rust
wit_bindgen::generate!({
    inline: r#"
        package acme:test

        world my_world {
            export wasi:http/incoming-handler@0.2.0-rc-2023-10-18;
        }
    "#,
});
```

I don't believe there is any way to specify how the `wasi:http` package should be resolved. I believe the only possibility to get this to work is copy the world into a `wit` directory and provide all dependencies in the deps folder. 

The `wasmtime::component::bindgen` macro allows specifying `inline` and `path` together where `inline` will be assumed to be the world definition and `path` will be used to resolve dependencies. This is, however, not allowed by `wit_bindgen::generate` macro which fails with the following error:

```
cannot specify second source
```

Should we follow the lead of `wasmtime::component::bindgen` and allow `path` to be used as a way to specify a folder with dependencies or is this behavior in `wasmtime` actually not what we want and we should forge a new path in `wit-bindgen`?

rylev:
Fixes #742 

Previously, users could only specify 'inline' or 'path' based wit packages. However, the wasmtime component macro allows for 'path' to specify a directory where the inline wit package's dependencies live.

This change brings the same semantics to the wit-bindgen macro.

### DIFF
--- a/tests/runtime/main.rs
+++ b/tests/runtime/main.rs
@@ -11,12 +11,13 @@ use wasmtime::component::{Component, Instance, Linker};
 use wasmtime::{Config, Engine, Store};
 use wasmtime_wasi::preview2::{Table, WasiCtx, WasiCtxBuilder, WasiView};
 use wit_component::{ComponentEncoder, StringEncoding};
-use wit_parser::{Resolve, WorldItem};
+use wit_parser::{Resolve, WorldId, WorldItem};
 
 mod flavorful;
 mod lists;
 mod many_arguments;
 mod numbers;
+mod other_dependencies;
 mod ownership;
 mod records;
 mod resource_aggregates;
@@ -117,10 +118,6 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
     let mut dir = PathBuf::from("./tests/runtime");
     dir.push(dir_name);
 
-    let mut resolve = Resolve::new();
-    let (pkg, _files) = resolve.push_dir(&dir).unwrap();
-    let world = resolve.select_world(pkg, None).unwrap();
-
     let mut rust = Vec::new();
     let mut c = Vec::new();
     let mut java = Vec::new();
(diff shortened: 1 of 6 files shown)


tests/runtime/main.rs
```diff
@@ -179,87 +176,90 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
...
-        drop(fs::remove_dir_all(&out_dir));
-        fs::create_dir_all(&out_dir).unwrap();
-
-        let snake = world_name.replace("-", "_");
-        let mut files = Default::default();
-        let mut opts = wit_bindgen_c::Opts::default();
-        if let Some(path) = path.file_name().and_then(|s| s.to_str()) {
-            if path.contains("utf16") {
-                opts.string_encoding = wit_component::StringEncoding::UTF16;
+    if !c.is_empty() {
```
rylev, line 179: Most of this change is from indenting this code. This is needed so that we only call `resolve_wit_dir` when there are actual non-Rust tests to run. This is because Rust now has one test where the world name cannot be determined from the contents of `dir` (the world is located in the 'inline' key of the call to `generate!`). 

The downside of this is that we're now resolving the world for each language instead of once. Let me know if you want me to refactor to avoid this duplicate work.

alexcrichton (approved):
Thanks! Happy to merge as-is, but if you're interested an easier testing story might be to only add a codegen test in `crates/rust/tests/codegen.rs` since the runtime behavior of this option isn't the main bit but rather ensuring the right bindings are generated.

alexcrichton:
I'm going to [merge the version bump](https://github.com/bytecodealliance/wit-bindgen/pull/768) now and am going to merge this ahead of that to get this in. Given that this might result in a slight update to tests that's fine to go either way, and if you'd like feel free to follow-up with that, but it's also ok as-is!

rylev:
@alexcrichton I'm happy to change the test, but I'm not sure exactly how - the test condition requires changes to the `generate!` macro invocation and the codegen tests all share the same `generate!` macro call. I guess I could change that call (which would change it for all codgen tests), but that seems like a rather indirect way of testing this scenario. I agree the current approach is a bit verbose, but it does seem like the most straight forward way the wit-bindgen testing framework allows for testing this scenario. Let me know if I'm missing something obvious!

alexcrichton:
Oh sorry, I don't mean tests in `tests/codegen/*.wit`, but rather one-off tests [like this](https://github.com/bytecodealliance/wit-bindgen/blob/6d6c315c23ce1a5b64cb40cf204e96b0ed45cfe3/crates/rust/tests/codegen.rs#L376-L402) for specific options/features of the Rust bindgen macro.

PR #772, opened 2023-11-28
Set the path to the wasi adapter properly

### LINKED ISSUE #766: Runtime tests don't respect `CARGO_TARGET_DIR`
The runtime tests will fail if `CARGO_TARGET_DIR` is set to anything other than to the target dir in the root of the workspace. The `CARGO_TARGET_DIR` should be respected when running the tests.

rylev:
Fixes #766 

Previously, the adapter as assumed to be at set relative path to the target directory which was not true when CARGO_TARGET_DIR was set. Now, the adapter is assumed to be at a set relative path to the artifacts crate which should always be the case.


alexcrichton (approved):
r=me, but I think there's a rustfmt issue

PR #774, opened 2023-11-30
fix(go, c): escape ret and err

Mossaka:
`"ret"` and `"err"` are preserved as variable names facilitating pointer types and "option" and "result" flattening in the C bindgen. As a result, they should be escaped as parameters and otherwise will cause issues as shown in the added test. This also fixed affliated issue in Go bindgen.


alexcrichton:
Thanks! You mention an added test, but was that perhaps forgotten to get added?

Mossaka:
> but was that perhaps forgotten to get added?

Yeah you're right 🙀

PR #776, opened 2023-11-30
adds new WasmImportLinkAttribute for C# and initial support for mono …

silesmo:
WIP


jsturtevant:
can I try this out with the nightly wasi-experimental workload? I think CI install scripts will need to be updated.

silesmo:
> can I try this out with the nightly wasi-experimental workload? I think CI install scripts will need to be updated.

Yes the CI scripts needs to be updated. I will do that once it works as expected.

silesmo:
Dependent upon https://github.com/dotnet/runtime/issues/95084

crates/csharp/src/csproj.rs
```diff
@@ -3,7 +3,9 @@ use std::{fs, path::PathBuf};
 
 use heck::ToUpperCamelCase;
 
-pub struct CSProject {
+pub struct CSProject;
+
+pub struct CSProjectLLVMBuilder {
```
yowl, line 8: Somewhere in here we will need to add the DIrectPInvoke instructions.  The WasmImportLinkageAttribute tells the compiler to emit the wasm import attributes to LLVM, but these methods also need to be "Direct PInvokes" from a NAOT perspective https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/interop#direct-pinvoke-calls.  This means adding someting like 
```cs
 <ItemGroup>
   <DirectPInvoke Include="xyz!SimpleDirectPInvokeTestFunc" />
 </ItemGroup>
```
where xyz is the Wasm import module name, and SimpleDirectPInvokeTestFunc is the Wasm import function name
or perhaps better, more concise
```cs
 <ItemGroup>
   <DirectPInvoke Include="xyz />
 </ItemGroup>
```
which will mark all exports in the `xyz` module as direct P/invokes.

silesmo:
There are still some issues with the generated components as they can't be ran. There seems to be something incorrect with the types. But I think we can merge this for now as it doesn't break anything already existing and it allows us to actually build using mono as well and test the code gen with mono.

crates/csharp/src/csproj.rs
yowl: We don't actually need this now, as Naot for Wasm treats all `WasmImportLinkage` marked methods as direct pInvoke

PR #781, opened 2023-12-02
c#: Encode version in namespace to support wit versions

jsturtevant:
Prior to this change the multi-version wit was putting the functions from different versions into the same class which was causing the test to fail as there were multiple implementations of the same function:

```
C:\Users\jstur\projects\wit-bindgen\target\codegen-tests\guest-csharp-multiversion\foo\Wit.imports.my.dep.AInterop.cs(25,27): error CS0102: The type 'AInterop' already contains a definition for 'XInterop' [C:\Users\jstur\projects\wit-bindgen\target\codegen-tests\guest-csharp-multiversion\foo\TheWorld.csproj]
C:\Users\jstur\projects\wit-bindgen\target\codegen-tests\guest-csharp-multiversion\foo\Wit.imports.my.dep.AInterop.cs(31,33): error CS0111: Type 'AInterop' already defines a member called 'X' with the same parameter types [C:\Users\jstur\projects\wit-bindgen\target\codegen-tests\guest-csharp-multiversion\foo\TheWorld.csproj]
C:\Users\jstur\projects\wit-bindgen\target\codegen-tests\guest-csharp-multiversion\foo\Wit.imports.my.dep.AInterop.cs(28,36): error CS0111: Type 'AInterop.XInterop' already defines a member called 'wasmImportX' with the same parameter types [C:\Users\jstur\projects\wit-bindgen\target\codegen-tests\guest-csharp-multiversion\foo\TheWorld.csproj]
```

Now the it generates a file for each version (if present) with the functions namespaced by version:

```
wit_foo.wit.imports.my.dep.v0_1_0.A;
```

It also creates a file per version:

file per version of wit

To note:

-  namespaces parts can't start with numbers so put `v` in front.  
- This also lowercases all the names in the namespace, but we had a mixture of upper and lower case so this makes it atleast consistent.  Can maybe follow up to make it more csharp like

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -567,13 +567,13 @@ impl InterfaceGenerator<'_> {
         if let TypeOwner::Interface(id) = &ty.owner {
             if let Some(name) = self.gen.interface_names.get(id) {
                 if name != self.name {
-                    return format!("{}.", name.to_upper_camel_case());
+                    return format!("{}.", name);
                 }
             }
         }
 
         if when {
-            let name = self.name.to_upper_camel_case();
+            let name = self.name;
             format!("{name}.")
         } else {
             String::new()
@@ -583,7 +583,7 @@ impl InterfaceGenerator<'_> {
     fn add_interface_fragment(self, is_export: bool) {
         self.gen
             .interface_fragments
-            .entry(self.name.to_upper_camel_case())
+            .entry(self.name.to_string())
             .or_insert_with(|| InterfaceTypeAndFragments::new(is_export))
             .interface_fragments
             .push(InterfaceFragment {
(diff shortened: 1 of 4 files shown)


jsturtevant:
The refactor here exposed an issue with the way namespaces and files where generating:

Prior to this change (on main) we were getting files like 

```
namespace wit_w.wit.exports.name3Name3Name3;

## file name
Wit.exports.name3Name3Name3.cs
```

image


Now we get 

```
namespace wit_w.wit.exports.name3.name3.Name3;

## file name
Wit.exports.name3.name3.Name3.cs
```

Which is actual what we really want.  With #782 and a small tweak to the import Interop class this will pass

jsturtevant:
has cherry-pick from #782

silesmo (approved):
Looks good. Thank you!

yowl:
Thanks!

PR #783, opened 2023-12-03
Add a `--rename` option to the C generator

alexcrichton:
The generated symbols for types like
`wasi:http/types@0.2.0-rc-2023-10-18` can get quite long so add an option to rename them explicitly if desired.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -119,8 +119,13 @@ impl WorldGenerator for C {
         }
 
         for (from, to) in self.opts.rename.iter() {
-            if let Some(key) = interfaces.get(from) {
-                self.renamed_interfaces.insert(key.clone(), to.clone());
+            match interfaces.get(from) {
+                Some(key) => {
+                    self.renamed_interfaces.insert(key.clone(), to.clone());
+                }
+                None => {
+                    eprintln!("warning: rename of `{from}` did not match any interfaces");
+                }
             }
         }
     }


crates/c/src/lib.rs
```diff
@@ -90,6 +110,19 @@ impl WorldGenerator for C {
...
+
+        let mut interfaces = HashMap::new();
+        let world = &resolve.worlds[world];
+        for (key, _item) in world.imports.iter().chain(world.exports.iter()) {
+            let name = resolve.name_world_key(key);
+            interfaces.insert(name, key.clone());
+        }
+
+        for (from, to) in self.opts.rename.iter() {
+            if let Some(key) = interfaces.get(from) {
```
dicej, line 122: Sound we consider printing warning here if the user specifies a rename that's not used?

dicej (approved):
This looks great, thanks.

In the future, it would also be nice to have an option to leave the version numbers out of the generated bindings in cases where there's no ambiguity, which would be a quick, easy way to get shorter names without explicitly renaming everything.  And maybe another that leaves the package name out, too.

alexcrichton:
Agreed! I think it'd be good to have a PR for that and there's probably some common logic to share with the Rust implementation which has similar behavior.

Mossaka:
Chime in here, I am working on splitting the Go bindings namespaces to each Go packages and have also used the same logic to hide version if there is no ambiguitiy in the generated bindings. 

I can help to refactor rust, c and go bindgen to use a common logic that enables this and provides a consistent behaviour.

alexcrichton:
Yeah I think that'd be great to have!

PR #784, opened 2023-12-04
feat(go-namespaces): Split generated Go files into multiple packages

Mossaka:
This commit overhauls the `WorldGenerator` file structure in the tinygo bindgen. Key changes include:
- Splitting generated Go files into different packages for each interface, imported functions, and exported functions.
- Simplifying function and symbol names by removing namespaces.
- Adding a `package_name` option to Go bindgen for prefixed package names. Defaults to `main` if not specified.

Internally, the C files are placed at a special Go package named "c_files_", and a dummy Go file is generated to expose the C APIs. Each interface generates a Go package that imports this special "c_files_" package to get the C apis. It still has to import the C header file which is also placed at "c_files_" by setting CFLAGS to locate where that is for that Go package.

This restructuring enhances usability by making symbol and function names more readable and less verbose, significantly improving the user experience.

## TODO

- [ ] Generate a dummy `main` file for each codegen test, and that `main` file needs to import all the generated Go packages.
- [ ] Enable the rest of the runtime tests
- [ ] Figure out how to import Go packages to implement type aliasing
- [ ] More testing with wasi:http

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -5,6 +5,7 @@ use heck::*;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Write;
 use std::mem;
+use std::path::PathBuf;
 use wit_bindgen_core::abi::{self, AbiVariant, Bindgen, Bitcast, Instruction, LiftLower, WasmType};
 use wit_bindgen_core::{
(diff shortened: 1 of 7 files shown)


Mossaka:
This work has been paused for a few months since there is [a new pure-Go implementation](https://github.com/ydnar/wasm-tools-go) of the wit-bindgen-go undergoing.

PR #789, opened 2023-12-08
c: Emit single types for monomorphic primitives

elliottt:
For types like `list<u8>`, we don't need to emit a definition for each interface that uses that type. Instead we can define a single one per-world, and refer to it consistently in the generated code.

This arises when depending on other packages that use types composed of `tuple`, `list`, `option`, and primitive types. For example, you might end up with multiple definitions of `list<u8>` when using both wasi-http and wasi-io, yielding the following two declarations in your bindings.h:
```
  typedef struct {
    uint8_t *ptr;
    size_t len;
  } wasi_io_0_2_0_rc_2023_10_18_streams_list_u8_t;

  typedef struct {
    uint8_t *ptr;
    size_t len;
  } wasi_http_0_2_0_rc_2023_10_18_types_list_u8_t;
```

This PR avoids this situation by generating consistent names for types composed of `tuple`, `list`, `option`, and other primtivie types, and then only generating these declarations once. As a result, the following two declarations above collapse into the single decl (in this case the world name was `bindings`):
```
  typedef struct {
    uint8_t *ptr;
    size_t len;
  } bindings_list_u8_t;
```

Having a consistent name for composite primitives like `list<u8>` is useful for writing generic functions that operate on those values, and needing to duplicate code for each dependency's version of that type increases the maintenance burden for guest code quite a bit.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -593,21 +593,7 @@ fn is_prim_type_id(resolve: &Resolve, id: TypeId) -> bool {
     }
 }
 
-#[derive(Clone, Copy, Eq, PartialEq)]
-enum Scope {
-    External,
-    Internal,
-}
-
-pub fn push_internal_ty_name(resolve: &Resolve, ty: &Type, src: &mut String) {
-    push_ty_name(Scope::Internal, resolve, ty, src)
-}
-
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):
Thanks for this!

Issue #777, opened 2023-12-01
c#: dotnet Runtime not started correctly when using strings (or heap objects) as first call to exported function 


jsturtevant:
One thing that doesn't seem to work quite right is the current .NET implementation of `cabi_realloc`. It doesn't seem to start the runtime correctly. So, if you're exporting a function that accepts a string, and the very first call into your preview2 component is to that method, the very first thing it will do is call `cabi_realloc` which will fail with something like this:

```
Caused by:
    0: failed to invoke `run` function
    1: error while executing at wasm backtrace:
           0: 0x18fa7d7 - wit-component:shim!indirect-wasi:cli/environment@0.2.0-rc-2023-10-18-get-environment
           1: 0x18ef789 - wit-component:adapter:wasi_snapshot_preview1!wasi_snapshot_preview1::State::get_environment::h3923f56a626a4100
           2: 0x18ef996 - wit-component:adapter:wasi_snapshot_preview1!environ_sizes_get
           3: 0x18fa80d - wit-component:shim!adapt-wasi_snapshot_preview1-environ_sizes_get
           4: 0x145aac8 - <unknown>!__wasi_environ_sizes_get
           5: 0x145b358 - <unknown>!__wasilibc_initialize_environ
           6: 0x145b3f8 - <unknown>!__wasilibc_initialize_environ_eagerly
           7: 0xcc784c - <unknown>!__wasm_call_ctors
           8: 0xcc787f - <unknown>!_initialize
           9: 0xccb10c - <unknown>!InitializeRuntime()
          10: 0xccb4c6 - <unknown>!Thread::ReversePInvokeAttachOrTrapThread(ReversePInvokeFrame*)
          11: 0xccb5e1 - <unknown>!RhpReversePInvokeAttachOrTrapThread2
          12: 0xccb658 - <unknown>!RhpReversePInvoke
          13: 0x13a86eb - Adder_wit_computer_Intrinsics__cabi_realloc
```

As a workaround, if you first call any other export that only deals with `int` or other non-heap-objects, that will initialize the runtime correctly, and then `cabi_realloc` will no longer crash so you can then call exports that accept `string`.

_Originally posted by @SteveSandersonMS in https://github.com/bytecodealliance/wit-bindgen/issues/760#issuecomment-1834333327_

esoterra:
[...]

PR #791, opened 2023-12-15
c#: Update to use latest packages incorporating WasmImportLinkage

yowl:
This PR removes the `WasmImport` tag from the project file now that the `WasmImportLinkageAttribute` has been accepted upstream and merged into Ilc.  Note that it is not in the published dotnet SDK yet so we need to add the attribute to the generated code until we update to dotnet 9 SDK (and tooling which supports it).

Also moves `cabi_realloc` to c code, fixing #777  .  I did try a few things to avoid adding C code and invoking emcc to compile that code to a wasm object file.

cc @jsturtevant @silesmo


crates/csharp/src/lib.rs
```diff
@@ -535,10 +530,17 @@ impl WorldGenerator for CSharp {
                 "// Generated by `wit-bindgen` {version}. DO NOT EDIT!
                 {CSHARP_IMPORTS}
 
-                namespace {namespace}.{name};
+                // temporarily add this attribute until it is available in dotnet 9
+                namespace System.Runtime.InteropServices
```
jsturtevant, line 534: This is in a for loop so gets printed twice in some cases.  

I think we could move put this where `internal static class Intrinsics` was so it only gets printed once

tests/runtime/main.rs
```diff
@@ -541,30 +541,36 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
...
-            }
-
             // Copy test file to target location to be included in compilation
             let file_name = path.file_name().unwrap();
             fs::copy(path, out_dir.join(file_name.to_str().unwrap()))?;
 
             csproj.generate()?;
 
+            // generate the cabi_realloc (and TODO: release)
+            let mut cmd_emcc = Command::new("emcc.bat");
```
jsturtevant, line 551: maybe this can be a task in the csproj.cs that builds if that c file is present?

jsturtevant:
do we need bump the dotnet AOT compiler version anywhere?

crates/csharp/src/lib.rs
yowl: Thanks, Yes I have some more changes pending, but until the packages are published https://github.com/dotnet/runtimelab/pull/2462, then I've left this as draft

yowl:
> do we need bump the dotnet AOT compiler version anywhere?

Yes we do.  Incoming, when the packages publish.

tests/runtime/main.rs
yowl: So that it builds as part of the user `dotnet publish` ?

jsturtevant: that's what I was thinking, since its a file we generate and it needs to be compiled to be usable component

yowl: Thanks, added to the csproj.

PR #792, opened 2023-12-15
chore(go): remove unused wit-component dep

vados-cosmonic:
Thanks for all the awesome work on `wit-bindgen` -- in trying to debug some issues with generation (basically, I was trying to find where newer-than-expected WIT definitions might have come from), I found that the `wit-component` dependency in `wit-bindgen-go` doesn't seem to be necessary.

Please feel free to correct me if I've misunderstood somehow!


alexcrichton (approved):
Thanks!

PR #794, opened 2023-12-16
feat(c#): Multi-return support

jsturtevant:
This adds the codegen test for multi-return with c#. It creates an interface like:

```csharp
namespace wit_the_world.wit.exports.foo.foo.MultiReturn;

public interface IMultiReturn {
    static abstract void Mra();
    
    static abstract void Mrb();
    
    static abstract uint Mrc();
    
    static abstract uint Mrd();
    
    static abstract Tuple<uint, float> Mre();
    
}
```

and export interop looks like:

```csharp
[UnmanagedCallersOnly(EntryPoint = "foo:foo/multi-return#mre")]
    public static int wasmExportMre() {
        
        Tuple<uint, float> result =  MultiReturnImpl.Mre();
        uint result0 = result.Item1;
        float result1 = result.Item2;
        
        var ptr = returnArea.AddrOfBuffer();
        returnArea.SetS32(0, unchecked((int)(result0)));
        returnArea.SetF32(4, result1);
        return (int)(ptr);
        
    }
```

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -675,9 +675,9 @@ impl InterfaceGenerator<'_> {
                         return BitConverter.ToInt32(span);
                     }}
 
-                    internal float GetF32(IntPtr ptr, int offset)
+                    internal static float GetF32(IntPtr ptr, int offset)
                     {{
-                        var span = new Span<byte>((void*)ptr, {0});
+                        var span = new Span<byte>((void*)ptr, 4);
                         return BitConverter.ToSingle(span.Slice(offset, 4));
                     }}
 
@@ -1146,7 +1146,7 @@ impl InterfaceGenerator<'_> {
             count => {
                 self.gen.tuple_counts.insert(count);
                 format!(
-                    "Tuple<{}>",
+                    "({})",
                     func.results
                         .iter_types()
                         .map(|ty| self.type_name_boxed(ty, qualifier))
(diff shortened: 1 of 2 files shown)


crates/csharp/src/lib.rs
```diff
@@ -601,7 +608,14 @@ impl InterfaceGenerator<'_> {
...
                 {{
-                    private int GetS32(IntPtr ptr, int offset)
+                    internal int GetS32(IntPtr ptr, int offset)
+                    {{
+                        var span = new Span<byte>((void*)ptr, {});
+
+                        return BitConverter.ToInt32(span.Slice(offset, 4));
+                    }}
+
+                    internal int GetF32(IntPtr ptr, int offset)
```
yowl, line 618: Should this be returning a float ?

silesmo: Pushed a fix

jsturtevant: Thanks!  I haven't had a chance to get back to this but will try to in the next couple of days

yowl:
There's a bit of overlap here with #808  .  I went with the `(type1, type2, ..)` syntax rather than `Tuple`, thinking that at some point we could name the elements rather than just `Item1`, `Item2` etc.

jsturtevant:
@yowl  I've rebased on top of your changes for #808 and we are good to go.  Thanks @silesmo for the float fix too!

PR #799, opened 2023-12-23
fix(readme): add semicolon to WITs

### LINKED ISSUE #796: Example WIT file in readme needs semicolons
https://github.com/bytecodealliance/wit-bindgen/blob/04b715a08aa3859cf8941eb4fd3358db6bf6d9d7/README.md#L51-L57

```wit
package example:host;

world host {
  import print: func(msg: string);

  export run: func();
}
```

i got a syntax erroor trying the rust demo and this was the cause

Mossaka:
Close #796


alexcrichton (approved):
Thanks!

PR #800, opened 2023-12-23
C#: Add implementation for lowering and lifiting flags

yowl:
.. and enable the corresponding codegen test.  Implemented in c# as `enum`s.

(contributes to making the `records` runtime test pass)

cc @silesmo @jsturtevant

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -1164,14 +1164,6 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
 
         let name = name.to_upper_camel_case();
 
-        let ty = match flags.repr() {
-            FlagsRepr::U8 => "byte",
-            FlagsRepr::U16 => "ushort",
-            FlagsRepr::U32(1) => "uint",
-            FlagsRepr::U32(2) => "ulong",
-            repr => todo!("flags {repr:?}"),
-        };
-
         let enum_elements = flags
             .flags
             .iter()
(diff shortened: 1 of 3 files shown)


crates/csharp/src/lib.rs
```diff
@@ -1175,24 +1175,22 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
...
                 };
-                format!(
-                    "public static readonly {name} {flag_name} = new {name}(({ty}) (1{suffix} << {i}));"
-                )
+                format!("{flag_name} = 1{suffix} << {i},")
             })
             .collect::<Vec<_>>()
             .join("\n");
 
+        let enum_type = if matches!(flags.repr(), FlagsRepr::U32(2)) {
```
silesmo, line 1183: Shouldn't this be set to the corresponding type based on the number of flags as in the specification as seen here? https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#alignment

tests/README.md
```diff
@@ -25,6 +25,12 @@ To run just `codegen` tests for a single language (replace rust with language of
 cargo test -p wit-bindgen-rust
 ```
 
+To run just `codegen` tests for a single language (replace rust with language of choice: `go`, `c`, `csharp`, etc.) and a single wit file (replace `flags` with whatever wit file should be tested):
```
silesmo, line 28: Duplicated in the text below?

yowl: It is, but this is the specific instruction for running a `codegen` test for a single language and single wit, which wasn't covered explicitly elsewhere.

crates/csharp/src/lib.rs
yowl: Thanks, updated to use `ushort` and `byte`.

PR #803, opened 2024-01-03
c: Automatically drop borrows of external resources in exports

elliottt:
Instead of requiring that the embedder know when it's necessary to drop a borrow of a resource, automatically insert them where necessary. As this isn't always possible to automatically drop borrows with this implementation, the behavior can be disabled with the `--autodrop-borrows=no` flag, going back to the current state of requiring explicit drops for borrows of imported resources given to exports.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -72,6 +72,10 @@ pub struct Opts {
     /// the component type.
     #[cfg_attr(feature = "clap", arg(long))]
     pub type_section_suffix: Option<String>,
+
+    /// Disable the autodropping of borrows in exports.
+    #[cfg_attr(feature = "clap", arg(long, default_value_t = false))]
+    pub no_export_autodrop: bool,
 }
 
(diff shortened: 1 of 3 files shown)


crates/c/src/lib.rs
```diff
@@ -63,6 +64,10 @@ pub struct Opts {
     /// Rename the interface `K` to `V` in the generated source code.
     #[cfg_attr(feature = "clap", arg(long, name = "K=V", value_parser = parse_rename))]
     pub rename: Vec<(String, String)>,
+
+    /// Disable the autodropping of borrows in exports.
+    #[cfg_attr(feature = "clap", arg(long, default_value_t = false))]
+    pub no_export_autodrop: bool,
```
alexcrichton, line 70: Could this perhaps be phrased in the positive such as `export_autodrop: Option<bool>`?

alexcrichton (commented):
Seems reasonable to me!

alexcrichton (approved):

PR #806, opened 2024-01-05
fix(C): Resource name in import should be the original name

kateinoigakukun:
The wrong import name makes an invalid component and `wasm-tools component new` fails

```
    0: import of `[resource-rep]rb_abi_value` is not a valid resource function
```


alexcrichton (approved):
Thanks! Would it be possible to add a test for this as well?

kateinoigakukun:
While writing tests, I found tinygo bindgen also had identifier casing issues. Fixed them also and added tests that kill those issues.

alexcrichton (approved):
Thanks!

PR #807, opened 2024-01-07
(C#) -  Basic types, Lists, Results & Variants

silesmo:
Adds in support the following types:
- Lists
- Results
- Variants
Also converts the C# option type to use generics as it caused issues where the same option type could be generated multiple times before.

We still seem to have an issue with return pointer alignment that will be addressed in a separate PR.


crates/csharp/src/lib.rs
```diff
@@ -635,20 +657,80 @@ impl InterfaceGenerator<'_> {
...
+                    unsafe internal static T[] GetArray<T>(IntPtr ptr, int length)
+                    {{
+                        var span = new Span<byte>((void*)ptr, sizeof(T)*length);
+        
+                        var array = new T[length];
+        
+                        fixed (T* dstPtr = &array[0])
+                        {{
+                            //TODO: Optimize this. The ToArray call is not needed and does an additional heap allocation.
+                            Marshal.Copy(span.ToArray(), 0, (IntPtr)dstPtr, length);
```
pavelsavara, line 721: maybe you can use `Copy(nint source, byte[] destination, int startIndex, int length);` 
without the need for `var span = new Span<byte> ... ` and `fixed` at all

jsturtevant: I ended up with `span.CopyTo(new Span<T>(array));`  I believe this addresses the additional heap allocation.  I kept Span<T> since it was useful for handling all the different lengths of types

jsturtevant:
quick update, I've got the lists runtime tests passing 🚀 

 I need to clean up quite a bit of the code as I was learning as I was doing and there are a lot of places that can be made consistent and more efficient.  The other thing to finish up is add logic for cleaning up allocated heap memory.

jsturtevant:
I've done the clean up and added logic to free the memory.  Ready for Review

silesmo:
Added in variants and results. There still seems to be an issue with return pointer alignment. Haven't found where it's happening yet.

crates/csharp/src/lib.rs
```diff
@@ -1985,17 +2000,74 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
+                        let buffer = self.locals.tmp("buffer");
+                        let gc_handle = self.locals.tmp("gcHandle");
+                        uwrite!(
+                            self.src,
+                            "
+                        byte[] {buffer} = new byte[{op}.Length];
+                        Buffer.BlockCopy({op}, 0, {buffer}, 0, {op}.Length);
+                        var {gc_handle} = GCHandle.Alloc({op}, GCHandleType.Pinned);
+                        var {address} = {gc_handle}.AddrOfPinnedObject();
+                        Marshal.Copy({buffer}, 0, (IntPtr){address}, {op}.Length);
```
yowl, line 2040: Why do we need 2 copies, the `BlockCopy` and the `Marshal.Copy` ?  Shouldn't we be pinning `buffer` and using its address for `address`, and delete the `Marshal.Copy` ?

jsturtevant: I thought the same thing, but when I tried to clean this up it wouldn't get the correct values.  I can try it again...

jsturtevant: Not sure what I did last time but I removed the `Marshal.Copy` and it is working still

crates/csharp/src/lib.rs
```diff
@@ -2018,23 +2089,89 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
+                let ty = self.gen.type_name(element);
+                let index = self.locals.tmp("index");
+                let list: String = self.locals.tmp("list");
+
+                let buffer: String = self.locals.tmp("buffer");
+
+                uwrite!(
+                    self.src,
+                    "
+                    void* {buffer} = stackalloc byte[{size} + {align} * ({op}).Count];
```
yowl, line 2118: I think we need to allocate this buffer from the heap.  When this method returns the buffer is going to be dangling and in danger or being overwritten.

jsturtevant: Since this is on the import direction I thought the stack was OK? We are doing this for the return area as well

crates/csharp/src/lib.rs
```diff
@@ -1673,8 +1529,159 @@ impl<'a, 'b> FunctionBindgen<'a, 'b> {
...
+
+        uwrite!(
+            self.src,
+            r#"
+            {ty} {lifted};
+
+            switch ({op}) {{
+                {cases}
+
+                default: throw new ArgumentException($"invalid disciminant: {{{op}}}");
```
yowl, line 1678:
```suggestion
                default: throw new ArgumentException($"invalid discriminant: {{{op}}}");
```

crates/csharp/src/lib.rs
```diff
@@ -1082,7 +963,7 @@ impl InterfaceGenerator<'_> {
             self.csharp_interop_src,
             r#"
             [UnmanagedCallersOnly(EntryPoint = "{export_name}")]
-            public static {wasm_result_type} {interop_name}({wasm_params}) {{
+            unsafe public static {wasm_result_type} {interop_name}({wasm_params}) {{
```
yowl, line 966:
```suggestion
            public static unsafe {wasm_result_type} {interop_name}({wasm_params}) {{
```
Also, can we drop `unsafe` ?

crates/csharp/src/lib.rs
yowl: Have I marked the wrong rust code.  I meant to mark what generates this export:
```cs
        unsafe public static int wasmExportListResult3() {
            var ret = TestImpl.ListResult3();
            
            void* buffer = stackalloc byte[8 + 4 * (ret).Count];
            var list = ret;
            
            for (int index = 0; index < list.Count; ++index) {
                string element = list[index];
                int basePtr = (int)buffer + (index * 8);
                
                var result = element;
                IntPtr interopString = InteropString.FromString(result, out int lengthresult);
                BitConverter.TryWriteBytes(new Span<byte>((void*)(basePtr + 4), 4), unchecked((int)lengthresult));
                BitConverter.TryWriteBytes(new Span<byte>((void*)(basePtr + 0), 4), unchecked((int)interopString.ToInt32()));
                
            }
            BitConverter.TryWriteBytes(new Span<byte>((void*)(ptr + 4), 4), unchecked((int)list.Count));
            BitConverter.TryWriteBytes(new Span<byte>((void*)(ptr + 0), 4), unchecked((int)(int)buffer));
            return ptr;
            
        }
```

jsturtevant: In this case we are using [`BitConverter.TryWriteBytes`](https://learn.microsoft.com/en-us/dotnet/api/system.bitconverter.trywritebytes?view=net-8.0) which will write the value to the destination provided at ptr, so we are writing the value onto the returnArea which is on the heap so we should be ok.

crates/csharp/src/lib.rs
jsturtevant: > Also, can we drop unsafe ?

Not currently, a few of these still use pointers directly, we might able to clean that up by using Spans and other techniques at a later point.

crates/csharp/src/lib.rs
yowl: But aren't we writing the address of `buffer` which is a stack location into the return area?

jsturtevant: I've updated this, I think I mis-understood the return pointer scenario and this scenario which aren't the same.

jsturtevant (approved):

PR #809, opened 2024-01-09
c: Use drop-flags when auto-dropping borrows

elliottt:
To expand the cases where auto-dropping is possible in wit-bindgen-c, use drop flags to indicate when borrows need to be dropped. This enables us to also auto-drop borrows present in optional and variant types, leaving only lists as the outlier for values that cannot have their borrows automatically dropped.

This implementation takes advantage of `0` being a newly reserved resource id, allowing us to use that value to indicate when the borrow wasn't observed during lifting.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -1967,6 +1967,11 @@ impl InterfaceGenerator<'_> {
     }
 }
 
+struct DroppableBorrow {
+    name: String,
+    ty: TypeId,
+}
+
 struct FunctionBindgen<'a, 'b> {
     gen: &'a mut InterfaceGenerator<'b>,
     locals: Ns,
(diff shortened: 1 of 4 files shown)


alexcrichton:
Actually I pushed up a fix for that specific issue to this PR so if CI is green I'll merge.

crates/c/src/lib.rs
```diff
@@ -2264,8 +2273,13 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                         if !self.gen.in_import && self.gen.autodrop_enabled() {
                             // Here we've received a borrow of an imported resource, which is the
                             // kind we'll need to drop when the exported function is returning.
-                            let target = dealias(self.gen.resolve, *id);
-                            self.borrows.push((op.clone(), target));
+                            let ty = dealias(self.gen.resolve, *id);
+
+                            let name = self.locals.tmp("borrow");
+                            uwriteln!(self.borrow_decls, "int32_t {name} = 0;");
```
elliottt, line 2279: Is there a better way to refer to the type of a handle here, or is using `int32_t` directly fine?

alexcrichton: Nah int32_t is fine, I don't think there's anything "more official" at this time

PR #813, opened 2024-01-10
c and go: only include WIT package version in type and function names if multiple versions of the package exist

tschneidereit:
Most projects won't ever mix multiple versions of the same package, so have no need to include the version number in generated names. Instead, they'd include different headers to use different bindings for the different versions.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -765,14 +765,21 @@ fn interface_identifier(
             ns.push_str("_");
             ns.push_str(&pkg.name.name.to_snake_case());
             ns.push_str("_");
-            if let Some(version) = &pkg.name.version {
-                let version = version
-                    .to_string()
-                    .replace('.', "_")
-                    .replace('-', "_")
-                    .replace('+', "_");
-                ns.push_str(&version);
-                ns.push_str("_");
+            let pkg_has_multiple_versions = resolve.packages.iter().any(|(_, p)| {
+                p.name.namespace == pkg.name.namespace
+                    && p.name.name == pkg.name.name
+                    && p.name.version != pkg.name.version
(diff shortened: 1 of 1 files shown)


alexcrichton:
Instead of having a user-configurable flag for this, could the C generator copy [this logic](https://github.com/bytecodealliance/wit-bindgen/blob/89f6d36d09985956be9007b2d8df61514022d9ee/crates/rust/src/lib.rs#L313-L356) (more or less) from the Rust generator to automatically omit versions and only include them if disambiguation is necessary?

tschneidereit:
> Instead of having a user-configurable flag for this, could the C generator copy [this logic](https://github.com/bytecodealliance/wit-bindgen/blob/89f6d36d09985956be9007b2d8df61514022d9ee/crates/rust/src/lib.rs#L313-L356) (more or less) from the Rust generator to automatically omit versions and only include them if disambiguation is necessary?

Oh, that is indeed much nicer!

@Mossaka this change also makes the Go generator omit the version in type/function names if only one version of a package exists. Let me know if you think that's not a good change, and we can change it.

alexcrichton (approved):

Mossaka:
> @Mossaka this change also makes the Go generator omit the version in type/function names if only one version of a package exists

I don't see how the Go generator names are changed. In fact no changes are made to the Go crate. Am I missing anything?

tschneidereit:
> I don't see how the Go generator names are changed. In fact no changes are made to the Go crate. Am I missing anything?

Ah, if this doesn't change the output of the Go generator, then never mind :) I thought it would, because the Go generator invokes the functions I changed here. But maybe that's just internal to the bindings, and the Go-exposed names don't change?

PR #818, opened 2024-01-16
use "<a></a>Content" instead of "<a>Content</a>"

jcbhmr:
Why this is a good idea:
- Makes it much easier to apply classless CSS to the resulting HTML
- Conform with ecosystem convention

All of these classless CSS frameworks style all `<a>` tags the same. https://github.com/dbohdan/classless-css

ALL OF THEM treat `<a>` as "a link" even if it has no href. This is usually what you want since you might apply an `onclick` that you want to appear as a phantom link. This isn't the case with `<a name="...">` in the generated HTML from `wit-bindgen markdown`. I suggest instead following a precedent of empty `<a name="...">` tags.

This `<a name=""></a>` thing has been done before. Quite a lot actually. https://github.com/search?q=%2F%3Ca%20name%3D%5B%22%27%5D.%2B%5B%22%27%5D%3E%3C%5C%2Fa%3E%2F&type=code

Consider:

Before:
image

After:
image

Imagine a whole page full of phantom links and you can see why I want to make this change.

This would still preserve the `page.html#the-thing` of the `<a name="the-thing"></a>` while making it compatible with the vast majority of CSS that already exists in the ecosystem of classless whole-page CSS presets.

Yes, you can solve this by using `:any-link { ... }` https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link but realistically I'm not going to open 100+ PRs on projects that are one-and-done and don't get any more attention -- CSS presets rarely seem to get updated after they are pretty enough. It's much easier just to tweak the generation here to conform to ecosystem expectations.

Alternatives:
- Add custom CSS to the generated HTML that accounts for this. This could be as basic as forking one of the existing classless CSS projects and tweaking it and then using the resulting style.css in the generated HTML. But that's more in wit-doc territory, not `wit-bindgen markdown`.

### DIFF
--- a/crates/markdown/src/lib.rs
+++ b/crates/markdown/src/lib.rs
@@ -121,7 +121,7 @@ impl WorldGenerator for Markdown {
         let name = resolve.name_world_key(name);
         uwriteln!(
             self.src,
-            "## <a name=\"{}\">Import interface {name}</a>\n",
+            "## <a name=\"{}\"></a>Import interface {name}\n",
             name.to_snake_case()
         );
         self.hrefs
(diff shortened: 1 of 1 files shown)


pchickey (approved):
Thanks for the fix and the explanation.

PR #819, opened 2024-01-18
[C#] Support for  options

yowl:


yowl:
Merged by mistake, sorry, have reverted main back to where it was.

PR #820, opened 2024-01-18
[c#] Support for options

yowl:
This PR adds support to c# codegen for `option<>`

Redoes #819 

A large change for just enabling options, but this also implies adding support for anonymous types.  I also hit problems with namespace clashes, so reworked some of the namespace logic to upper camel case more names.

I've largely copied Java for the lower/lifting, but we have a class for each option type.  We cannot use `Nullable<T>` in C# because in wit `option<option...` is possible, but `Nullable` is only available over value types so `Nullable<Nullable` is not valid.  A generic `Option<T>` should be possible as a code size optimisation.

Split the options codegen test so that simple options can be brought up without support for `result`.

cc @alexcrichton  for the the split of the option test


alexcrichton:
Splitting the test up looks good to me! I'll leave it to the others for the C# review too

crates/csharp/src/lib.rs
```diff
@@ -207,7 +224,7 @@ impl WorldGenerator for CSharp {
         _files: &mut Files,
     ) -> Result<()> {
         let name = &format!("{}-world", resolve.worlds[world].name);
-        let mut gen = self.interface(resolve, name, false);
+        let mut gen = self.interface(resolve, name, false, false);
```
jsturtevant, line 227: could we alias the true/false to something?  It is hard to know what the intent is here

jsturtevant (commented):
LGTM, Thanks for improving the namespacing!

crates/csharp/src/lib.rs
yowl: good call, in fact the second one was no longer used, and we can use the existing `Direction` for the first one.  Have made that change

yowl: edit, ok, that flag is needed, have made them both enums to make the code a bit more readable.

yowl:
Hi, if this is ok, can @jsturtevant  or @silesmo approve it?

jsturtevant (approved):

yowl:
Thanks!

PR #821, opened 2024-01-20
feature(c#): Adds enum lifting and lowering

jsturtevant:
Adds support for enums in c#.  It also adds a simple code-gen test for enums which makes it easy to validate.

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -1972,16 +1972,17 @@ impl Bindgen for FunctionBindgen<'_, '_> {
             Instruction::EnumLower { .. } => results.push(format!("(int){}", operands[0])),
 
             Instruction::EnumLift { ty, .. } => {
-                let t = self.gen.type_name(&Type::Id(*ty));
+                let t = self.gen.type_name_with_qualifier(&Type::Id(*ty), true);
                 let op = &operands[0];
                 results.push(format!("({}){}", t, op));
 
                 uwriteln!(
                     self.src,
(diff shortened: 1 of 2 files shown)


jsturtevant:
Do we need to protect against invalid integer values for the enums or is that covered via the parser/abi?

yowl:
https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#despecialization

I don't know if I undestand this correctly but it seems enums can be `None` ?

alexcrichton:
You should be able to rely on all the values here being trusted and valid in a purist sense. It's not possible to call an API with a nonexistent enum discriminant with a standards-conformant runtime. If C# has an equivalent for debug assertions I'd recommend debug asserting that the discriminant is in bounds, but otherwise you're good to assume it's always valid.

(I think I'm answering the question here but I'm also not 100% sure what you mean by `None` handling)

yowl:
Basically I was reading that link and here 

image


I saw "None", and wasn't sure what it meant.

alexcrichton:
Ah that "None" means there's no type payload, so the canonical ABI of `enum foo { a, b }` is the same as `variant foo { a, b }` (since the `a` and `b` cases of `variant foo` have no payload)

jsturtevant:
> @jsturtevant do you want to add the `Debug.Assert` for `Enum.IsDefined` maybe or leave it as a TODO?

I rebased on top of options work and added debug statement

jsturtevant:
Thanks for the clarification on both points @alexcrichton

yowl (approved):
Thanks!

PR #827, opened 2024-01-31
Add runtime test for options

vyta:
Currently tests for options are embedded in other tests, this adds a dedicated runtime test for options.

cc: @jsturtevant

### DIFF
--- a/crates/test-rust-wasm/Cargo.toml
+++ b/crates/test-rust-wasm/Cargo.toml
@@ -45,6 +45,10 @@ name = "results"
 test = false
 
 [[bin]]
+name = "options"
+test = false
+
+[[bin]]
 name = "owning"
 test = false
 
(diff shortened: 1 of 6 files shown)


jsturtevant (commented):
LGTM, but need to run `cargo fmt`.  

@alexcrichton for context, when we were adding `Option` support to c#(https://github.com/bytecodealliance/wit-bindgen/pull/820) I noticed we didn't have a way to run any runtime tests since we don't fully support all the types where options are used.  I thought having a test like this makes it easier to adopt types in iterative fashion for future languages and have runtime tests not just code-gen.

alexcrichton:
Sounds good to me! Definitely feel free to add whatever tests you like and if they only have one C# guest that's no problem, other guests can fill in implementations as necessary over time.

jsturtevant (approved):

PR #828, opened 2024-01-31
Add a debugging utility for Rust `generate!`

alexcrichton:
Lift the approach I added in bytecodealliance/wasmtime#7794


jsturtevant:
Super useful!  I guess this could be used instead of `cargo expand`? Maybe can update the readme where `cargo expand` is mentioned to called this option out?

Mossaka (approved):
lgtm!

alexcrichton:
Good point @jsturtevant! Looks like the README could use an update on a few other dimensions as well so I'll follow-up with that.

Issue #841, opened 2024-02-14
rust: Generate an error when a `with` key is not used


alexcrichton:
There's been some confusion around `with` historically and I know I've typo'd things before. There's a failure mode where if a `with` key isn't used, either because it's configured incorrectly or there's a typo, `wit-bindgen` won't actually generate any warnings or errors and the resulting generated code probably isn't what's expected. This unexpected code generation can lead to surprising error messages down the road too.

Ideally the `wit-bindgen` Rust generator would generate an error message for any configuration in `with` which ended up not being used after all bindings were generated. That could help move the error sooner and perhaps weed out these sorts of errors more quickly.

sehz:
Some more doc would be helpful

alexcrichton:
Fixed in https://github.com/bytecodealliance/wit-bindgen/pull/842, @sehz mind filing a separate issue for improving documentation?

sehz:
Done.  https://github.com/bytecodealliance/wit-bindgen/issues/847

PR #842, opened 2024-02-14
Require the use of remappings provided via --with

fibonacci1729:
This is a quick first attempt to address #841.

Alternatively I could update the `Generator` so that `finish` returns an error to eliminate panicking but that would require updating all the language generators for a change currently isiolated to the rust generator.

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -48,6 +48,9 @@ struct RustWasm {
     resources: HashMap<TypeId, ResourceInfo>,
     import_funcs_called: bool,
     with_name_counter: usize,
+    // Track the with options that were used. Remapped interfaces provided via `with`
+    // are required to be used.
+    used_with_opts: HashSet<String>,
 }
 
 #[cfg(feature = "clap")]
@@ -288,6 +291,7 @@ impl RustWasm {
         let with_name = resolve.name_world_key(name);
         let entry = if let Some(remapped_path) = self.opts.with.get(&with_name) {
             let name = format!("__with_name{}", self.with_name_counter);
+            self.used_with_opts.insert(with_name);
             self.with_name_counter += 1;
             uwriteln!(self.src, "use {remapped_path} as {name};");
             InterfaceName {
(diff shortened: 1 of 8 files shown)


crates/rust/src/lib.rs
```diff
@@ -582,6 +586,16 @@ impl WorldGenerator for RustWasm {
...
         files.push(&format!("{module_name}.rs"), src.as_bytes());
+
+        let remapping_keys = self.opts.with.keys().cloned().collect::<HashSet<String>>();
+
+        let unused_keys = remapping_keys
+            .difference(&self.used_with_opts)
+            .collect::<Vec<&String>>();
+
+        if unused_keys.len() > 0 {
+            panic!("unused remappings provided via `with`: {unused_keys:?}")
```
fibonacci1729, line 597: I'm not married to the language in this panic so please feel free to suggest something better.

crates/rust/src/lib.rs
```diff
@@ -582,6 +586,16 @@ impl WorldGenerator for RustWasm {
 
         let module_name = name.to_snake_case();
         files.push(&format!("{module_name}.rs"), src.as_bytes());
+
+        let remapping_keys = self.opts.with.keys().cloned().collect::<HashSet<String>>();
+
+        let unused_keys = remapping_keys
+            .difference(&self.used_with_opts)
+            .collect::<Vec<&String>>();
```
alexcrichton, line 594: To ensure a deterministic error message, could you sort this list?

crates/rust/src/lib.rs
alexcrichton: Happy to start here and iterate!

Would you be up for a bit of refactoring though to thread around a `Result` instead of panicking here? That I think would provide a better experience for macro users since they wouldn't have to worry about getting a whole stack trace.

alexcrichton (commented):
Thanks for this! This repository isn't really equipped to test this right now (e.g. via something like the `trybuild` crate) but that's ok. Can you confirm locally though that the error message looks reasonable when using the `generate!` macro?

crates/rust/src/lib.rs
fibonacci1729: Absolutely! I actually started there but reduced to this.

crates/rust/src/lib.rs
fibonacci1729: Good catch!

fibonacci1729:
> Thanks for this! This repository isn't really equipped to test this right now (e.g. via something like the `trybuild` crate) but that's ok. Can you confirm locally though that the error message looks reasonable when using the `generate!` macro?

Manually confirmed locally the error message looks reasonable when failing with the `generate!` macro:
Screenshot 2024-02-16 at 10 29 27 AM

alexcrichton (approved):
Thanks!

PR #844, opened 2024-02-16
Build a new release process for this repository

alexcrichton:
Like with the wasm-tools repository in bytecodealliance/wasm-tools#1416, this repository is going to follow suit with the same release process. Notably everything should now be largely automated in terms of publication and such so anyone can make a release so long as they have write access to the repository.


alexcrichton:
I'll note that my plan here, like with https://github.com/bytecodealliance/wasm-tools/pull/1416, is to perform a release right after this merges.

Unlike with `wasm-tools` however I have performed no testing of this on my own repos yet. I'm crossing my fingers that I copied everything over from `wasm-tools` correctly and figure I can try to clean up any issues afterwards.

pchickey (approved):

PR #851, opened 2024-02-20
rust: Define `cabi_realloc` as a weak symbol

### LINKED ISSUE #849: Compilation error when multiple versions of `wit-bindgen` present in project
When there are multiple versions of `wit-bindgen` present in a project compilation fails with an error like:
```
  = note: rust-lld: error: duplicate symbol: cabi_realloc
          >>> defined in /Volumes/workplace/wit-bindgen-symbol-repro/target/wasm32-wasi/release/deps/libwit_bindgen-796469586291d0cd.rlib(wit_bindgen-796469586291d0cd.wit_bindgen.3ccd0498004a9c9c-cgu.0.rcgu.o)
          >>> defined in /Volumes/workplace/wit-bindgen-symbol-repro/target/wasm32-wasi/release/deps/libwit_bindgen-fb8c62ead352e205.rlib(wit_bindgen-fb8c62ead352e205.wit_bindgen.15338958f611d1dd-cgu.0.rcgu.o)
```

In my case this seems to be because I am consuming the `wasi` crate which has been published with the Preview2 bindings built on `wit-bindgen = "0.16.0"`. So my bindings must also be generated with `wit-bindgen = "0.16.0"`. Hopefully this can be fixed in a way that avoids having to coordinate versions of `wit-bindgen` across all dependencies. 

I have a [minimal repro](https://github.com/landonxjames/wit-bindgen-symbol-repro) available, toggling the version of `wit-bindgen` in the Cargo.toml will let you observe the behavior.

alexcrichton:
This commit updates the `wit-bindgen` Rust crate to define the `cabi_realloc` symbol as a weak symbol. This is not easy because Rust does not offer a stable means by which to do this. Despite this through a combination of shell scripts and CI it should be possible to get this all working together by using C to define a weak symbol calling a known Rust symbol.

Closes #849


crates/guest-rust/src/cabi_realloc.c
```diff
@@ -0,0 +1,8 @@
+#include <stdint.h>
+
+extern void *cabi_realloc_wit_bindgen_0_18_0(void *ptr, size_t old_size, size_t align, size_t new_size);
+
+__attribute__((__weak__, __export_name__("cabi_realloc")))
+void *cabi_realloc(void *ptr, size_t old_size, size_t align, size_t new_size) {
+  return cabi_realloc_wit_bindgen_0_18_0(ptr, old_size, align, new_size);
+}
```
sunfishcode, line 10: If we're doing the `_0_18_0`  mangling anyway, would it be simpler to teach all the places that name `cabi_realloc` to name `cabi_realloc_0_18_0` instead, rather than using a precompiled archive to get a weak symbol?

alexcrichton: Right now that's a hardcoded convention in `wasm-tools component new`, there's nothing in the actual core wasm binary produce that refers to `cabi_realloc` at all currently.

sunfishcode (approved):
This should work for now. There are build environments where people won't want to bring in precompiled binaries, but we can land this until we have other solutions.

PR #855, opened 2024-02-21
rust: Fix cases where `with` provides resources

### LINKED ISSUE #832: rust: `with` not working as expected
Hello,

I'm trying to use the `with` option to reference existing code for wasi interfaces. For example:

```rust
wit_bindgen::generate!(
    world: "my-world",
    exports: {
        world: MyWorld,
    },
    runtime_path: "::my_lib::wit_bindgen::rt",
    with: {
        "wasi:io/error": ::my_lib::wit::wasi::io::error,
        "wasi:io/streams": ::my_lib::wit::wasi::io::streams,
        "wasi:io/poll": ::my_lib::wit::wasi::io::poll,
        "wasi:clocks/monotonic-clock": ::my_lib::wit::wasi::clocks::monotonic_clock,
        "wasi:http/types": ::my_lib::wit::wasi::http::types,
        "wasi:http/outgoing-handler": ::my_lib::wit::wasi::http::outgoing_handler,
    },
);
```

However, this doesn't seem to do what I expect. That is, the generated code includes code for `wasi` instead of aliasing the modules specified in `with`. Note that the same behavior occurs with the CLI.

Please let me know if I can provide any additional context, or if I'm missing something.

Thanks!

alexcrichton:
This commit fixes a bug where if `with` was used to provide a resource then the internal `resources` map wasn't updated correctly. This could lead to "index not found" panics in #832. To fix this I've opted to remove the `resources` map entirely as the `owned` field was already not tracked and this additionally brings the Wasmtime and Rust generators a bit closer together.

Instead now each `InterfaceId` has an entry of whether it was last seen as an export or an import and this is consulted when determining whether a resource is exported or not.

This fix then required shuffling around where the stub implementations are generated to ensure that when the stubs for top-level exported functions are generated the right metadata of "was this interface last seen as an import" is generated correctly.

Closes #832

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -9,23 +9,11 @@ use std::process::{Command, Stdio};
 use std::str::FromStr;
 use wit_bindgen_core::abi::{Bitcast, WasmType};
 use wit_bindgen_core::{
-    uwriteln, wit_parser::*, Direction, Files, InterfaceGenerator as _, Source, Types,
-    WorldGenerator,
+    uwriteln, wit_parser::*, Files, InterfaceGenerator as _, Source, Types, WorldGenerator,
 };
 
 mod bindgen;
 mod interface;
-
-#[derive(Default)]
-struct ResourceInfo {
-    // Note that a resource can be both imported and exported (e.g. when
-    // importing and exporting the same interface which contains one or more
-    // resources).  In that case, this field will be `Import` while we're
-    // importing the interface and later change to `Export` while we're
-    // exporting the interface.
-    direction: Direction,
-    owned: bool,
-}
 
 struct InterfaceName {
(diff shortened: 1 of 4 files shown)


crates/rust/src/lib.rs
```diff
@@ -45,12 +33,13 @@ struct RustWasm {
     export_modules: Vec<(String, Vec<String>)>,
     skip: HashSet<String>,
     interface_names: HashMap<InterfaceId, InterfaceName>,
-    resources: HashMap<TypeId, ResourceInfo>,
+    interface_last_seen_as_import: HashMap<InterfaceId, bool>,
```
pchickey, line 36:
```suggestion
    /// Each imported and exported interface is stored in this map. Value indicates if last use was import.
    interface_last_seen_as_import: HashMap<InterfaceId, bool>,
```

PR #863, opened 2024-02-23
Render the Rust custom section a bit more readably

alexcrichton:
Instead of using a large literal of integers instead use a byte string which is a bit more compact and allows rendering some ascii inline where possible.


pchickey (approved):
Looks good in principle but I can't see the bug that is blowing up the tests.

PR #868, opened 2024-02-26
Move most of the `wit-bindgen` crate into generated code

alexcrichton:
This PR is borne out of discussion on https://github.com/bytecodealliance/cargo-component/issues/236 where the conclusion was that it'd be best if the generated code that `wit-bindgen` emits is largely standalone and doesn't need any extra supporting crate. That enables the `cargo component` generator to not exactly match the `wit-bindgen` crate, if any, going forward.

Given that this PR is the implementation of everything there except the `cabi_realloc` bits. All other runtime bits from the `wit-bindgen` crate have now moved into the generated code itself. Specifically, when needed, a top-level `_rt` module is generated at the macro invocation site. Any references to the runtime now go through that module. This means that `Resource<T>` won't be shared between two invocations of the macro, for example.

I've left behind all the bits in the `wit-bindgen` crate from the current version, 0.19.0, and marked them as "someone please delete them in the future". That should make a future release of this crate as suitable for use with the 0.19.0 release of this generator, meaning that we won't break `cargo component` workflows immediately. Eventually once `cargo component` has upgraded to a release including this PR then the bits can be phased out.

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -1,6 +1,7 @@
 use crate::interface::InterfaceGenerator;
 use anyhow::{bail, Result};
 use heck::*;
+use indexmap::IndexSet;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt::{self, Write as _};
 use std::io::{Read, Write};
@@ -41,6 +42,21 @@ struct RustWasm {
     // are required to be used.
     used_with_opts: HashSet<String>,
     world: Option<WorldId>,
+
+    rt_module: IndexSet<RuntimeItem>,
+}
+
+#[derive(PartialEq, Eq, Clone, Copy, Hash, Debug)]
+enum RuntimeItem {
+    AllocCrate,
+    StringType,
+    StdAllocModule,
+    VecType,
+    StringLift,
+    InvalidEnumDiscriminant,
+    CharLift,
+    BoolLift,
+    CabiDealloc,
 }
 
 #[cfg(feature = "clap")]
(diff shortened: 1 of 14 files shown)


alexcrichton:
I'll also note that there's more discussion on [this Zulip thread](https://bytecodealliance.zulipchat.com/#narrow/stream/407292-cargo-component/topic/wasi.20crate.20.22hello.20world.22.20not.20working/near/423424883).

And finally, for `cabi_realloc`, given how "weird" that is I don't think we can remove it for now. Perhaps the saving grace though will be that with a `wasm32-wasip2` target we can assume that `cabi_realloc` is built into the standard library itself which means we won't have to provide it at all, meaning that at that point the `wit-bindgen` runtime crate isn't needed at all (yay!)

crates/rust/src/lib.rs
```diff
@@ -307,6 +336,388 @@ impl RustWasm {
...
+            }
+
+            RuntimeItem::AsF64 => {
+                self.emit_runtime_as_trait("f64", &["f64"]);
+            }
+
+            RuntimeItem::ResourceType => {
+                self.src.push_str(
+                    r#"
+type RawRep<T> = Option<T>;
```
dicej, line 509: At some point we might want to move this to its own .rs file and `include!` it.  Likewise any other big code literals.

dicej (approved):

crates/rust/src/lib.rs
dicej: `include_str!`, I meant.

PR #870, opened 2024-02-27
Preserve pointer provenance in the Rust backend.

sunfishcode:
Use the new `Pointer` and `Length` types in the Rust backend to emit code that uses `*mut c_void` and `usize` instead of `i32` when working with pointers and array lengths.

To represent `PointerOrI64`, use a `MaybeUninit<u64>`, since that type can hold any `u64` and is [documented](https://github.com/rust-lang/rfcs/blob/master/text/3559-rust-has-provenance.md#reference-level-explanation) to also preserve provenance.

This change happens to get the generated Rust code close to supporting memory64, however it isn't complete; the abi code still emits hard-coded `+ 4` offsets for loading the length of a pointer+length pair in memory, and record layout isn't aware of 64-bit offsets, and possibly other things.

~~Opening as a draft PR, as it depends on https://github.com/bytecodealliance/wasm-tools/pull/1423.~~

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -379,7 +379,7 @@ impl RustWasm {
                 self.rt_module.insert(RuntimeItem::StdAllocModule);
                 self.src.push_str(
                     "\
-pub unsafe fn cabi_dealloc(ptr: *mut ::core::ffi::c_void, size: usize, align: usize) {
+pub unsafe fn cabi_dealloc(ptr: *mut u8, size: usize, align: usize) {
     if size == 0 {
         return;
     }
@@ -1248,7 +1248,7 @@ fn wasm_type(ty: WasmType) -> &'static str {
         WasmType::I64 => "i64",
         WasmType::F32 => "f32",
         WasmType::F64 => "f64",
-        WasmType::Pointer => "*mut ::core::ffi::c_void",
+        WasmType::Pointer => "*mut u8",
         WasmType::Length => "usize",
 
         // `PointerOrI64` can hold either a `u64` or a pointer with provenance.
@@ -1293,7 +1293,7 @@ fn bitcast(casts: &[Bitcast], operands: &[String], results: &mut Vec<String>) {
                 format!(
                     "{{
                         let mut t = ::core::mem::MaybeUnunit::<u64>::uninit();
-                        t.as_mut_ptr().cast::<*mut core::ptr::c_void>().write({});
+                        t.as_mut_ptr().cast::<*mut u8>().write({});
                         t
                     }}",
                     operand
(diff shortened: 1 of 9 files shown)


crates/rust/src/lib.rs
```diff
@@ -1248,6 +1248,15 @@ fn wasm_type(ty: WasmType) -> &'static str {
         WasmType::I64 => "i64",
         WasmType::F32 => "f32",
         WasmType::F64 => "f64",
+        WasmType::Pointer => "*mut ::core::ffi::c_void",
```
alexcrichton, line 1251: could this use `*mut u8` to be a bit easier to type and enable `.add` with a byte size directly?

crates/rust/src/lib.rs
```diff
@@ -1248,6 +1248,15 @@ fn wasm_type(ty: WasmType) -> &'static str {
...
+        WasmType::Length => "usize",
+
+        // `PointerOrI64` can hold either a `u64` or a pointer with provenance.
+        // Neither a `u64` nor a pointer type can portably do both, so we use
+        // `MaybeUninit<u64>`, since `MaybeUninit` is [documented] to preserve
+        // provenance.
+        // [documented]: https://github.com/rust-lang/rfcs/blob/master/text/3559-rust-has-provenance.md#reference-level-explanation
```
alexcrichton, line 1258: My read of the text in the RFC here is that `MaybeUninit<T>` has the same provenance of `T` where `u64` has no provenance here, but do you know if the RFC text meant differently?

Either way though I don't know of anything better so I think it's also ok to leave as-is.

crates/rust/src/lib.rs
```diff
@@ -1272,6 +1281,40 @@ fn bitcast(casts: &[Bitcast], operands: &[String], results: &mut Vec<String>) {
...
+            Bitcast::P64ToP => {
+                format!(
+                    "{}.as_mut_ptr().cast::<*mut core::ptr::c_void>().read()",
+                    operand
+                )
+            }
+            // Convert an `i32` into a pointer.
+            Bitcast::I32ToP => {
+                format!("{} as *mut ::core::ffi::c_void", operand)
+            }
```
alexcrichton, line 1313: Could this be removed entirely? It seems like we should avoid generating this ever because this implies that provenance can be manifest out of nowhere right?

alexcrichton: (and similarly below, could `PToI32` be removed entirely?

alexcrichton (approved):
Looks good to me! I'm mostly curious if you think we can remove `I32ToP` and `PToI32` entirely

crates/rust/src/lib.rs
sunfishcode: When we have a variant of an `i32` and a pointer, we unify the two types to a pointer type, so that when the variant does hold a pointer, it preserves provenance. When the variant holds an i32, there is no provenance anyway, so it's ok that we're doing these conversions without preserving provenance.

crates/rust/src/lib.rs
alexcrichton: Ah ok then in that case it sounds like exactly what we want here 👍

crates/rust/src/lib.rs
alexcrichton: Aha I see, makes sense!

Could you add some docs with respect to provenance on the `Bindgen` enum variants? If I understand right it's intended that `I32ToP` adds arbitrary provenance and `PToI32` removes that arbitrary provenance, right?

sunfishcode: I've now added some docs.

The way it's discussed in the Rust provenance docs, they say that what `I32ToP` does is creates a pointer with no provenance. Once the API is stablized in Rust, the functions that will replace these `as` casts have names like [`without_provenance`](https://github.com/rust-lang/rust/pull/117658).

PR #871, opened 2024-02-27
rust: Revert back to historical way of exporting functions

alexcrichton:
This commit is unfortunately a bit of whiplash for Rust users as it transitions the `wit_bindgen::generate!` macro to what it was a few months ago in terms of how exports work. Before describing that, though, let's first motivate this commit.

Today exports are not easy to work with in the Rust generator. To be clear this is no one's fault (either that or it's mostly mine), it's just the best we could think of at the time to get resources working. Whenever the `generate!` macro is invoked it's required to specify all exports to the macro via an `exports: { ... }` map. This is required for exported resources so generated bindings know what types to refer to. The problems with this approach are:

* You can't generate bindings ahead-of-time, bindings can only be generated when everything is "done". This means crates like `wasi` have to do weird things to bind exports and it wouldn't even work if there were exported resource types.

* There's a circular nature between user code and generated code. Generated code uses user types, but user types also use generated code. While this can work it has a very high risk of generating confusing error messages that are difficult to debug.

The upside of today's `exports` approach is that it works! These downsides were well-known before they were implemented, but at the time we couldn't think of anything better. Yesterday, however, I had an epiphany that I think we can do better.

This PR removes all the `exports: { ... }` bits entirely. Instead a completely different system is now in place for managing the exports of a crate. This system is basically the same as what everything was before `exports: { ... }` with a few cosmetic tweaks:

1. The `generate!` macro generates everything but does not export anything. This means that from a build process perspective `generate!` should be a noop. (e.g. it can be gc'd away)

2. The `generate!` macro does not refer to any user-defined types, so it should be a little silo which is much nicer from a composability point of view.

3. Resource exports still have concrete types generated. Generated bindings look very similar to before. The main difference is that instead of `OwnT` for `own<T>` exports and `&MyT` for `borrow<T>` exports they generate `T` and `TBorrow<'_>` where both of those types are bindings-generated.

4. The `generate!` macro itself generates a macro: `export_{name}!`. Here `{name}` is based on the world being bound. This macro is what actually exports an implementation and generates `#[no_mangle]` functions.

Overall this is basically what we had before (if I'm remembering correctly). The cosmetic tweaks come in the implementation. For example the `export_*!` macro has two forms instead of one:

    export_foo!(MyWorldImplementation);
    // .. is the same as ...
    export_foo!(MyWorldImplementation with_types_in self);

Here the second form solves the problem where `export_foo!` doesn't know the crate path back to itself, so it's optionally specified after `with_types_in`. By default this is `self` meaning that it's expected that `export_foo!` is located adjacent to the call to `generate!`. This can be configured though with a new Rust macro option `default_bindings_module: "..."`. Additionally the `export_foo!` macro is not exported from the crate by default, but that can also be configured with `pub_export_macros: true` now.

Included in this commit are other changes such as:

* Improved documentation for `generate!`, including documenting the state of the world after this commit.
* A new test for cross-crate behavior of the `generate!` macro.
* The Rust macro `export` option and `--export` CLI flag are both removed.
* Generated bindings for exported resource types are different and require manual method calls like `.get()` with an annotation of what the destination type is.
* Guests may need to implement more traits now as an `interface` with `resource`s but no functions must be implemented to configured resource types as associated types.
* The `WasmResource` trait, `Resource<T>` type, and `RustResource` trait are all now purely internal to the bindings and shouldn't show up in the public API.
* A new `default_bindings_module` option configures the default location to find where bindings are generated in the `export_*!` macro.
* A new `pub_export_macros` option configures whether macros are exported out of the current crate.

And finally, a good way to see the impact of this change is to browse the changes in the `tests/runtime/*` directory to all the `wasm.rs` files.


alexcrichton:
Also, I can write more words as to what my epiphany was about what empowered this change.

The reason we switch to `export: { ... }` was because of resources. Generated types need to refer to exported resources and they do so as `pub type GeneratedWitType = Resource<UserSpecifiedType>`. This meant that the knowledge of user types was required at bindgen time. My realization, though, was "what if we didn't do that?" Instead we now have:

```wit
package a:b;

interface i {
  resource my-resource {
    constructor();
    get: func() -> u32;
    set: func(x: u32);
  }
}

world foo {
  export i;
}
```

generates:

```rust
pub mod exports {
    pub mod a {
        pub mod b {
            pub mod i {
                #[derive(Debug)]
                #[repr(transparent)]
                pub struct MyResource {
                    handle: _rt::Resource<MyResource>,
                }

                type _MyResourceRep<T> = Option<T>;

                impl MyResource {
                    pub fn new<T: GuestMyResource>(val: T) -> Self {
                        // ...
                    }

                    pub fn get<T: GuestMyResource>(&self) -> &T {
                        // ..
                    }

                    pub fn get_mut<T: GuestMyResource>(&mut self) -> &mut T {
                        // ..
                    }

                    pub fn into_inner<T: GuestMyResource>(self) -> T {
                        // ..
                    }
                }

                #[derive(Debug)]
                #[repr(transparent)]
                pub struct MyResourceBorrow<'a> {
                    rep: *mut u8,
                    _marker: core::marker::PhantomData<&'a MyResource>,
                }

                impl<'a> MyResourceBorrow<'a> {
                    pub fn get<T: GuestMyResource>(&self) -> &T {
                        // ..
                    }
                }

                pub trait Guest {
                    type MyResource: GuestMyResource;
                }

                pub trait GuestMyResource: 'static {
                    fn new() -> Self;
                    fn get(&self) -> u32;
                    fn set(&self, x: u32);
                }
            }
        }
    }
}
```

(heavily edited)

Effectively what type is being accessed is deferred via a type parameter to when the user requests it. The `GuestMyResource` trait has defaulted methods that are `#[doc(hidden)]` which deal with resource intrinsics and the methods manipulate on.

So all-in-all, the "epiphany" was: what if things weren't quite so ergonomic as to implement `Deref`, and the answer was this PR.

peterhuene:
> For the last point I know that cargo component used to generate a crate but now it only generates src/bindings.rs, and I suspect that was probably influence by the wonkiness of how exports worked. I'm not sure if you would want to switch cargo-component back to the previous strategy or if the current strategy was chosen for other reasons.

Bindings in `cargo-component` has had a tortuous journey. 😭

The real reason `cargo-component` doesn't operate like `wit-bindgen::generate!` is that it can fully resolve dependencies from a registry and component authors can depend their component on other components directly; this synthesizes a target world for bindings that isn't defined by any WIT package on disk. This synthesis needs to happen in an async context where network requests are performed and procedural macros don't lend themselves to that. Thus, `cargo-component` generates bindings (or, historically, at least an encoded representation of the target world) before `cargo` is invoked.

Originally `cargo-component` magically generated a bindings crate and moved to procedural macro at the time when resources were introduced; the magic bindings crate had a lot of downsides, chiefly `cargo-component` *had* to wrap the `cargo` crate so that it could manipulate the workspace in-memory before passing it to various cargo ops; `cargo` is a *ginormous* build-time dependency and it was a maintenance burden to define, in `clap`, all the supported commands and keep them in sync.

When it moved to the proc-macro, `cargo-component` generated a binary-encoded representation of the target world and the macro read it a known location to feed it to the `wit-bindgen` machinery. Ultimately, this too had downsides, chiefly `rust-analyzer` really doesn't expect a procedural macro to read random files to base the expanded code on and so has no mechanism to watch those files (`include_bytes!` does not introduce a watch); thus the first expansion of the macro was cached and often would not re-evaluate when expected to. But, on the plus side, we could now spawn `cargo` as a subprocess rather than wrap the `cargo` crate and `cargo-component` was much faster to build and substantially smaller.

To fix the `rust-analyzer` issue, I ended up eliminating the macro entirely and just have `cargo-component` write the bindings into `src` directly. It also had the benefit of further minimizing the dependencies a component author has for building their components (there's only two now: `wit-bindgen` with only the `rt` feature and `bitflags` as a transitive dependency of `wit-bindgen`).

> First off, do you agree we should consider a change to this? In that do you agree that the previous way of exports was better?

I think we should consider it and I would agree that this way does seem to lend itself towards a cleaner implementation than having the generated bindings refer directly to user code.

That said, I like to work backwards from what this would look like from `cargo-component`. I think we'd still be generating bindings ahead-of-time, so it would probably look like this for the source file created by `cargo component new --lib --target foo:bar/baz`:

```rust
#[allow(dead_code)]
mod bindings;

struct Component;

impl bindings::Guest for Component {
    // ...
}

// other exported interfaces...

bindings::export!(Component);
```

It would be nice to have an option to control the export macro name like we had this before; having it based on the target world name complicates things slightly for `cargo-component`.

Other than that, do you see problems with having the above based on the changes in this PR?

> If so, will this break cargo component today, and if so do you know how hard it would be to fix that?

In addition to the break changes anyone using `wit-bindgen` would encounter (adding an `export!` invocation, changes to resources, etc.), I think `cargo-component` would only be additionally impacted by the removal of  `package.metadata.component.bindings.implementor` and `package.metadata.component.bindings.resources` settings in `Cargo.toml` which correspond to the `exports` argument to `wit_bindgen::generate!`; currently `cargo-component` picks some default values for the implementing type names, so these settings are infrequently used.

So far, I've been publishing breaking changes (and their remediation when manual intervention is required) with the release notes; I haven't bothered attempting to deprecate or provide an automatic fixing mechanism (eventually we'll get there with more stability), so I think this would just be covered by release notes in the breaking release.

> Also if so, would this be easier or harder than today for cargo component to integrate with in an "ideal" fashion?

To be honest, even if this were difficult for `cargo-component` to integrate, I wouldn't let that affect doing whatever we all agree is the right thing to do here. My biggest concern is the DX of component authors: what does their code look like, how easy is it to understand, and how easy is it to maintain?

That said, I don't think this would be a particularly difficult change to ingest from `cargo-component`.

alexcrichton:
Thanks for weighing in! I'm hopeful that this is the last time we change how the generated bindings work in a major way, but I tend to say that a lot...

> That said, I like to work backwards from what this would look like from cargo-component

I like this way of thinking!

> Other than that, do you see problems with having the above based on the changes in this PR?

Sounds perfect to me, I'll actually go ahead and just always name the macro `export!` since I don't think there should be an issue dropping the world name there.

---

Orthogonally I've also pushed up a documentation scheme which should render example generated documentation on docs.rs so that's hopefully a bit more useful than just a single crate with a single item.

alexcrichton:
> I'll actually go ahead and just always name the macro export!

Ok scratch that, I'll change the default to `export!` but the `wasi` crate will want to export two macros so I'll add an option for the macro name.

tests/runtime/resource_aggregates/wasm.rs
```diff
@@ -1,22 +1,42 @@
...
 });
 
+use core::ops::{Deref, DerefMut};
 use test::resource_aggregates::test::{foo as import_foo, Thing};
 
 pub struct Test {}
 
+export!(Test);
+
+impl Deref for exports::test::resource_aggregates::test::Thing {
```
dicej, line 12: Would it make sense to generate `Deref` and `DerefMut` automatically as part of bindings generation?

tests/runtime/rust_xcrate/world.wit
```diff
@@ -0,0 +1,26 @@
+package test:xcrate;
```
dicej, line 1: Would it make sense to throw a resource type or two in here, just to test that cross-crate resource exports work as expected?

dicej (approved):
This is a big improvement.  Thanks for doing it!

tests/runtime/resource_aggregates/wasm.rs
alexcrichton: Ideally! Unfortunately though that can't be done because the `Target` of the deref, `MyThing`, isn't known at bindings generation time. 

It's possible that during the `export!` macro we could add these impls, but I was running into some issues doing that with weird "this type isn't public enough" when I thought it was, so I didn't bottom that out much further.

alexcrichton: Oh I also realized that for `export!` implementing `Deref` doesn't work cross-crate which is also why I didn't include it (as that would define a trait for a foreign type)

PR #875, opened 2024-02-29
Support multiple `--with` arguments in the Rust generator

alexcrichton:
Don't require one large `--with`, support having multiple arguments.


crates/rust/src/lib.rs
```diff
@@ -889,6 +877,11 @@ impl WorldGenerator for RustWasm {
         }
         self.types.analyze(resolve);
         self.world = Some(world);
+
+        dbg!(&self.opts.with);
```
sunfishcode, line 881: Did you mean to leave this `dbg` in?

alexcrichton: Oops, no!

PR #877, opened 2024-03-03
Add `no_std` to `wit-bindgen-rt` crate.

peterhuene:
We forgot to add this when we refactored `wit-bindgen-rt` into its own crate.

### DIFF
--- a/crates/guest-rust/rt/src/lib.rs
+++ b/crates/guest-rust/rt/src/lib.rs
@@ -1,3 +1,5 @@
+#![no_std]
+
 extern crate alloc;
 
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks!

I'm not sure how best to release this since I'm wary to make two special cases in a row for this crate, I'm more tempted to do an 0.21.0 release to prevent the released versions from all getting out of sync, would that be ok for cargo-component?

peterhuene:
@alexcrichton sure that'd be fine.

alexcrichton:
Ok I've posted a bump to https://github.com/bytecodealliance/wit-bindgen/pull/879

PR #878, opened 2024-03-04
Rust: avoid emitting trailing whitespace

sunfishcode:
This works around what appears to be a bug in rustfmt when generating bindings for wasi-http. If the code is too complex, rustfmt appears to stops formatting, and then abort when it leaves behind trailing whitespace.


sunfishcode:
This is hitting some errors that I don't yet understand.

alexcrichton (approved):

sunfishcode:
Thanks, that was indeed the spot!

PR #881, opened 2024-03-05
doc: remove `exports` section from `generate!` options

rkrishn7:
This PR removes the documentation for the `exports` map in the `generate!` macro as the export behavior has changed in #871


alexcrichton (approved):
Oops I kept telling myself to not forget to remove this and then I forgot...

PR #885, opened 2024-03-07
Several tidyings in the C bindings

sunfishcode:
- Add several comments to the C bindings output

- Move the `#include <stdlib.h>` out of the `.h` file and into the `.c` file, since it's for `realloc`/`free` etc. and those are only used in the `.c` file.

- Use named structs inside of typedefs, so that the structs have names in the generated debug info.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -1691,7 +1691,7 @@ impl InterfaceGenerator<'_> {
         // canonical ABI.
         uwriteln!(
             self.src.c_adapters,
-            "\n__attribute__((__used__, __export_name__(\"{export_name}\")))"
+            "\n__attribute__((__export_name__(\"{export_name}\")))"
         );
         let name = self.c_func_name(interface_name, func);
         let import_name = self.gen.names.tmp(&format!("__wasm_export_{name}"));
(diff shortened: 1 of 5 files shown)


Mossaka (approved):

crates/c/src/lib.rs
```diff
@@ -1725,7 +1733,7 @@ impl InterfaceGenerator<'_> {
         if abi::guest_export_needs_post_return(self.resolve, func) {
             uwriteln!(
                 self.src.c_fns,
-                "__attribute__((__weak__, __export_name__(\"cabi_post_{export_name}\")))"
+                "__attribute__((__used__, __export_name__(\"cabi_post_{export_name}\")))"
```
alexcrichton, line 1736: Can this be both "used" and "weak"? The intention for weak IIRC is to enable overriding this symbol if the default behavior isn't desired

alexcrichton (approved):
looks good to me as well, just one minor question

crates/c/src/lib.rs
sunfishcode: Ah, those were left over from an experiment I was doing but ended up reverting. I've now removed the `__used__`s too.

crates/c/src/lib.rs
```diff
@@ -1725,7 +1733,7 @@ impl InterfaceGenerator<'_> {
         if abi::guest_export_needs_post_return(self.resolve, func) {
             uwriteln!(
                 self.src.c_fns,
-                "__attribute__((__weak__, __export_name__(\"cabi_post_{export_name}\")))"
+                "__attribute__((__export_name__(\"cabi_post_{export_name}\")))"
```
alexcrichton, line 1736: Oh could this keep `__weak__` though? (to ensure that if needed it can be overridden without extra bindings options)

sunfishcode: Oops, yeah. Fixed.

PR #897, opened 2024-03-18
Add `#[dead_code]` to the Rust generated bindings.

### LINKED ISSUE #896: Add `#![allow(dead_code)]` to generated Rust bindings
I cloned Dan's ["hello-wasi-http" project](https://github.com/sunfishcode/hello-wasi-http) to use as the base [for a demo](https://github.com/yoshuawuyts/wasi-clippy), and I noticed that the generated Rust bindings generate a *lot* of warnings:

Screenshot 2024-03-18 at 17 27 39

If this was a separate crate with `pub` items these warnings would not be emitted. But because they're just a regular Rust module it triggers these lints. I think we would be well-served by including a `#![allow(dead_code)]` pragma at the top of the generated bindings. Thanks!

sunfishcode:
This uses attributes on each `pub mod` instead of one at the top of the file, because wit-bindgen's output can sometimes be nested inside of other constructs.

Fixes #896.

### DIFF
--- a/crates/rust/src/interface.rs
+++ b/crates/rust/src/interface.rs
@@ -419,7 +419,7 @@ macro_rules! {macro_name} {{
         let path_to_root = self.path_to_root();
         let module = format!(
             "\
-                #[allow(clippy::all)]
+                #[allow(dead_code, clippy::all)]
                 pub mod {snake} {{
                     #[used]
                     #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


alexcrichton:
This looks reasonable to me, although unfortunately there's not a great way to test this. I believe macro-generated code is generally automatically exempt from this lint which is why this hasn't cropped up before. Could you try running the test suite with `WIT_BINDGEN_DEBUG=1` set? That should simulate rustc not seeing macro-generated code and it may serve for at least locally verifying that no extra warnings are generated.

I think, for example, that this won't handle dead code related to top-level world types/imports, only those in packages/interfaces.

sunfishcode:
I've tested that this PR fixes the warnings eported in #896.

I've now also run the tests with `WIT_BINDGEN_DEBUG` however that hits several pre-existing errors. I filed #900 to fix some warnings, at least.

PR #901, opened 2024-03-20
Update to Wasmtime 19

alexcrichton:


tests/runtime/resources.rs
```diff
@@ -94,7 +94,6 @@ fn run_test(exports: Guest, store: &mut Store<crate::Wasi<MyImports>>) -> Result
 
     let dropped_zs_start = z.call_num_dropped(&mut *store)?;
 
-    ResourceAny::resource_drop(x_instance, &mut *store)?;
```
sunfishcode, line 97: Do I understand correctly that without this line, `x_instance` is leaked? And that's ok because this code is just testing for the number of z's dropped?

alexcrichton: This was actually mistakenly allowed by Wasmtime prior and is now caught by stricter checks. The previous use of this `x_instance` variable actually took ownership of it, so this deletion is fixing the mistake.

sunfishcode: Ah, the `add` for the z's takes `borrow`s, but the `add` for the `x` takes an owned `x`.

sunfishcode (approved):

PR #902, opened 2024-03-20
Fix the C lowering for `RecordLift` to be compatible with C++.

sunfishcode:
C++ prohibits non-constant struct literal initializers from being implicitly converted to narrower types, so insert explicit casts. This allows the generated code to be compiled as both C and C++.

Specifically, this fixes the following error in C++:
```
test.c:40:5: error: non-constant-expression cannot be narrowed from type 'int32_t' (aka 'int') to 'enum_t' (aka 'unsigned char') in initializer list [-Wc++11-narrowing]
```

While debugging this, I also tidied up some redundant parens in `load_ext`.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -2068,7 +2068,7 @@ impl<'a, 'b> FunctionBindgen<'a, 'b> {
     fn load_ext(&mut self, ty: &str, offset: i32, operands: &[String], results: &mut Vec<String>) {
         self.load(ty, offset, operands, results);
         let result = results.pop().unwrap();
-        results.push(format!("(int32_t) ({})", result));
+        results.push(format!("(int32_t) {}", result));
     }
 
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):

alexcrichton:
Do you know if it would be possible to add a test for this by compiling with `-Werror` or similar?

sunfishcode:
I tried testing the old generated code with `-Weverything` in C mode and clang has no warning for this.

alexcrichton:
I think right now the test framework supports `wasm.c` as C source files, but could `wasm.cpp` also be supported and an existing test translated to C++? (and/or just copied over)

sunfishcode:
I've now modified the test runner to also compile all the tests as C++. That uncovered several more compilation errors, which I've now also fixed.

alexcrichton (approved):

PR #912, opened 2024-03-28
fix(tinygo): syntax error when field name is 'type' in record and a function exists that uses that record

rajatjindal:
This PR fixes generator for go when:

- world has a keyword as name
- record has a field with keyword as name AND that record is used in a function

in addition, it added following two testcases which worked without any explicit changes:

- interface has keyword
- enum has keyword

an example of failure can be seen below. Notice `lift_ret_ptr.type` in the below generated code, here `type` is a golang keyword. and thus fails.

error:
```
<standard input>:25:11: expected selector or type assertion, found 'type'
test types ... FAILED
```

generated code:

```go
    if lift_ret_ptr.type == 0 {
      lift_ret_val_Type = WasiFilesystem0_2_0_TypesDescriptorTypeUnknown()
    }
```


alexcrichton (approved):
Nice, thanks!

crates/go/src/lib.rs
```diff
@@ -294,7 +294,7 @@ impl WorldGenerator for TinyGo {
         // prepend package and imports header
         let src = mem::take(&mut self.src);
         wit_bindgen_core::generated_preamble(&mut self.src, env!("CARGO_PKG_VERSION"));
-        let snake = self.world.to_snake_case();
+        let snake = avoid_keyword(self.world.to_snake_case().as_str()).to_owned();
```
Mossaka, line 297: Good find, thanks!

Mossaka (approved):

crates/c/src/lib.rs
```diff
@@ -3246,5 +3246,41 @@ pub fn to_c_ident(name: &str) -> String {
         "ret" => "ret_".into(),
         "err" => "err_".into(),
         s => s.to_snake_case(),
+    };
+
+    // escape golang keywords because they get used in CGO
```
Mossaka, line 3251: It seems strange to me that the C gen needs to understand what Go keywords are. Could we move this logic to the go gen? Probably what we can do is to sanitize the result of calling `to_c_ident(*)` in the Go gen.

Mossaka: I just pushed a commit that swapped the underscore from postfix to prefix to Go keywords. The reason behind this change is that if C struct's field name is a Go keyword, then CGO can access that by prefixing them with the underscore.

Mossaka (approved):

PR #913, opened 2024-03-29
feat(rust): allow generate unused records and variants

### LINKED ISSUE #910: rust: How to export all structs
How to make `wit_bindgen::generate!`  generate all types in `world export <<interface>>`?

Dead code elimination is performed by default. 

If a record or variant is not an input or return value of any function, such a type will not be generated as rust struct and enum.

oovm:
Close #910


alexcrichton:
Thanks!

I think that the CI failure is coming from the C# bindings so you may need to ignore the new test case. That being said I also think the new test case isn't testing anything since the allow-unused option isn't specified by default. Can you add a specific test for this to `crates/rust/tests/codegen.rs` as well?

crates/rust/src/lib.rs
```diff
@@ -182,6 +182,10 @@ pub struct Opts {
     /// candidate for being exported outside of the crate.
     #[cfg_attr(feature = "clap", arg(long))]
     pub pub_export_macro: bool,
+
+    /// Whether to generate unused structures, not generated by default (false)
+    #[cfg_attr(feature = "clap", arg(long))]
+    pub allow_unused: bool,
```
alexcrichton, line 188: To bikeshed a bit, how about "generate_unused_types"?

alexcrichton:
Er, sorry for the lots of emails, but can you also add [documentation here](https://github.com/bytecodealliance/wit-bindgen/blob/ed26dbf2b85fab3b6e533fa8169ac88cc98f4463/crates/guest-rust/src/lib.rs#L642)?

alexcrichton (approved):

PR #923, opened 2024-04-06
Testing tweaks to the csharp-mono project generation

lewing:
This is just an initial pass to become familiar with the tests and start reducing the differences between the csharp targets

### DIFF
--- a/crates/csharp/src/csproj.rs
+++ b/crates/csharp/src/csproj.rs
@@ -185,18 +185,6 @@ impl CSProjectMonoBuilder {
 
         let aot = self.aot;
 
-        fs::write(
-            self.dir.join("rd.xml"),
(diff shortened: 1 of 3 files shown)


jsturtevant (approved):
LGTM

jsturtevant:
@lewing anything else needed on this?

jsturtevant:
We are bumping into the known issue with the return pointer alignment.  I will disable that test until we can track down the issue.  fyi @yowl

yowl:
I have a fix incoming, which at least solves the variant problem

yowl:
> I have a fix incoming, which at least solves the variant problem

https://github.com/bytecodealliance/wit-bindgen/pull/945

Issue #922, opened 2024-04-06
Clippy warning inside macro, docs for unsafe trait missing Safety section


sporkmonger:
This is happening with the 0.24.0 release, not sure about other versions. Because it's inside the macro, I couldn't see an obvious way to ignore the warning w/ the usual method. This is a warning but I have warnings upgraded to errors in our CI.

```
error: docs for unsafe trait missing `# Safety` section
  --> crates/wasm-sdk/src/lib.rs:8:5
   |
8  | /     wit_bindgen::generate!({
9  | |         world: "bulwark:plugin/platform",
10 | |     });
   | |______^
```

For the time being I ended up just disabling it on the command line, but it's likely to also affect anyone using our SDK as it also triggers for our own macros which are wrapping the `generate!` macro.

sporkmonger:
Created a PR to fix.

sporkmonger:
Closed by #926.

PR #926, opened 2024-04-08
fix(rust): suppress safety warning on `generate` macro (#922)

sporkmonger:
I'm not sure if it'd be better to add the Safety section Clippy is looking for or if it's preferable to ignore the warning, but it seemed like latter might be better, so that's what I opted for.


alexcrichton (approved):
Thanks!

PR #928, opened 2024-04-16
Use a unique name for `libwit_bindgen_cabi_realloc.a` per-crate-version

alexcrichton:
This commit updates how the weak `cabi_realloc` symbol is linked. Previously rustc was told to use `-lwit_bindgen_cabi_realloc` with a `-L` pointing to the source directory of the crate which has a precompiled copy. This caused issues, however, when two versions were present in a crate graph but only one was used. With two versions present in a crate graph rustc would use two `-L` flags with directories that both contain the same-named library. Which one was picked depended on the order that rustc passed flags. If only one crate was used, however, then rustc would not pass the rlib for the other crate. This could end up in a situation where with two wit-bindgen versions A and B:

* During linking, A's `libwit_bindgen_cabi_realloc.a` file was used.
* Rustc only passed B's rlib since A wasn't actually used anywhere.
* The linker then tried to load A's version of the actual cabi_realloc symbol, but only B's was present.

The fix in this commit is to uniquely name the library name based on the crate version. This matches how the symbol name is different per crate version, for example.

### DIFF
--- a/crates/guest-rust/rt/build.rs
+++ b/crates/guest-rust/rt/build.rs
@@ -22,10 +22,6 @@ fn main() {
         "wit_bindgen_cabi_realloc{}",
         env!("CARGO_PKG_VERSION").replace(".", "_")
     );
-    let mut src = env::current_dir().unwrap();
-    src.push("src");
-    src.push("libwit_bindgen_cabi_realloc.a");
-
     let dst = out_dir.join(format!("lib{dst_name}.a"));
 
     std::fs::copy(&src, &dst).unwrap();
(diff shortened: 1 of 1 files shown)


crates/guest-rust/rt/build.rs
```diff
@@ -10,8 +14,22 @@ fn main() {
         panic!("only wasm32 supports cabi-realloc right now");
     }
 
-    println!("cargo:rustc-link-lib=wit_bindgen_cabi_realloc");
-    let cwd = std::env::current_dir().unwrap();
-    let cwd = cwd.display();
-    println!("cargo:rustc-link-search=native={cwd}/src");
+    let mut src = env::current_dir().unwrap();
+    src.push("src");
+    src.push("libwit_bindgen_cabi_realloc.a");
```
sunfishcode, line 19: These three lines look like an unintended copy of code below.

sunfishcode (approved):
Looks good with one fix:

PR #930, opened 2024-04-21
fix(go): fix the build error of generated code for empty resource type

### LINKED ISSUE #929: If an exported interface uses an exported resource type, the generated code won't compile
```wit
package foo:bar;

interface f {
    resource fd;
}

interface utils {
    use f.{fd};
    my-func: func() -> own<fd>;
}

world test {
    import f;
    export f;
    export utils;
}
```

The generated TinyGo code failed to compile for this WIT file. Notice that the `fd` from `utils` is the exported `f.{fd}` type.

Mossaka:
If a resource is exported and that resource doesn't have a constructor, the generated Go code would fail. This PR fixes this issue by allowing the generator to generate Go types and interfaces if there are exported resources. 

Closes #929


Mossaka:
Ahh it seems like the added codegen test cases also failed the rust generator, could you please take a look 👀? @alexcrichton

alexcrichton:
Want to exclude this test in Rust for now? I can come and fix it after this lands (looks benign and related to how things are tested)

Mossaka:
Skipped these tests for the rust bindgen

FrankReh:
May I ask the title of this PR be changed to reflect what the fix is? That would make it more consistent with other PRs in this repo. And adding Go or Golang to it would make it even more precise.

alexcrichton (approved):

PR #939, opened 2024-04-26
add resource support to C# generator

dicej:
This adds resource support to the C# generator and also fixes a few miscellaneous issues that were preventing tests from passing.

I believe the code generated for imported resources is reasonably ergonomic, although I'm pretty new to C#, so I'm open to suggestions to make it more idiomatic.

As with other languages, the code for exported resources is a bit less ergonomic, I'm using an `interface` to represent the API, but I also want implementations to inherit code which manages the resource handle, for which I'm currently using a `class`.  The upshot is that implementing an exported resource involves both extending a `class` and implementing an `interface`.  Again, I'm open to suggestions about improving this.  Note that exporting resources tends to be a lot less common in end-user code than importing them, so we probably don't need to obsess over ergonomics in that case.

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -2,8 +2,8 @@ mod component_type_object;
 
 use anyhow::Result;
 use heck::{ToLowerCamelCase, ToShoutySnakeCase, ToUpperCamelCase};
+use indexmap::IndexMap;
 use std::{
-    any::Any,
     collections::{HashMap, HashSet},
     fmt::Write,
     iter, mem,
@@ -17,8 +17,8 @@ use wit_bindgen_core::{
 use wit_bindgen_core::{
     uwrite, uwriteln,
     wit_parser::{
-        Docs, Enum, Flags, FlagsRepr, Function, FunctionKind, Int, InterfaceId, Record, Resolve,
-        Result_, SizeAlign, Tuple, Type, TypeDefKind, TypeId, TypeOwner, Variant, WorldId,
+        Docs, Enum, Flags, FlagsRepr, Function, FunctionKind, Handle, Int, InterfaceId, Record,
+        Resolve, Result_, SizeAlign, Tuple, Type, TypeDefKind, TypeId, TypeOwner, Variant, WorldId,
         WorldKey,
     },
     Files, InterfaceGenerator as _, Ns, WorldGenerator,
@@ -36,14 +36,16 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 ";
 
 #[derive(Default, Debug, Clone)]
 #[cfg_attr(feature = "clap", derive(clap::Args))]
 pub struct Opts {
-    /// Whether or not to generate a stub class for exported functions
     #[cfg_attr(feature = "clap", arg(long, default_value_t = StringEncoding::default()))]
     pub string_encoding: StringEncoding,
+
+    /// Whether or not to generate a stub class for exported functions
     #[cfg_attr(feature = "clap", arg(long))]
     pub generate_stub: bool,
 
@@ -58,6 +60,31 @@ impl Opts {
             opts: self.clone(),
             ..CSharp::default()
         })
+    }
+}
+
+#[derive(Clone)]
+struct ResourceInfo {
+    module: String,
+    name: String,
+    docs: Docs,
+    direction: Direction,
+}
+
+impl ResourceInfo {
+    /// Returns the name of the exported implementation of this resource.
+    ///
+    /// The result is only valid if the resource is actually being exported by the world.
+    fn export_impl_name(&self) -> String {
+        format!(
+            "{}Impl.{}",
+            CSharp::get_class_name_from_qualified_name(&self.module)
+                .1
+                .strip_prefix("I")
+                .unwrap()
+                .to_upper_camel_case(),
+            self.name.to_upper_camel_case()
+        )
     }
 }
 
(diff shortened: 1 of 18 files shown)


crates/csharp/src/lib.rs
```diff
@@ -1198,50 +1351,195 @@ impl InterfaceGenerator<'_> {
...
+                            internal static extern void wasmImportResourceDrop(int p0);
+
+                            [DllImport("{module_name}", EntryPoint = "[resource-new]{name}"), WasmImportLinkage]
+                            internal static extern int wasmImportResourceNew(int p0);
+                    
+                            [DllImport("{module_name}", EntryPoint = "[resource-rep]{name}"), WasmImportLinkage]
+                            internal static extern int wasmImportResourceRep(int p0);
+                        }}
+
+                        protected virtual void Dispose(bool disposing) {{
```
yowl, line 1439: Is there a value in `IDisposable` for the export.  There isn't anything unmanaged to worry about, I'd think we could just drop it from `RepTable`

crates/csharp/src/lib.rs
```diff
@@ -1198,50 +1351,195 @@ impl InterfaceGenerator<'_> {
...
+                    public class {upper_camel}: IDisposable {{
+                        internal int? Handle {{ get; set; }}
+
+                        internal readonly record struct THandle(int Handle);
+
+                        internal {upper_camel}(THandle handle) {{
+                            Handle = handle.Handle;
+                        }}
+
+                        public void Dispose() {{
```
yowl, line 1380: I believe you aslo want a destructor in this class that calls `Dispose(false)` .  We have no managed resources to clean up (so we do nothing with the bool), but it is the pattern.  If the user does not call `Dispose`, and the object is collected, we want to drop the imported resource.

crates/csharp/src/lib.rs
```diff
@@ -1198,50 +1351,195 @@ impl InterfaceGenerator<'_> {
...
+                            Dispose(true);
+                            GC.SuppressFinalize(this);
+                        }}
+
+                        [DllImport("{module_name}", EntryPoint = "[resource-drop]{name}"), WasmImportLinkage]
+                        private static extern void wasmImportResourceDrop(int p0);
+        
+                        protected virtual void Dispose(bool disposing) {{
+                            if (Handle.HasValue) {{
+                                wasmImportResourceDrop((int) Handle);
```
yowl, line 1390:
```suggestion
                                wasmImportResourceDrop(Handle.Value);
```

crates/csharp/src/lib.rs
```diff
@@ -1198,50 +1351,195 @@ impl InterfaceGenerator<'_> {
...
+                    r#"
+                    public abstract class {upper_camel}: IDisposable {{
+                        internal int? Handle {{ get; set; }}
+
+                        public void Dispose() {{
+                            Dispose(true);
+                            GC.SuppressFinalize(this);
+                        }}
+
+                        internal static class WasmInterop {{
```
yowl, line 1428: Do we need this class on the export side?

yowl (commented):
Thanks, this is great.  
Some initial comments and minor c# changes.

crates/csharp/src/lib.rs
dicej: I _think_ we could technically get rid of it, but there's an argument for keeping it:

When a component passes ownership of an exported resource instance to the host, it calls the `[resource-new]` function to allocate a handle for it.  Later, if the host passes ownership of that same instance back to the component, the component could technically dispose of the handle immediately using `[resource-drop]`, but if it wants to pass ownership back to the host again, it will have to call `[resource-new]` and allocate a new handle.  By keeping the original handle alive, we avoid the churn of calling `[resource-new]` and `[resource-drop]` repeatedly, which could noticeably reduce overhead in a tight loop.  The drawback, as you point out, is that we need to implement `IDisposable` to ensure we call `[resource-drop]` once the instance is no longer needed.

dicej: Well now that I look at the code I wrote, it doesn't seem to be tracking the handle the way I described above, so I'll dig into it some more.

dicej: Okay, _now_ the code does what I described above.

dicej: I've added a comment.

dicej:
I believe I've addressed all the feedback so far.  Please let me know if there's anything I've missed, or anything else that needs attention.

jsturtevant (approved):

crates/csharp/src/lib.rs
yowl: Thanks, I didn't appreciate that the ownership could be passed over.

crates/csharp/src/lib.rs
yowl: As it is fine with me, thanks.

crates/csharp/src/rep_table.cs
```diff
@@ -0,0 +1,46 @@
+/**
+ * This class is used to assign a unique integer identifier to instances of
+ * exported resources, which the host will use as its "core representation" per
+ * https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#definition-types.
+ * The identifier may be used to retrieve the corresponding instance e.g. when
+ * lifting a handle as part of the canonical ABI implementation.
+ */
+internal class RepTable {
```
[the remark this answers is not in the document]
silesmo: Sorry I had forgotten to press publish on the review @dicej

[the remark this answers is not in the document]
dicej: Sorry, I'm not quite understanding.  I don't think there are any classes which inherit from `RepTable`.  The classes we generate for exported resources each have a `static RepTable<T>` field, but there's no inheritance involved.  In any case, it looks like your comment is addressing an older version of the code.  Do you have a concern with the latest version, and if so, can you clarify what you'd like to see changed?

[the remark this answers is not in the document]
dicej: Ah, I think I see your point: you'd prefer to avoid having a base class for each exported resource type.  We can certainly avoid it per the https://github.com/bytecodealliance/wit-bindgen/pull/939#discussion_r1581628568 discussion.  I suspect we'll either need to remove that optimization (i.e. call `[resource-new]` and `[resource-drop]` each time as necessary rather than reuse handles) or leave it up to the implementer somehow.

PR #940, opened 2024-04-28
Fix C and C++ keywords with underscores collisions

primoly:
Convert to snake case *before* checking for keyword collisions, so C and C++ keywords with underscores will be “escaped” as well.


jsturtevant:
@primoly maybe it makes sense to add this commit and the one from #941 to https://github.com/bytecodealliance/wit-bindgen/pull/942 which now has the wit for the generator test to catch these scenarios?

PR #941, opened 2024-04-28
Escape C# keywords

primoly:


jsturtevant:
@primoly could you give a description as to what this is fixing?  was there a bug you ran into?

jsturtevant:
maybe #942?

PR #945, opened 2024-04-30
[c#] Fix GC hole in return areas

yowl:
This fixes a misconception (mine) with inlined arrays.  They are subject to GC moving them, so now we use `FixedAddressValueType` for exports and for imports a local array, `fixed` for the duration of its use.

Enabled the variant test.

Minor tidy.

cc @jsturtevant @silesmo @dicej


dicej (approved):

crates/csharp/src/lib.rs
```diff
@@ -2335,19 +2316,26 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
                     self.src,
                     "
-                    var {ret_area} = new {name}WasmInterop.ImportReturnArea();
-                    var {ptr} = {ret_area}.AddressOfReturnArea();
-                    "
+                    var returnArea = new {0}[{1}];
+                    fixed ({0}* retArea = &returnArea[0])
+                    {{
+                        var ptr = (nint)retArea;
+                    ",
```
jsturtevant, line 2334: are we missing `}}` here?

yowl: Well it should close here https://github.com/bytecodealliance/wit-bindgen/pull/945/files#diff-0426c4d75f64d7f2434ca6b6e2da3cfaefbd63389fec8eee07155942cd8f689dR2285.  I'm trying to repro the CI error locally.

yowl: I missed a few things, correct use of variable names, and close all fixed blocks.

jsturtevant (approved):

PR #946, opened 2024-05-06
fix(c): Prefix import module name for resource-drop with `[export]`

kateinoigakukun:
Import module name for exported resource's drop should be prefixed with [export]

### DIFF
--- a/tests/runtime/resources/wasm.cs
+++ b/tests/runtime/resources/wasm.cs
@@ -10,6 +10,11 @@ namespace ResourcesWorld.wit.exports
         {
             return new Z(((Z) a).val + ((Z) b).val);
         }
+
+        public static void Consume(IExports.X x)
+        {
+          x.Dispose();
+        }
         
         public static Result<None, string> TestImports()
         {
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Thanks! Mind updating the host-side of the test you're modifying here to call the new function as well?

kateinoigakukun:
Sure!

alexcrichton:
I think that this dotnet generator may need to exclude this test now, but otherwise looks good

kateinoigakukun:
I don't have enough hardware access to fix c-sharp generator, so let me keep it disabled for now :cry:

alexcrichton (approved):

tests/runtime/resources/wasm.cs
```diff
@@ -10,6 +10,12 @@ public static IExports.Z Add(IExports.Z a, IExports.Z b)
         {
             return new Z(((Z) a).val + ((Z) b).val);
         }
+
+        public static void Consume(IExports.X x)
+        {
+          // FIXME: c-sharp generator seems wrong here
+          // x.Dispose();
```
kateinoigakukun, line 17: CC: @dicej There may be still resource GC hole around here in C# generator

PR #947, opened 2024-05-07
fix C# resource dtor recursion issue

dicej:
Given than the destructor function we export for a given exported resource will take care of removing the rep from the `RepTable`, we should not try to do that in the `Dispose(bool)` method; just call `WasmInterop.wasmImportResourceDrop`, which will call the host, which will call back into the guest, which will remove the entry.


dicej:
Thanks to @kateinoigakukun for pointing this out!

alexcrichton (approved):

PR #949, opened 2024-05-09
Correct name in README

FrankReh:
Since #799, the comment in the README no longer matched the example WIT code.


FrankReh:
The old line in the example WIT file read
```
  import plugin: self.my-plugin-api
```
and this was changed to
```
import my-plugin-api;
```
If the old version were still valid (with perhaps a semicolon added) it is a more interesting example because it shows renaming an interface to something that presumably would be more useful in the guest's code that is implementing that world.

So if it is still valid, maybe the line should be reverted?

alexcrichton (approved):
Nah this looks good to me, thanks!

PR #958, opened 2024-05-29
Fix the mono runtime test config

lewing:
Update the build to get things working with the latest mono net9 targets

### DIFF
--- a/crates/csharp/src/csproj.rs
+++ b/crates/csharp/src/csproj.rs
@@ -194,7 +194,7 @@ impl CSProjectMonoBuilder {
         </PropertyGroup>
 
         <ItemGroup>
-          <NativeLibrary Include=\"{camel}_component_type.o\" />
+          <NativeLibrary Include=\"$(MSBuildProjectDirectory)/{camel}_cabi_realloc.c\" />
         </ItemGroup>
 
         <Target Name=\"_FixRootAssembly\" AfterTargets=\"PrepareForILLink\">
--- a/crates/csharp/src/csproj.rs
+++ b/crates/csharp/src/csproj.rs
@@ -71,6 +71,8 @@ impl CSProjectLLVMBuilder {
             <ImplicitUsings>enable</ImplicitUsings>
             <Nullable>enable</Nullable>
             <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+            <RuntimeIdentifier>wasi-wasm</RuntimeIdentifier>
+            <OutputType>Library</OutputType>
         </PropertyGroup>
         
         <PropertyGroup>
@@ -109,6 +111,7 @@ impl CSProjectLLVMBuilder {
                 <!--To inherit the global NuGet package sources remove the <clear/> line below -->
                 <clear />
                 <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+                <add key="dotnet9" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json" />
                 <add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />
                 <!--<add key="dotnet-experimental" value="C:\github\runtimelab\artifacts\packages\Debug\Shipping" />-->
               </packageSources>
@@ -168,7 +171,7 @@ impl CSProjectMonoBuilder {
         };
 
         let mut csproj = format!(
-            "<Project Sdk=\"Microsoft.NET.Sdk\">
+            "<Project Sdk=\"Microsoft.NET.Sdk.WebAssembly\">
     
         <PropertyGroup>
             <TargetFramework>net9.0</TargetFramework>
@@ -182,6 +185,7 @@ impl CSProjectMonoBuilder {
             <ImplicitUsings>enable</ImplicitUsings>
             <Nullable>enable</Nullable>
             <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+            <WasmGenerateAppBundle>true</WasmGenerateAppBundle>
         </PropertyGroup>
         
         <PropertyGroup>
@@ -190,9 +194,14 @@ impl CSProjectMonoBuilder {
         </PropertyGroup>
 
         <ItemGroup>
-          <NativeFileReference Include=\"{camel}_component_type.o\" Condition=\"Exists('{camel}_component_type.o')\"/>
+          <NativeLibrary Include=\"{camel}_component_type.o\" />
         </ItemGroup>
 
+        <Target Name=\"_FixRootAssembly\" AfterTargets=\"PrepareForILLink\">
+            <ItemGroup>
+                <TrimmerRootAssembly Update=\"@(TrimmerRootAssembly)\" Condition=\" '%(RootMode)' == 'EntryPoint' \" RootMode=\"Library\" />
+            </ItemGroup>
+        </Target>
         "
         );
 
--- a/crates/csharp/src/csproj.rs
+++ b/crates/csharp/src/csproj.rs
@@ -196,9 +196,6 @@ impl CSProjectMonoBuilder {
 
         <ItemGroup>
           <_WasiLinkStepArgs Include=\"-Wl,--component-type,{camel}_component_type.wit\" />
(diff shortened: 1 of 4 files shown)


mkhamoyan:
@lewing https://github.com/dotnet/runtime/pull/103881 fixed the conflicting naming error, but now when running wit-bindgen with this PR changes I see this failure 
csharp_mono_error
Because of this I couldn't validate if the fix is enough.

lewing:
`test result: ok. 27 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 965.25s`

lewing:
@dicej 
with this version I'm seeing
```
---- variants::run stdout ----
Error: failed to read component type file: D:\bytcodealliance\wit-bindgen\target\runtime-tests\variants\csharp-variants\VariantsWorld_component_type.o

Caused by:
    The system cannot find the file specified. (os error 2)
```
any thoughts?

crates/csharp/src/csproj.rs
```diff
@@ -190,9 +195,17 @@ impl CSProjectMonoBuilder {
         </PropertyGroup>
 
         <ItemGroup>
-          <NativeFileReference Include=\"{camel}_component_type.o\" Condition=\"Exists('{camel}_component_type.o')\"/>
+          <_WasiLinkStepArgs Include=\"-Wl,--component-type,{camel}_component_type.wit\" />
+          <!-- both versions of these seem to fail to find the .o
```
jsturtevant, line 199: Wit-bindgen doesn't produce the object file directly anymore,

I've used:
```
<CustomLinkerArg Include="@(WitComponentImports->Replace('\', '/')->'-Wl,--component-type,&quot;%(Identity)&quot;')" />
```

In https://github.com/bytecodealliance/componentize-dotnet/pull/35/files#diff-7f4dfa8be7454aa047e35760bfbb0f95c0e1402fd34943606ccd9975e45b8b71 

Not sure if `_WasiLinkStepArgs` is the same?

lewing:
> @dicej with this version I'm seeing errors like
> 
> ```
> ---- variants::run stdout ----
> Error: failed to read component type file: VariantsWorld_component_type.o
> 
> Caused by:
>     The system cannot find the file specified. (os error 2)
> ```
> 
> any thoughts?

haha, I see the problem now.  That is what I get for not looking again in the morning before.

dicej:
> @dicej with this version I'm seeing errors like
> 
> ```
> ---- variants::run stdout ----
> Error: failed to read component type file: VariantsWorld_component_type.o
> 
> Caused by:
>     The system cannot find the file specified. (os error 2)
> ```
> 
> any thoughts?

As of https://github.com/dotnet/runtime/pull/103752, the `*_component_type.o` files have been replaced with `*_component_type.wit` files.  Perhaps you're using an older runtime?

crates/csharp/src/csproj.rs
lewing: it's another case of bad coevolution https://github.com/dotnet/runtime/pull/107194 but they will be

lewing:
> > @dicej with this version I'm seeing errors like
> > ```
> > ---- variants::run stdout ----
> > Error: failed to read component type file: VariantsWorld_component_type.o
> > 
> > Caused by:
> >     The system cannot find the file specified. (os error 2)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > any thoughts?
> 
> As of [dotnet/runtime#103752](https://github.com/dotnet/runtime/pull/103752), the `*_component_type.o` files have been replaced with `*_component_type.wit` files. Perhaps you're using an older runtime?

Sadly I was rushing to get it working last night before heading out, I should have looked again this morning.  Sorry for the noise.

lewing:
Ok, so it was also very confusing that the features were chained so the native aot test build was writing over the mono tests once the compilation issue was fixed.

the `'NuGet-Migrations` failure is not related to these changes .

There do now appear to be a couple of failures on the mono side https://github.com/dotnet/runtime/issues/107212

crates/csharp/src/csproj.rs
```diff
@@ -161,23 +162,20 @@ impl CSProjectMonoBuilder {
         let camel = format!("{}World", world.to_upper_camel_case());
 
         let aot = self.aot;
-
-        let maybe_aot = match aot {
-            true => format!("<WasmBuildNative>{aot}</WasmBuildNative>"),
-            false => String::new(),
-        };
+        let tfm = "net9.0";
```
pavelsavara, line 165: Should we move to Net10 ? 
Because we are not going to backport fixes to Net9, right ?
One more below.

tests/runtime/main.rs
```diff
@@ -586,75 +587,16 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
                 panic!("failed to compile");
             }
 
-            let out_wasm = out_wasm.join("AppBundle").join(assembly_name);
+            let out_wasm = out_dir
+                .join("bin")
+                .join(configuration)
+                .join("net9.0")
```
pavelsavara, line 593: `tmf`

tests/runtime/main.rs
```diff
@@ -563,10 +564,10 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
...
             cmd.current_dir(&out_dir);
 
-            cmd.arg("build")
+            cmd.arg("publish")
                 .arg(out_dir.join(format!("{camel}.csproj")))
                 .arg("-c")
-                .arg("Debug")
+                .arg(configuration)
                 .arg("/p:PlatformTarget=AnyCPU")
                 .arg("--self-contained")
```
pavelsavara, line 572: for me `--self-contained` doesn't' work with `9.0.100-preview.6.24328.19` without workload. The publish doesn't fail but the file is located elsewhere and it's not single file.

After I install workload `Microsoft.NET.Runtime.WebAssembly.Wasi.Sdk version 9.0.0-preview.6.24327.7` I get
```
FooWorld failed with 4 error(s) (11.4s) → xxx\
    wasm-ld : error : unknown argument: --component-type
    wasm-ld : error : unknown file type: FooWorld_component_type.wit
```

Which makes sense for preview6.

With preview7 I get
```
    1: error while executing at wasm backtrace:
           0: 0x296a7b - .tmpJCkP17!fseek
           1: 0x8efe - .tmpJCkP17!load_icu_data
           2: 0x8ff4 - .tmpJCkP17!mono_wasm_load_runtime
           3: 0x9299 - .tmpJCkP17!initialize_runtime
           4: 0x92b8 - .tmpJCkP17!main
           5: 0x28bc5c - .tmpJCkP17!__main_void
           6: 0x7e78 - .tmpJCkP17!_start
           7: 0xb40aff - wit-component:adapter:wasi_snapshot_preview1!wasi:cli/run@0.2.0#run
```

tests/runtime/main.rs
```diff
@@ -563,10 +564,10 @@ fn tests(name: &str, dir_name: &str) -> Result<Vec<PathBuf>> {
 
             cmd.current_dir(&out_dir);
 
-            cmd.arg("build")
+            cmd.arg("publish")
                 .arg(out_dir.join(format!("{camel}.csproj")))
                 .arg("-c")
-                .arg("Debug")
+                .arg(configuration)
```
pavelsavara, line 570:
```suggestion
                .arg(configuration)
                .arg("-bl:publish-mono.binlog")
```

tests/runtime/main.rs
lewing: nothihng about this will work without the workload

crates/csharp/src/csproj.rs
lewing: there won't be a working net10 sdk for a while unfortunately

tests/runtime/main.rs
lewing: that is only one of the issues here, build doesn't respect the project name, and the publish doesn't respect the output directory.  @maraf we should discuss what options we have here

tests/runtime/main.rs
jsturtevant: I had to add `<WasmBuildNative>true</WasmBuildNative>`  to get it to produce a single file

lewing:
Several of the issues I hit should be fixed in dotnet/runtime main, but it testing that will be challenging until a proper net10 sdk is produced.

PR #960, opened 2024-05-31
full qualify Address to avoid collision with wit record of same name

nick1udwig:
In trying to compile Java to a WASM component, I ran into trouble because [my wit file](https://github.com/kinode-dao/kinode-wit/blob/master/kinode.wit#L25-L28) defines a `record address`. This collides with the codegen which frequently uses [teavm Address](https://github.com/konsoletyper/teavm/blob/master/interop/core/src/main/java/org/teavm/interop/Address.java)es.

This PR changes the codegen to fully qualify all Addresses, avoiding collision with user-defined `Address` classes.


alexcrichton:
Thanks! Mind adding a test in `tests/codegen/*.wit` for this too?

alexcrichton (approved):

nick1udwig:
Hmm, looks like the windows tests failed. Is this a "re-run and it'll work" scenario or should I take a second look at my test wit file?

alexcrichton:
I think this is a retry situation, although I already retried once and it failed that time too... In any case I don't believe this is related to this PR so I'll try to take it from here

alexcrichton:
It appears this is less flaky than expected, I've pinged @dicej who might be able to help here a bit more on Monday with this.

dicej:
The "return pointer not aligned" issue was intended to be addressed in https://github.com/bytecodealliance/wit-bindgen/commit/bcd136e583c51cb6819876478bdfd6bf96e1ea2d, but it looks like there's still a problem.  I've pinged Scott to see what he thinks, and I'll try to debug it myself as well.

dicej:
I'm able to reproduce it locally; will try to debug.  If I don't make progress, we'll just need to disable the C# `variants` runtime test for the time being.

dicej:
The generated code appears to be correct AFAICT (e.g. using `ulong` for 64-bit alignment), so maybe something's off at a lower level. For the time being, I'm disabling that test: https://github.com/bytecodealliance/wit-bindgen/pull/963

PR #963, opened 2024-06-03
temporarily remove C# `variants` runtime test

dicej:
Despite earlier efforts to tame the dreaded "return pointer not aligned" issue, it's still happening, even though generated code appears to be correct (e.g. using `ulong` for 64-bit alignment).  I'm removing that test until we get to the bottom of it.


jsturtevant (approved):
/lgtm

alexcrichton (approved):
Thanks!

Issue #964, opened 2024-06-04
Make C# bindings more ergonomic


dicej:
Currently, the C# generator generates and uses `Result` and `Option` types in a pretty literal translation of the WIT input, forcing the application programmer to use e.g. `IsOk`/`AsOk` to extract the value, which is awkward.  For functions which return `result<T, E>`, a more ergonomic option would be to generate a method which returns `T` and throws `E` (or some exception type with a payload of type `E`).  For `option<T>` a more idiomatic C# representation would be `T?`.

We still have to be careful about nested types (e.g. `result<result<T, E>, E>` and `option<option<T>>`) to avoid ambiguity, and sometimes `result` is used as a field or parameter, so we may still need to generate `Result` and `Option` types for use in those cases, but in most cases we can do the idiomatic thing.  Also, a function returning `result<result<T, E>, E>` could be translated into a method that returns `T` and throws `WitException<E>` with a `NestingLevel` read-only property to disambiguate which level of `err` occurred -- all without resorting to use of the `Result` type.

If there are no concerns, I'm planning to start working on a PR that implements the above.

cc @SteveSandersonMS

yowl:
Makes sense, thanks

jsturtevant:
sounds good to me

PR #968, opened 2024-06-06
C#: use exceptions for functions returning `result`

dicej:
This addresses part of #964.

For WIT functions returning `result<T, E>`, we now generate methods which return `T` and throw `WitException` such that the `E` value may be retrieved using `WitException.Value`.

I considered making `WitException` generic, but I couldn't find a precedent for generic exception classes when searching online, so I'm not sure how idiomatic that would be.  So, for now at least, you need to cast the value obtained using `WitException.Value` to the expected type.  Happy to revisit this if desired.

For WIT functions returning nested results, e.g. `result<result<T, E1>, E2>`, we generate methods which return `T` and throw `WitException`, using `WitException.NestingLevel` to disambiguate when `E1` and `E2` are the same.

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -135,6 +135,7 @@ pub struct CSharp {
     needs_interop_string: bool,
     needs_export_return_area: bool,
     needs_rep_table: bool,
+    needs_wit_exception: bool,
     interface_fragments: HashMap<String, InterfaceTypeAndFragments>,
     world_fragments: Vec<InterfaceFragment>,
     sizes: SizeAlign,
@@ -509,6 +510,23 @@ impl WorldGenerator for CSharp {
                         length = utf8Bytes.Length;
                         var gcHandle = GCHandle.Alloc(utf8Bytes, GCHandleType.Pinned);
                         return gcHandle.AddrOfPinnedObject();
+                    }
+                }
+                "#,
+            )
+        }
+
+        if self.needs_wit_exception {
+            src.push_str(
+                r#"
+                public class WitException: Exception {
+                    public object Value { get; }
+                    public uint NestingLevel { get; }
+
+                    public WitException(object v, uint level)
+                    {
+                        Value = v;
+                        NestingLevel = level;
                     }
                 }
                 "#,
@@ -929,14 +947,25 @@ impl InterfaceGenerator<'_> {
             _ => unreachable!(),
         };
 
-        let result_type = if let FunctionKind::Constructor(_) = &func.kind {
-            String::new()
+        let (result_type, results) = if let FunctionKind::Constructor(_) = &func.kind {
+            (String::new(), Vec::new())
         } else {
             match func.results.len() {
-                0 => "void".to_string(),
+                0 => ("void".to_string(), Vec::new()),
                 1 => {
-                    let ty = func.results.iter_types().next().unwrap();
-                    self.type_name_with_qualifier(ty, true)
+                    let (payload, results) = payload_and_results(
+                        self.resolve,
+                        *func.results.iter_types().next().unwrap(),
+                    );
+                    (
+                        if let Some(ty) = payload {
+                            self.gen.needs_result = true;
+                            self.type_name_with_qualifier(&ty, true)
+                        } else {
+                            "void".to_string()
+                        },
+                        results,
+                    )
                 }
                 _ => {
                     let types = func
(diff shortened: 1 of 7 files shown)


dicej:
Ugh, the dreaded "return pointer not aligned" issue.  Curiously, I'm not able to repro it locally.  I'm on Linux/ARM64 rather than Windows/x86_64, but I wouldn't expect that to make a difference.

jsturtevant:
> Ugh, the dreaded "return pointer not aligned" issue. Curiously, I'm not able to repro it locally. I'm on Linux/ARM64 rather than Windows/x86_64, but I wouldn't expect that to make a difference.

I've also had hard time repo'ing.  I believe we tracked it down to the garbage collector moving memory which explains why it's hard to repo.

dicej:
> I've also had hard time repo'ing. I believe we tracked it down to the garbage collector moving memory which explains why it's hard to repo.

Even so, there are very few sources of non-determinism in single-threaded Wasm, particularly in the simple tests cases we're running here, so I would have expected even GC scheduling to be fully deterministic.  And we appear to be using [`fixed` statements](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/fixed) to pin the return areas, so the GC shouldn't be moving them in any case.

dicej:
I've updated the `results` test so it uses only 32-bit values and thus avoids the `return pointer not aligned` issue.

alexcrichton (approved):
Unsure if you're waiting on me for this, but looks reasonable to me! I figure someone more familiar with C# ought to take a look as well too.

ThorstenHans:
I haven't seen a generic exception in C# so far. 

What you usually do is inheriting from the `Exception` class and making essential properties available on your custom class. 

It may make sense to inherit from a more specific exception class such as `IndexOutOfRangeException` or `ArgumentException` (which is use-case specific). 

Although it's a best practice to provide 3 constructors:

- `WitException()`
- `WitException(string message)`
- `WitException(string message, Exception innerException)`

It may make more sense - at least in this scenario (where you control creation of new `WitException` instances) - to go with something like this (assuming you can construct a meaningful exception message from `value`:

```csharp
public class WitException : Exception
{
    public WitException(object value) 
       : base($"Error due to {value}")
    {
        Value = value;
    }

    public WitException(object value, Exception innerException)
        : base($"Error due to {value}", innerException)
    {
        Value = value;
    }

    public object Value { get; }
}
```

This would allow you to create new `WitExceptions` using:

```csharp
throw new WitException(1234);
throw new WitException(1234, new ArgumentException("name has an invalid value"));
throw new WitException(1234, existingException);
throw new WitException(1234, new WitException(4321));
```

Users can examine the exception using something like this:

```csharp
try 
{ 
    // Do something
} 
catch (WitException ex) 
{ 
    if (ex.InnerException is ArgumentException) 
    { 
        // Handle ArgumentException
    }
    else if (ex.InnerException is WitException)
    {
        // Handle WitException
    }
    else 
    { 
        throw;    // Always rethrow exceptions you don't know how to handle.
    } 
}
```

dicej:
Thanks for the feedback, @ThorstenHans.

I definitely agree that the `WitException` constructor(s) should call `base(string, ...)` with a meaningful error message whenever possible.  The C# binding generator doesn't currently override `object.ToString` for any of the types it generates, so I don't expect that `$"Error due to {value}"` would communicate anything useful currently, but we could definitely start adding `ToString` overrides at least for types that are used as errors.

For a concrete example, here's what `wasi:http` uses as an error type: https://github.com/WebAssembly/wasi-http/blob/a81c61fc584278f801393b22289840a439e51f50/wit/types.wit#L31-L78.  I imagine we could generate a `ToString` method for that type which at least prints the variant case name.

I'm not sure how useful it would be to provide a `WitException` constructor that takes another exception, though.  There aren't any places in the generated code where we'd use that, and there's no way to pass an inner exception back to the host if a guest-exported function throws a `WitException` that contains an inner exception (i.e. the generated bindings for the export would just ignore it since it has no meaning at the component level).  In other words, the purpose of `WitException` is to translate to and from WIT-level errors, which have no notion of inner exceptions, so any properties other than `Value` would be unused in the case of imports and ignored in the case of exports.

dicej:
I don't want to make this PR more complex, and we can add `ToString` methods to types and use them to construct error messages later in a purely additive manner, so I've opened #976 to track that work.

PR #971, opened 2024-06-08
replace `rustfmt` with `prettyplease` for codegen

### LINKED ISSUE #880: replace `cargo fmt` with the `prettyplease` crate
In order to more structurally work around code formatting issues like those described in https://github.com/bytecodealliance/wit-bindgen/pull/878, we should consider switching from `cargo fmt` to Dtolnay's [`prettyplease`](https://docs.rs/prettyplease/latest/prettyplease/) crate. This crate was designed specifically to format codegen; meaning it will format faster, fail less frequently, but may occasionally result in subtly different formatting than what `cargo fmt` would have come up with.

I've used this [succesfully](https://github.com/yoshuawuyts/html/blob/e0fec5810c11e312112195ee37f95695c3c9bc49/crates/html-bindgen/src/utils.rs#L6) in my `http-bindgen` crate, and I believe this might work better for `wit-bindgen` too. In practice I've yet to experience any divergent code formatting output. If we're worried about this becoming an issue however, we could always add a `#![rustfmt::skip]` attribute to the top of our generated files. Thanks!

yoshuawuyts:
Closes #880. ~~Filing as a draft PR because I haven't yet run the tests locally.~~ Yay, tests pass!

As a neat side-effect: this should enable people to use the [rust docker images](https://hub.docker.com/_/rust) to build projects using `cargo-component`. Those images don't come with either `rustfmt` or `rustup` pre-installed, so `wit-bindgen` depending on `rustfmt` can actually be inconvenient in surprising ways. This PR should (hopefully) fix that!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1087,10 +1087,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
 
 [[package]]
-name = "proc-macro2"
-version = "1.0.79"
+name = "prettyplease"
+version = "0.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e835ff2298f5721608eb1a980ecaee1aef2c132bf95ecc026a11b7bf3c01c02e"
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Thanks!

PR #972, opened 2024-06-10
[Rust] flip the generation default for non-local interfaces

### LINKED ISSUE #904: Default to only generating bindings for the local wit package
*This issue was originally encountered using `wasmtime::bindgen`, but I'm posting this here since I think it likely impacts all bindings generators.*

Currently, bindings are generated for the package that contains the target world as well as all its dependencies. This can *sometimes* be the right thing, especially if the generated code does not need to interact with pre-existing code where bindings have already been generated. However, many users will want to interact with pre-existing code (mainly `wasi` projects) that already have their own bindings. 

`wit-bindgen` and `wasmtime::bindgen` support pointing the generator at existing types in the form of the `with` directive. However, if the user forgets to do this, and generates code that is redundant with existing bindings that live somewhere else, the user can run into some very cryptic and hard to understand errors (e.g., resource types not type-checking). 

In order to fix this, I propose we flip the default around to requiring the user to specify how non-local types should be treated. This can happen in two ways. First, the user can use the existing `with` directive. If the user forgets to add a `with` directive , they will get a constructive error message saying that they need to point to where that type lives. 

Second, is the case of users who don't want to use existing bindings (i.e., because there might not even be an authoritative set of bindings, and this is totally fine as long as they don't interact with code that expects authoritative bindings). In that case, we would add a second option which is essentially "please, generate the bindings for me". This would mirror the current behavior with the key difference that it is opt-in. The exact syntax for this is left as an exercise for the reader.

rylev:
Fixes #904 

Instead of non-local interfaces being generated automatically (and only not being generated when a `with` option is specified) - this PR changes the default to require a `with` option for all non-local interfaces.

If a foreign type is not present in the with `with` mapping an error is produced saying that the user must specify an entry in with map for all foreign types. The user can add an entry to with like they do today, or they can specify `generate` to have the type generated for them.

The error message when the `with` option is not specified is:

```
error: no remapping found for "my:package/my-interface" - use the `with` option to `generate!` to force the interface to be generated or specify where it is already defined:
       ```
       with: {
           "my:package/my-interface": generate
       } 
       ```
```

### DIFF
--- a/crates/guest-rust/src/examples/_1_interface_imports.rs
+++ b/crates/guest-rust/src/examples/_1_interface_imports.rs
@@ -24,4 +24,9 @@ crate::generate!({
     "#,
 
     path: "wasi-cli@0.2.0.wasm",
+
+    // specify that this interface dependency should be generated as well.
+    with: {
+        "wasi:cli/environment@0.2.0": generate,
+    }
 });
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -39,9 +39,9 @@ struct RustWasm {
     interface_last_seen_as_import: HashMap<InterfaceId, bool>,
     import_funcs_called: bool,
     with_name_counter: usize,
-    // Track the with options that were used. Remapped interfaces provided via `with`
+    // Track which interfaces were generated. Remapped interfaces provided via `with`
     // are required to be used.
-    used_with_opts: HashSet<String>,
+    generated_interfaces: HashSet<String>,
     world: Option<WorldId>,
 
     rt_module: IndexSet<RuntimeItem>,
@@ -75,9 +75,9 @@ impl GenerationConfiguration {
 
 /// How an interface should be generated.
 enum InterfaceGeneration {
-    // Remapped to some other type
+    /// Remapped to some other type
     Remap(String),
-    // Generate as normal
+    /// Generate the interface
     Generate,
 }
 
@@ -314,10 +314,10 @@ impl RustWasm {
 with: {{\n\t{with_name:?}: generate\n}}
 ```")
         };
+        self.generated_interfaces.insert(with_name);
         let entry = match remapping {
             InterfaceGeneration::Remap(remapped_path) => {
                 let name = format!("__with_name{}", self.with_name_counter);
-                self.used_with_opts.insert(with_name);
                 self.with_name_counter += 1;
                 uwriteln!(self.src, "use {remapped_path} as {name};");
                 InterfaceName {
@@ -1101,13 +1101,12 @@ impl WorldGenerator for RustWasm {
         let remapped_keys = self
             .with
             .iter()
-            .filter(|(_, r)| matches!(r, InterfaceGeneration::Remap(_)))
             .map(|(k, _)| k)
             .cloned()
             .collect::<HashSet<String>>();
 
         let mut unused_keys = remapped_keys
-            .difference(&self.used_with_opts)
+            .difference(&self.generated_interfaces)
             .collect::<Vec<&String>>();
 
         unused_keys.sort();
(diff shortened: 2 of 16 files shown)


crates/rust/src/lib.rs
```diff
@@ -825,8 +843,19 @@ impl WorldGenerator for RustWasm {
         self.types.analyze(resolve);
         self.world = Some(world);
 
+        let world = &resolve.worlds[world];
+        // Specify that all imports local to the world's package should be generated
+        for (key, item) in world.imports.iter().chain(world.exports.iter()) {
+            if let WorldItem::Interface { id, .. } = item {
+                if resolve.interfaces[*id].package == world.package {
+                    let name = resolve.name_world_key(key);
+                    self.with.insert(name, InterfaceGeneration::Generate);
```
alexcrichton, line 852: Could this avoid overriding anything previous there if something was already specified?

crates/rust/src/lib.rs
```diff
@@ -1049,9 +1080,15 @@ impl WorldGenerator for RustWasm {
         let module_name = name.to_snake_case();
         files.push(&format!("{module_name}.rs"), src.as_bytes());
 
-        let remapping_keys = self.with.keys().cloned().collect::<HashSet<String>>();
+        let remapped_keys = self
+            .with
+            .iter()
+            .filter(|(_, r)| matches!(r, InterfaceGeneration::Remap(_)))
```
alexcrichton, line 1086: Could this work by removing `InterfaceGeneration::Generate` throughout generating bindings? That way erroneous `"foo": generate` keys in `with` could still be warned against

crates/rust/tests/codegen.rs
```diff
@@ -5,6 +5,10 @@ mod codegen_tests {
     macro_rules! codegen_test {
         (wasi_cli $name:tt $test:tt) => {};
         (wasi_http $name:tt $test:tt) => {};
+        (wasi_clocks $name:tt $test:tt) => {};
+        (wasi_filesystem $name:tt $test:tt) => {};
+        (issue569 $name:tt $test:tt) => {};
+        (multiversion $name:tt $test:tt) => {};
```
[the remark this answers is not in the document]
alexcrichton: Could you add an option to the macro which says "just `generate` everything" to use in tests here? It's ok to be a top-level option that's a bit of a hack, but it might also be useful for others transitioning from the previous behavior to this new behavior too.

crates/rust/src/lib.rs
rylev: Interfaces can be encountered more than once, so we don't want to remove the `InterfaceGeneration::Generate` as we go as we will then fail to generate later references to that interface.

I believe I've found a solution that works though.

crates/guest-rust/src/examples/_1_interface_imports.rs
```diff
@@ -24,4 +24,9 @@ crate::generate!({
     "#,
 
     path: "wasi-cli@0.2.0.wasm",
+
+    // specify that this interface dependency should be generated as well.
+    with: {
+        "wasi:cli/environment@0.2.0": generate,
+    }
```
alexcrichton, line 31: Would it actually be possible to have a `dev-dependency` on the `wasi` crate to showcase `with` pointing to `wasi` in this situation?

alexcrichton (approved):
Thanks again for this!

crates/guest-rust/src/examples/_1_interface_imports.rs
rylev: I believe dev-dependencies are not available for `#[cfg(docsrs)]` so it would have to be a regular dependency.

alexcrichton: Ah excellent point, in that case seems fine to leave this.

PR #979, opened 2024-06-19
Haskell guest bindings generator

primoly:


alexcrichton (commented):
Oh wow this is great, thanks for this!

I'm by no means a haskell expert myself so I won't be able to review the generated code myself, but regardless I think it's fine to have this. If you're up for filling out some tests and such I think it'd be reasonable to land this even if it's a little experimental

PR #1002, opened 2024-07-12
add component type WIT file to C# output

dicej:
This is to avoid the need to add binary .o files (which are difficult to audit from a security perspective) to the .NET runtime repository.  I will open related PRs on the `wasm-tools` and `wasm-component-ld` repos to enable the latter to accept WIT files at link (and componentization) time.

Note that I've taken this opportunity to remove the cabi_realloc C file from the output, since we can rely on `wasi-libc` providing an implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1438,10 +1438,10 @@ name = "test-helpers"
 version = "0.0.0"
 dependencies = [
  "codegen-macro",
- "wasm-encoder 0.212.0",
+ "wasm-encoder 0.213.0",
  "wit-bindgen-core",
  "wit-component",
- "wit-parser 0.212.0",
+ "wit-parser 0.213.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):
Looks reasonable to me, but a few questions/points:

* Will CI need to be temporarily disabled for C#? If not should this perhaps be a flag of some kind in the meantime?
* How does the eventual invoker of `wasm-component-ld` know which world is to be used from the `*.wit` output file?

dicej:
>     * Will CI need to be temporarily disabled for C#? If not should this perhaps be a flag of some kind in the meantime?

No; the failure was due to me failing to check everything in; should be fixed now.

>     * How does the eventual invoker of `wasm-component-ld` know which world is to be used from the `*.wit` output file?

My understanding is there will only ever be exactly one world per .wit file produced this way, and I've added code to `wasm-component-ld` to assert that.

alexcrichton:
I think more changes would be needed to print only a single world, right now this prints all the contents of all known packages which if you loaded up `wasi:cli` you'd load at least the `command` world and the `imports` world and I think those would both be printed.

dicej:
> I think more changes would be needed to print only a single world, right now this prints all the contents of all known packages which if you loaded up `wasi:cli` you'd load at least the `command` world and the `imports` world and I think those would both be printed.

Oh interesting.  I guess I assumed that since you can only specify one world as input to `wit-bindgen`, only one world would come out the other end (e.g. with all the `include`d worlds flattened into the the output world).

dicej:
OK, I don't know what the deal is with `cabi_realloc`.  Tests pass fine on my machine, but I've got a bunch of not-yet-merged .NET changes there.  I'll revert that part.

dicej:
> OK, I don't know what the deal is with `cabi_realloc`. Tests pass fine on my machine, but I've got a bunch of not-yet-merged .NET changes there. I'll revert that part.

Ah, now I understand: because my WASIp1->WASIp2 upgrade has not yet been merged into the NativeAOT-LLVM branch of the .NET runtimelab repo, it's still using the `wasm32-wasi` sysroot of `wasi-libc`, which doesn't have `cabi_realloc`.  We'll be able to remove the `cabi_realloc` code out of the C# generator once that's merged.

dicej:
> > I think more changes would be needed to print only a single world, right now this prints all the contents of all known packages which if you loaded up `wasi:cli` you'd load at least the `command` world and the `imports` world and I think those would both be printed.
> 
> Oh interesting. I guess I assumed that since you can only specify one world as input to `wit-bindgen`, only one world would come out the other end (e.g. with all the `include`d worlds flattened into the output world).

It looks to me like that _is_ what's happening.  When I target the `wasi:cli/command` using e.g. `componentize-py` and then run `wasm-tools component wit` on the result, there's only a single world in the output:

```
// ...
package root:component {
  world root {
// ...
```

[cli.txt](https://github.com/user-attachments/files/16236256/cli.txt)

alexcrichton (approved):

PR #1003, opened 2024-07-16
Support to parse WIT files from multiple paths

iawia002:
Background information reference: https://github.com/bytecodealliance/wasmtime/pull/8939#issuecomment-2225909824

This patch allows us to parse WIT files from multiple paths, which is often useful in testing. We can unify all the generated code into a single mod. For example, for Wasmtime's case, we can do it like this:

before:

```rust
wit_bindgen::generate!({
    world: "test-command",
    path: "../wasi/wit",
    generate_all,
});

pub mod config {
    wit_bindgen::generate!({
        path: "../wasi-runtime-config/wit",
        world: "wasi:config/imports",
    });
}

// use the generated code:

use test_programs::wasi::sockets;
use test_programs::config::wasi::config;
```

after:

```rust
wit_bindgen::generate!({
    inline: "
        package wasmtime:test;

        world test {
            include wasi:cli/imports@0.2.0;
            include wasi:config/imports@0.2.0-draft;
            import wasi:http/types@0.2.0;
            import wasi:http/outgoing-handler@0.2.0;
        }
    ",
    paths: ["../wasi/wit", "../wasi-runtime-config/wit"],
    generate_all,
});

// use the generated code:

use test_programs::wasi::sockets;
use test_programs::wasi::config;
```

I am not familiar with the whole process of wit parsing, so I am not sure if there will be any side effects of doing this. I have tested the wasmtime case locally and it does work.

cc @alexcrichton

### DIFF
--- a/crates/guest-rust/macro/src/lib.rs
+++ b/crates/guest-rust/macro/src/lib.rs
@@ -5,7 +5,8 @@ use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering::Relaxed};
 use syn::parse::{Error, Parse, ParseStream, Result};
 use syn::punctuated::Punctuated;
-use syn::{braced, token, Token};
+use syn::spanned::Spanned;
+use syn::{braced, token, LitStr, Token};
 use wit_bindgen_core::wit_parser::{PackageId, Resolve, UnresolvedPackageGroup, WorldId};
 use wit_bindgen_rust::{Opts, Ownership, WithOption};
 
@@ -50,9 +51,9 @@ struct Config {
 /// The source of the wit package definition
 enum Source {
     /// A path to a wit directory
-    Path(String),
+    Paths(Vec<PathBuf>),
     /// Inline sources have an optional path to a directory of their dependencies
-    Inline(String, Option<PathBuf>),
+    Inline(String, Option<Vec<PathBuf>>),
 }
 
 impl Parse for Config {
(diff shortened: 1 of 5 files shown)


crates/guest-rust/src/lib.rs
```diff
@@ -649,6 +649,12 @@
 ///     // folder adjacent to your `Cargo.toml`.
 ///     path: "../path/to/wit",
 ///
+///     // Multiple paths to parse WIT and its dependencies from.
+///     // Usually used in testing, our test suite may want to generate code
+///     // from wit files located in multiple paths within a single mod, and we
+///     // don't want to copy these files again.
+///     paths: ["../path/to/wit1", "../path/to/wit2"],
```
alexcrichton, line 656: Instead of using a new key for this could this syntax be folded into the previous key? For example `path: "..."` or `path: [...]` would be supported

tests/runtime/paths/wasm.rs
```diff
@@ -0,0 +1,12 @@
+wit_bindgen::generate!({
```
alexcrichton, line 1: Adding a runtime test for this is pretty heavyweight and not much runtime functionality is tested here, so could this be added to `crates/rust/tests/*` instead?

crates/guest-rust/macro/src/lib.rs
```diff
@@ -168,31 +178,40 @@ fn parse_source(
...
-        // the path is not correct. Fallback to the original path if normalization fails
-        // (probably return an error somewhere else).
-        let normalized_path = match std::fs::canonicalize(path) {
-            Ok(p) => p,
-            Err(_) => path.to_path_buf(),
-        };
-        let (pkg, sources) = resolve.push_path(normalized_path)?;
-        files.extend(sources);
-        Ok(pkg)
+    let mut parse = |paths: &Vec<String>| -> anyhow::Result<()> {
```
alexcrichton, line 183: Can this be `&[PathBuf]` instead of `&Vec<String>`? That keeps the type-level annotation that these are paths and additionally avoids the dance of `into_os_string().into_string().unwrap()` below too

alexcrichton (commented):
Thanks for this!

iawia002:
Thanks for the review, All comments addressed, PTAL

alexcrichton (approved):
Thanks!

PR #1007, opened 2024-07-16
WIP: rust xcrate issue with free function

pchickey:
rust_xcrate test: pass the free function types defined in another interface


alexcrichton:
A similar issue at https://github.com/bytecodealliance/wasm-tools/issues/1685 was opened on wasm-tools and in debugging that the test case I've mentioned in that issue reproduces the same issue here. I'll work on fixing this over there so I'm going to close this in favor of that issue.

PR #1008, opened 2024-07-23
fix(markdown): update primary link to match behavior from #818

lachieh:
Just aligning the h1 output to match the rest of the "empty anchor before text" pattern for headings


alexcrichton (approved):
Thanks!

lachieh:
No problems! Technically, using `name` on an `a` element is [obsolete as of HTML5 spec](https://html.spec.whatwg.org/multipage/obsolete.html#obsolete-but-conforming-features). It definitely still works to use hash link to target the `name` but the `id` is the standard way to handle this. Would you accept a PR to change these from `name` to `id`?

PR #1009, opened 2024-07-28
fix(markdown): use id attribute instead of obsolete name attribute

lachieh:
Follow up to #1008

Using the `name` attribute on an `a` element is [obsolete as of HTML5 spec](https://html.spec.whatwg.org/multipage/obsolete.htxml#obsolete-but-conforming-features) so I've updated it to the `id` attribute. This might be considered a breaking change since people could be referencing the `name` attribute in css/js selectors but the functionality of referencing an element through a hash identifier still works the same as it does with the `name` attribute.


alexcrichton (approved):
Thanks!

PR #1021, opened 2024-08-09
Add an option to opt-out of `#[used]` statics

alexcrichton:
This is not needed for the wasi_snapshot_preview1 adapter in Wasmtime and is causing issues with LLVM 19, so add an option to turn it off which will be enabled in Wasmtime.


alexcrichton:
cc @dicej do you happen to know of anything off the top of your head which could be causing these C# failures?

```
---- resource_borrow_in_record::run stdout ----
thread 'resource_borrow_in_record::run' panicked at tests\runtime\main.rs:757:18:
pull custom sections from module: decoding a component is not supported
```

AFAIK this PR shouldn't actually affect C# so that seems like tooling used by C# maybe got an update recently which accidentally broke something or something like that?

dicej:
Yeah, this is because WASIp2 support has been merged into the NativeAOT-LLVM branch of the .NET runtimelab repo, so it's using wasm-component-ld, making the wit-bindgen behavior of invoking `wasm-tools component new` redundant.  I'll update it when I'm back at my computer.

dicej:
To keep this from happening in the future, we ought to be able to lock the versions of the .net dependencies somehow.  I'll look into that.

pchickey (approved):

PR #1025, opened 2024-08-12
update C# generator to match latest ILC package

dicej:
This also locks the ILC dependencies to a specific version to avoid future surprises.


alexcrichton (approved):
Thanks!

PR #1027, opened 2024-08-12
C#: make imported functions public by default

dicej:
This allows them to be called from other assemblies.

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -987,6 +987,8 @@ impl InterfaceGenerator<'_> {
             }
         };
 
+        let access = self.gen.access_modifier();
+
         let extra_modifiers = extra_modifiers(func, &camel_name);
 
         let interop_camel_name = func.item_name().to_upper_camel_case();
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):

pavelsavara:
Maybe they could be also `partial` so that we can add methods to the classes.

dicej:
> Maybe the classes/types could be also `partial` so that we can add methods to the classes.

Sounds interesting.  I'm still learning .NET idioms, so perhaps you could explain how that would be useful vs. e.g. extension methods?

pavelsavara:
> > Maybe the classes/types could be also `partial` so that we can add methods to the classes.
> 
> Sounds interesting. I'm still learning .NET idioms, so perhaps you could explain how that would be useful vs. e.g. extension methods?

You can have another hand written .cs file in our project which will be able to modify the type.
For example by adding another interface and implementation. 

Or by overriding existing virtual methods. 
Custom `ToString` implementation my be handy, for example.

https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods

PR #1030, opened 2024-08-15
Allow rust enumerations to be ordered

jtsylve:
Currently enumerations in the generated rust bindings are not ordered, and can only be compared by equality.  This patch adds the `PartialOrd` and `Ord` derives to generated enums.

I'm aware that I can add these using additional_derives for my use case, but that adds it for all types, not just enumerations, and it's generally not appropriate for structs in many cases.


alexcrichton (approved):
Sounds good to me, thanks!

PR #1032, opened 2024-08-19
Use stderr for diagnostics

joonas:
I would like to propose switching the diagnostics output (i.e. which file is being generated) to be sent to stderr instead of stdout.

This would have the benefit of making it any other programs that embed `wit-bindgen` not have to worry about redirecting the diagnostics output to stderr instead so as to prevent their own program output from including things from `wit-bindgen` unless they explicitly choose to do so by including them from stderr.


rvolosatovs (approved):

alexcrichton (approved):
Thanks!

PR #1033, opened 2024-08-20
fix(moonbit): incorrect wasm instructions

peter-jerry-ye:
I made some mistakes when writing the inline wasm instructions. This PR fixes them.


peter-jerry-ye:
Sorry, just found out there're a bunch of other things to be fixed

alexcrichton (approved):

PR #1040, opened 2024-08-29
C#: do not drop resource handles in finalizers

### LINKED ISSUE #1039: C#: the generated finalization logic does not take child resources into account
Currently, the generated resource types have a finalizer, like [this](https://github.com/dotnet/runtime/blob/7ff684af49cdf008bebbe10efcc10a2649795d65/src/libraries/System.Net.Http/src/System/Net/Http/WasiHttpHandler/WasiHttpWorld.wit.imports.wasi.http.v0_2_1.ITypes.cs#L5279-L5281).

This logic is not correct if the resource type may have child resources - this will trap on `drop`. See also this [Discord discussion](https://discord.com/channels/143867839282020352/1141126727028985877/1278079699972788325).

(Side note: it is also not correct due to https://github.com/dotnet/runtime/issues/103522, but that's a different story.)

Options:
1) Drop the finalizer. It means the user code will be exposed to leaks and traps on incorrect usage, but at least they will be determinisic (finalizers are not deterministic). I will mention that _traps_ in managed code are to be avoided, since they can't be converted to managed exceptions (it's like sending `SIGSEGV` instead of raising a `NullReferenceException`).
2) Design some sort of automatic scheme that would close the child resources. This is complicated by the fact that the parent-child relationship is currently only encoded in documentation (it means this can only be done for 'well-known' resource types).
3) Something else?

@dicej, @yowl, @pavelsavara (I would also recommend tagging the .NET interop team)

dicej:
As of this writing, we cannot safely drop a handle to an imported resource from a .NET finalizer because it may still have one or more open child resources. Once WIT has explicit syntax for indicating parent/child relationships, we should be able to use that information to keep track of child resources automatically in generated code, at which point we'll be able to drop them in the correct order from finalizers.

Fixes #1039

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -1463,7 +1463,6 @@ impl InterfaceGenerator<'_> {
 
                         public void Dispose() {{
                             Dispose(true);
-                            GC.SuppressFinalize(this);
                         }}
 
                         [DllImport("{module_name}", EntryPoint = "[resource-drop]{name}"), WasmImportLinkage]
@@ -1474,10 +1473,6 @@ impl InterfaceGenerator<'_> {
                                 wasmImportResourceDrop(Handle);
                                 Handle = 0;
                             }}
-                        }}
-
-                        ~{upper_camel}() {{
-                            Dispose(false);
                         }}
                     "#
                 );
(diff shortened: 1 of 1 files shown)


crates/csharp/src/lib.rs
```diff
@@ -1465,7 +1470,7 @@ impl InterfaceGenerator<'_> {
                         private static extern void wasmImportResourceDrop(int p0);
 
                         protected virtual void Dispose(bool disposing) {{
-                            if (Handle != 0) {{
+                            if (disposing && Handle != 0) {{
```
pavelsavara, line 1473: Please also remove or comment out ~finalizer code at lines 1479-1481. It doesn't do anything after this change but adds the object into finalizer queue.

pavelsavara (approved):

alexcrichton (approved):

PR #993, opened 2024-07-02
Revert "temporarily remove C# `variants` runtime test (#963)"

dicej:
This reverts commit 3ca773953a74dfbc907784072726fd541eba1f2b.

As of https://github.com/dotnet/runtimelab/pull/2609, the "return pointer not aligned" issue which forced us to remove the C# `variants` runtime test case _should_ be resolved. :crossed_fingers:

PR #1055, opened 2024-09-17
[c#] Align lists, free alloc'ed memory before return

yowl:
This PR fixes the alignment of align 8 list elements.  We also now use `NativeMemory.AlignedAlloc` for the allocation on imports, and fix the free of these allocs to be before the return instead of at the end of the block


jsturtevant:
would this fix some of the align issues we were having with the tests?

yowl:
If they are around lists them possibly.  I'm not saying I've found all the align errors, but this one manifested with WebGPU.  Have we disabled some tests because of alignment issues?  If so I can try enabling those.

jsturtevant:
I was thinking of the variants test but looks like we fixed that a couple months ago https://github.com/bytecodealliance/wit-bindgen/pull/993

PR #1058, opened 2024-09-25
fix(moonbit): eliminate warning and fix ownership

peter-jerry-ye:
This PR eliminates all the warnings of the generated code and use `--deny-warn` option in test to enforce this.

This PR also adds build process to the test instead of just checking.

This PR fixes an issue with ownership, which may occur when exporting a data structure and being freed later.


alexcrichton (approved):
Thanks!

PR #1067, opened 2024-10-18
Recommend using `wasm32-wasip2` in Rust example

### LINKED ISSUE #1066: Use build script instead of `wasm-tools` cli in the README example
If I understand correctly, `wasm-tools component new` uses `wit-component` crate under the hood, so we can use it in a `build.rs` directly. By using it, a single `cargo build` will produce the final wasm file and there is no need to install seperate cli components, so I think it makes the process a bit smoother.

HKalbasi:
fix #1066

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,11 +185,11 @@ world host {
 
 ### Guest: Rust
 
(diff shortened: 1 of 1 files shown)


pchickey (approved):
Thanks!

PR #1068, opened 2024-10-18
temporarily remove moonbit from CI language matrix

pchickey:
The moonbit CI job seems to be broken (https://github.com/bytecodealliance/wit-bindgen/actions/runs/11407779008/job/31744559397) due to some change to the compiler. This PR disables the CI jobs temporarily. Maybe when we re-enable it, we should pin a compiler version?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        lang: [c, rust, teavm-java, go, csharp, moonbit]
+        # moonbit removed from language matrix for now - causing CI failures
+        lang: [c, rust, teavm-java, go, csharp]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4


fitzgen (approved):
Alternatively, keep running it in CI but allow the job to fail and for CI to still pass?

pchickey:
Agreed thats preferable, but I couldn't figure out the right incantation to do that on a matrix for just one column, some googling indicates its not supported https://github.com/actions/runner/issues/2347

alexcrichton:
cc @peter-jerry-ye to make sure you see this

peter-jerry-ye:
@alexcrichton Thank you for letting me know. The MoonBit is still an evolving language, and the standard library is still evolving, so I expect there'll be failures from time to time.

How about I create a separate CI file that executes only on pull request (when there are changes for wit-bindgen-moonbit only) and on a weekly basis?

alexcrichton:
Ah ok makes sense. Personally I find intended-to-fail CI pretty hard to wrangle so unfortunately I'm not sure of a good way to manage that. Failures on PRs look scary to contributors "not in-the-know" so I'd be hesitant to do that and a cron-based schedule often sends failure emails to the wrong person (e.g. the commit author rather than those interested). For now it may be best to leave it disabled until MoonBit has stabilized some more?

PR #1069, opened 2024-10-23
fix: MoonBit update and refactor organization

peter-jerry-ye:
This PR contains 

- (BREAKING CHANGE) a reorganize in the generated directory structure, make it more convenient to have a project that exports multiple worlds.
- some language updates according to the MoonBit standard library API changes.
- ci change according to #1068

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,19 +384,26 @@ wasm-tools validate main.component.wasm --features component-model
 MoonBit can be compiled to WebAssembly using [its toolchain](https://moonbitlang.com/download):
 
 ```sh
-moon build --target wasm # -g to keep symbols
+moon build --target wasm # --debug to keep symbols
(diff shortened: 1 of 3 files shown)


.github/workflows/moonbit.yml
```diff
@@ -0,0 +1,47 @@
+name: CI-moonbit
```
alexcrichton, line 1: If you're ok with it I'd personally prefer to avoid adding this to our CI. Contributors won't understand that by default this CI isn't necessary for landing a PR and to ignore it if it's failing.

peter-jerry-ye: Sorry for the late reply. I removed the CI.

alexcrichton (approved):

PR #1070, opened 2024-10-27
[C#] minor improvement to reduce unused using statements

yowl:


jsturtevant:
nice!  This has been on the list under https://github.com/bytecodealliance/wit-bindgen/issues/713 "fix imports so they are only included if needed"

yowl:
There's more 2 do here, but its a start.

crates/csharp/src/lib.rs
```diff
@@ -133,6 +121,8 @@ pub enum FunctionLevel {
 pub struct CSharp {
     opts: Opts,
     name: String,
+    //TODO remove unused
```
jsturtevant, line 124: Why the todo here?

jsturtevant (commented):
Seem the tests are passing. Is there more to do?

crates/csharp/src/lib.rs
yowl: Because its not perfect, the Intefaces in particular still have unused `usings`, e.g. `BindingsWorld.wit.imports.foo.bar.IMyImport.cs` has 
```
// Generated by `wit-bindgen` 0.34.0. DO NOT EDIT!
// <auto-generated />
#nullable enable

using System.Text;
using System.Runtime.InteropServices;

namespace BindingsWorld.wit.imports.foo.bar;

public interface IMyImport {

    public class Address {
        public readonly uint field;
        public readonly bool attribute;

        public Address(uint field, bool attribute) {
            this.field = field;
            this.attribute = attribute;
        }
    }

}

```
Where neither `using` is required.  The generation of the `using`  statements is shared between files which makes it slightly tricky to get this last bit perfect.

yowl: Last commit addresses this.

PR #1081, opened 2024-11-06
Add support for async/streams/futures to Rust generator

dicej:
This adds support for generating bindings which use the [Async ABI](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Async.md) along with the [`stream`, `future`, and
`error-context`](https://github.com/WebAssembly/component-model/pull/405) types.

By default, normal synchronous bindings are generated, but the user may opt-in to async bindings for all or some of the imported and/or exported functions in the target world and interfaces -- provided the default-enabled `async` feature is enabled.

In addition, we generate `StreamPayload` and/or `FuturePayload` trait implementations for any types appearing as the `T` in `stream<T>` or `future<T>` in the WIT files, respectively.  That enables user code to call `new_stream` or `new_future` to create `stream`s or `future`s with those payload types, then write to them, read from them, and/or pass the readable end as a parameter to a component import or return value of a component export.

Note that I've added new `core::abi::Instruction` enum variants to handle async lifting and lowering, but they're currently tailored to the Rust generator and will probably change somewhat as we add support for other languages.

This does not include any new tests; I'll add those in a follow-up commit.

This is ready for review, but I'll leave it in draft mode until the following are complete:

- [ ] Codegen test coverage for new features in the Rust generator
- [ ] Switch to upstream `wasm-tools` once https://github.com/bytecodealliance/wasm-tools/pull/1895 is merged and released
- [ ] Runtime test coverage for new features in the Rust generator.  This requires async/future/stream support in Wasmtime, which I'll soon open a PR for.  Since the Wasmtime support might take a while to land, we could consider deferring this item to a follow-up PR.


dicej:
Oops, forgot to make this a "draft" PR, and GitHub won't let me change it 🤦 .  I'll close this and open another one.

PR #1083, opened 2024-11-07
[C#] Idiomatic formatting

yowl:
Generic types begin with `T`
Methods start with uppercase
some block `{` moved to own line.

draft as behind #1070


crates/csharp/src/lib.rs
```diff
@@ -3292,28 +3333,111 @@ fn is_primitive(ty: &Type) -> bool {
...
 
 trait ToCSharpIdent: ToOwned {
+    fn csharp_keywords() -> Vec<&'static str>;
     fn to_csharp_ident(&self) -> Self::Owned;
+    fn to_csharp_ident_upper(&self) -> Self::Owned;
 }
 
 impl ToCSharpIdent for str {
+    // Source: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/
+    // TODO: Repace with actual keywords
```
silesmo, line 3343: This todo should be have been resolved a long time ago I think?

silesmo:
Marking this as ready for review as #1070 has now been merged.

silesmo (approved):
Thank you! The only thing potentially is if the TODO I commented on should be removed. So leaving it up to you @yowl If that is the case or not.

crates/csharp/src/lib.rs
yowl: Correct, I have now removed this comment.

yowl:
Thanks for the review

PR #1109, opened 2024-12-19
c#: Use a generic Exception type to make handling easier

### LINKED ISSUE #1042: C#: improve WitException.Value type
The pattern where the user code needs to correctly guess type of the `ex.Value` is fragile.

```
            catch (WitException e)
            {
                var error = HeaderErrorToString((HeaderError)e.Value);
                throw new HttpRequestException($"Header validation error: {error}");
            }
```

The cast here could throw invalid cast exception or could give misleading information.

Should we generate `WitHeaderException` and make the conversion on behalf of the user ?

cc @jsturtevant @dicej

jsturtevant:
Partially fixes #1042.

Alternative to #1108.

### DIFF
--- a/crates/csharp/src/lib.rs
+++ b/crates/csharp/src/lib.rs
@@ -588,6 +588,14 @@ impl WorldGenerator for CSharp {
                         NestingLevel = level;
                     }}
                 }}
+
+                {access} class WitException<T>: WitException {{
+                    {access} T TypedValue {{ get {{ return (T)this.Value;}} }}
+
+                    {access} WitException(T v, uint level) : base(v!, level)
+                    {{
+                    }}
+                }}
                 "#,
             )
         }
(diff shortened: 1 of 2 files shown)


crates/csharp/src/lib.rs
```diff
@@ -2952,10 +2961,9 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                         1 => {
                             let mut payload_is_void = false;
                             let mut previous = operands[0].clone();
-                            let mut vars = Vec::with_capacity(self.results.len());
+                            let mut vars: Vec::<(String, Option<String>)> = Vec::<(String, Option<String>)>::with_capacity(self.results.len());
```
dicej, line 2964: I think you only need at most one of these type annotations (i.e. either `let mut vars: Vec::<(String, Option<String>)> = ...` or `let mut vars = Vec::<(String, Option<String>)>::...`).

dicej (approved):
Thanks for doing this!

PR #1110, opened 2024-12-19
c#: Mechanical refactors

jsturtevant:
There are a lot of changes here but no functional changes. Only a couple renames and mostly moved structs to their own modules to help with code organization and make it easier for some one new to understand the code.  There are probably a lot of little improvements that could be made after this in an incremental way to reduce sharing across the modules but this is a good start.  Each move/rename was done in a separate commit with `cargo fmt`/`cargo check` and some tests run against it.

I am suggesting these changes after having spent some time in the code.  It took my awhile to understand the general flow and I think having the code organized in this way will help others understand the code faster.


crates/csharp/src/csharp_ident.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub(crate) trait ToCSharpIdent: ToOwned {
+    fn csharp_keywords() -> Vec<&'static str>;
+    fn to_csharp_ident(&self) -> Self::Owned;
+    fn to_csharp_ident_upper(&self) -> Self::Owned;
+}
+
+impl ToCSharpIdent for str {
+    // Source: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/
+    fn csharp_keywords() -> Vec<&'static str> {
```
dicej, line 11: It would be more efficient to make this return a `&'static [&'static str]` instead of allocating and populating a fresh new new `Vec` every time it's called.  Something like:
```rust
fn csharp_keywords() -> &'static [&'static str] {
    static ARRAY: &[&str] = &["abstract", ...];
    ARRAY
}
```

dicej (approved):

crates/csharp/src/csharp_ident.rs
jsturtevant: Since that requires a bit more refactoring than just moving. I'll follow up on it with another PR

PR #1113, opened 2025-01-07
Fix mismatched tuple ABIs in Rust

### LINKED ISSUE #1112: Rust: corrupted function argument of type list<tuple<s8, s64, s8>>
### Description

```sh
echo "`wit-bindgen -V` | `rustc -V` | `wasmtime -V` | `uname -m`"
wit-bindgen-cli 0.36.0 | rustc 1.83.0 (90b35a623 2024-11-26) | wasmtime 28.0.0 (2e584e841 2024-12-20) | arm64
```

A data value of type `list<tuple<s8, s64, s8>>` is corrupted when passed as a function argument between two components. The components are compiled from Rust source programs using the `wit-bindgen` crate and the `wasm32-wasip2` target.

The two components (host and guest) implement the WIT worlds defined below:
```
package pkg:ns;

world guest {
  export x: func(name3: list<tuple<s8, s64, s8>>);
}

world host {
  import x: func(name3: list<tuple<s8, s64, s8>>);

  export wasi:cli/run@0.2.0;
}

package wasi:cli@0.2.0 {
  interface run {
    run: func() -> result;
  }
}
```

Below are the Rust source programs of the host and guest components, respectively:
```rust
wit_bindgen::generate!({
    world: "host",
    generate_all,
    path: "../wit",
});

struct MyStruct;

impl crate::exports::wasi::cli::run::Guest for MyStruct {
    fn run() -> Result<(), ()> {
        let name3: std::vec::Vec<(i8, i64, i8)> = vec![(0, -1, 0)];
        println!("Sent: {name3:?}");
        crate::x(&name3);
        Ok(())
    }
}

export!(MyStruct);
```

```rust
wit_bindgen::generate!({
    world: "guest",
    generate_all,
    path: "../wit",
});

struct MyStruct;

impl crate::Guest for MyStruct {
    fn x(name3: std::vec::Vec<(i8, i64, i8)>) {
        println!("Received: {name3:?}");
    }
}

export!(MyStruct);
```

Host and guest components are created from the source programs above using the Rust toolchain. The components are composed using wac. The resulting component is executed with Wasmtime. The value of `name3` is printed to standard output before and after it is corrupted.

### Steps to reproduce

Here is a zipped directory that reproduces the bug: 
[integer_tuple_list_corruption.zip](https://github.com/user-attachments/files/18294689/integer_tuple_list_corruption.zip)

1. Unzip `integer_tuple_list_corruption.zip`
2. Build and run with `./verify.sh`
3. Observe that the sent and received values of `name3` differ

I get the following output:
```
Sent: [(0, -1, 0)]
Received: [(0, 255, 0)]
```

### Expected behavior

The received value should match the sent value:
```
Sent: [(0, -1, 0)]
Received: [(0, -1, 0)]
```

### Actual behavior

The outputs differ. 

Note that some modifications to the type of name3 eliminate the unexpected behavior. For example, changing the type of name3 to `list<tuple<s8, s64, string>>`, `list<tuple<s8, s64>>`, or `tuple<s8, s64, s8>` causes the sent and received values to match.

### Additional context

The Rust source programs are derived from programs generated by a differential testing framework for wit-bindgen. The wit definitions are derived from a test case produced by wit-smith.

alexcrichton:
This fixes the mistaken assumption that the tuple ABI in Rust is the same as the component model ABI and generates different code for lifting/lowering lists.

Closes #1112

### DIFF
--- a/crates/core/src/types.rs
+++ b/crates/core/src/types.rs
@@ -30,6 +30,9 @@ pub struct TypeInfo {
     /// Whether this type (transitively) has a list (or string).
     pub has_list: bool,
 
+    /// Whether this type (transitively) has a tuple.
+    pub has_tuple: bool,
+
     /// Whether this type (transitively) has a resource (or handle).
     pub has_resource: bool,
 
@@ -46,6 +49,7 @@ impl std::ops::BitOrAssign for TypeInfo {
         self.owned |= rhs.owned;
         self.error |= rhs.error;
         self.has_list |= rhs.has_list;
+        self.has_tuple |= rhs.has_tuple;
         self.has_resource |= rhs.has_resource;
         self.has_borrow_handle |= rhs.has_borrow_handle;
         self.has_own_handle |= rhs.has_own_handle;
@@ -171,6 +175,7 @@ impl Types {
                 for ty in t.types.iter() {
                     info |= self.type_info(resolve, ty);
                 }
+                info.has_tuple = true;
             }
             TypeDefKind::Flags(_) => {}
             TypeDefKind::Enum(_) => {}
--- a/crates/rust/src/bindgen.rs
+++ b/crates/rust/src/bindgen.rs
@@ -292,7 +292,12 @@ impl Bindgen for FunctionBindgen<'_, '_> {
             return false;
         }
         match ty {
-            Type::Id(id) => !self.gen.gen.types.get(*id).has_resource,
+            // Note that tuples in Rust are not ABI-compatible with component
+            // model tuples, so those are exempted here from canonical lists.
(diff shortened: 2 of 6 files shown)


sunfishcode:
Is the issue here that the Rust tuple types are not `repr(C)`, so they're using the Rust struct layout, which is probably packing the two `s8`s together?

alexcrichton:
Right yeah, and it mistakenly fell through the cracks in the check for "can we just use the raw list pointer as-is" so this PR fixes that by updating that check to explicitly go through manual lift/lower steps for lists-of-tuples

sunfishcode (approved):

PR #1119, opened 2025-01-09
rust: Fix compatibility of generated code with the 2024 edition

alexcrichton:
This commit fixes compatibility of generated code with the Rust 2024 edition coming up next month. The new edition requires that `#[export_name]` is notably an `unsafe` attribute and is wrapped in `#[unsafe(export_name = "...")]` so the generated code is updated to do that. Note that this syntax is allowed on the 2021 edition as of Rust 1.82.0 but is not required, it's only required in the 2024 edition.


jsturtevant (approved):
LGTM

Issue #1080, opened 2024-11-06
c#: Use Spans instead of e.g. byte[] 


jsturtevant:
Use Spans in wit-bindgen-generated code to minimize copies (example [while using streams](https://github.com/dicej/dotnet-wasi-tls/blob/8b8e9f2a971228d355fb8fcb9b7172a985739fcb/src/NetworkStream.cs#L89-L92))

jsturtevant:
#1138 address using `span`/`Memory` for parameters where they are explicitly definded (such as in the [WriteAsync](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.writeasync?view=net-9.0)) but the link above also points to a different type of API where the paramater isn't a list but a value of the number of items to fill.

```
// TODO: should we add a special case to the bindings generator
                    // to allow passing a buffer to IStreams.InputStream.Read and
                    // avoid the extra copy?
                    var result = input.Read(16 * 1024);
```

I think the intent of the comment is that we would have a functions like: 


```csharp
input.Read(buffer.AsSpan());
// use buffer that was filed
```

The [streams wit](https://github.com/WebAssembly/wasi-io/blob/main/wit/streams.wit#L68) looks like 

```
@since(version = 0.2.0)
        read: func(
            /// The maximum number of bytes to read
            len: u64
        ) -> result<list<u8>, stream-error>;
```

We could return a Span here but I don't know how we detect that we should re-write the `input.Read` to special case it for taking a span and filling it.

jsturtevant:
The hang up for me here is how do translate the WIT:

```
read: func(len: u64) -> result<list<u8>, stream-error>;
```

Into a function, that we could pass a buffer so that it can be filled, ie `input.Read(buffer.AsSpan());`.  

Other than the specific wasi interface name there aren't any real heuristics to know that a function signature of `read: func(len: u64) -> result<list<u8>, stream-error>;` should be translated to `input.Read(buffer.AsSpan())`. But maybe I am missing something?

I am going to propose to not worry about his for now, since the API will change with WASI 0.3 and there is also discussions around a FillableArray type in 0.3.x that would solve this problem.

PR #1122, opened 2025-01-10
c#: Remove Copy of data during import call for base types

jsturtevant:
When calling an import we don't need to copy data to a new structure for canonical types. This avoid the extra copy of data in this scenario.

This isin prep for some more changes with https://github.com/bytecodealliance/wit-bindgen/issues/1080

### DIFF
--- a/crates/csharp/src/function.rs
+++ b/crates/csharp/src/function.rs
@@ -586,9 +586,9 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                     Direction::Import => {
                         let ptr: String = self.locals.tmp("listPtr");
                         let handle: String = self.locals.tmp("gcHandle");
-                        // Dispite the name GCHandle.Alloc here this does not actually allocate memory on the heap. 
-                        // It pins the array with the garbage collector so that it can be passed to unmanaged code
-                        // It is required to free the pin after use which is done in the Cleanup section 
+                        // Despite the name GCHandle.Alloc here this does not actually allocate memory on the heap. 
+                        // It pins the array with the garbage collector so that it can be passed to unmanaged code.
+                        // It is required to free the pin after use which is done in the Cleanup section.
                         uwrite!(
                             self.src,
                             "
--- a/crates/csharp/src/function.rs
+++ b/crates/csharp/src/function.rs
@@ -582,21 +582,23 @@ impl Bindgen for FunctionBindgen<'_, '_> {
 
             Instruction::ListCanonLower { element, realloc } => {
                 let list: &String = &operands[0];
-                let (_size, ty) = list_element_info(element);
-
                 match self.interface_gen.direction {
                     Direction::Import => {
-                        let buffer: String = self.locals.tmp("bufferPtr");
+                        let ptr: String = self.locals.tmp("listPtr");
+                        let handle: String = self.locals.tmp("gcHandle");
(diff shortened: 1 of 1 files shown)


crates/csharp/src/function.rs
```diff
@@ -1063,7 +1064,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
 
                 match direction {
                     Direction::Import => {
-			let import_name = self.interface_gen.type_name_with_qualifier(&Type::Id(id), true);
+                        let import_name = self.interface_gen.type_name_with_qualifier(&Type::Id(id), true);
```
jsturtevant, line 1067: Some how these two lines had tabs, while rest of the file was spaces

crates/csharp/src/function.rs
```diff
@@ -581,21 +581,24 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
-                let list = &operands[0];
-                let (_size, ty) = list_element_info(element);
-
+                let list: &String = &operands[0];
                 match self.interface_gen.direction {
                     Direction::Import => {
-                        let buffer: String = self.locals.tmp("buffer");
+                        let ptr: String = self.locals.tmp("listPtr");
+                        let handle: String = self.locals.tmp("gcHandle");
+                        // Dispite the name GCHandle.Alloc here this does not actually allocate memory on the heap.
```
dicej, line 589:
```suggestion
                        // Despite the name GCHandle.Alloc here this does not actually allocate memory on the heap. 
```

crates/csharp/src/function.rs
```diff
@@ -581,21 +581,24 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
-                let (_size, ty) = list_element_info(element);
-
+                let list: &String = &operands[0];
                 match self.interface_gen.direction {
                     Direction::Import => {
-                        let buffer: String = self.locals.tmp("buffer");
+                        let ptr: String = self.locals.tmp("listPtr");
+                        let handle: String = self.locals.tmp("gcHandle");
+                        // Dispite the name GCHandle.Alloc here this does not actually allocate memory on the heap. 
+                        // It pins the array with the garbage collector so that it can be passed to unmanaged code
```
dicej, line 590:
```suggestion
                        // It pins the array with the garbage collector so that it can be passed to unmanaged code.
```

crates/csharp/src/function.rs
```diff
@@ -581,21 +581,24 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
+                        // It is required to free the pin after use which is done in the Cleanup section
```
dicej, line 591:
```suggestion
                        // It is required to free the pin after use which is done in the Cleanup section.
```

dicej (approved):
LGTM module minor spelling and punctuation nits.  Thanks!

jsturtevant:
A quick note for myself in the future, I did run some non scientific tests to compare speed of the various implementations:

- using stackalloc (previous method) 
- using the ptr to the array (method implemented here)

The speed of execution difference varied but was never more than a few microseconds difference.  The biggest and probably more important change here is the extra copy and allocation isn't happening on the stack.  With larger byte arrays would cause a memory exception.  With this change large byte arrays are not an issue.

jsturtevant:
failure doesn't look related: `@github-actions[bot]
Test (macos-latest, teavm-java)`

PR #1124, opened 2025-01-14
drop borrows earlier for async exports in Rust generator

dicej:
While testing runtime borrow enforcement in Wasmtime, I discovered that the guest code generated by `wit-bindgen` was not dropping borrows early enough for async exports.  Specifically, all borrows need to be dropped before `task.return` is called, which is what this patch does.

### DIFF
--- a/crates/rust/src/bindgen.rs
+++ b/crates/rust/src/bindgen.rs
@@ -20,6 +20,7 @@ pub(super) struct FunctionBindgen<'a, 'b> {
     pub import_return_pointer_area_align: usize,
     pub handle_decls: Vec<String>,
     always_owned: bool,
+    pub async_result_name: Option<String>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/rust/src/bindgen.rs
```diff
@@ -877,10 +879,8 @@ impl Bindgen for FunctionBindgen<'_, '_> {
 
             Instruction::CallInterface { func, .. } => {
                 if self.async_ {
-                    let tmp = self.tmp();
-                    let result = format!("result{tmp}");
-                    self.push_str(&format!("let {result} = "));
-                    results.push(result);
+                    self.push_str(&format!("let result = "));
```
dicej, line 882: This change wasn't necessary, but I figured I'd simplify it a bit since there's no chance of a name conflict (i.e. we don't need to use `self.tmp()` here).

PR #1129, opened 2025-01-16
fix lowering results for async exports

dicej:
Previously, we were either dropping the result too early (e.g. `wit_bindgen_rt::async_support::ErrorContext`) or not at all (e.g. `String` and `Vec`) due to the lowering code expecting to pass ownership to the caller, which would later free memory using a post-return function.  But async exports don't have post-return functions; they use `task.return` instead, and they should free any memory (and/or drop handles) after `task.return` returns.  So now we do that!

### DIFF
--- a/crates/core/src/abi.rs
+++ b/crates/core/src/abi.rs
@@ -1474,11 +1474,12 @@ impl<'a, B: Bindgen> Generator<'a, B> {
     }
 
     fn list_realloc(&self) -> Option<&'static str> {
-        // Lowering parameters calling a wasm import means
-        // we don't need to pass ownership, but we pass
-        // ownership in all other cases.
-        match (self.variant, self.lift_lower) {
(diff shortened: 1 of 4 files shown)


crates/core/src/abi.rs
```diff
@@ -1477,8 +1477,9 @@ impl<'a, B: Bindgen> Generator<'a, B> {
         // Lowering parameters calling a wasm import means
         // we don't need to pass ownership, but we pass
         // ownership in all other cases.
-        match (self.variant, self.lift_lower) {
-            (AbiVariant::GuestImport, LiftLower::LowerArgsLiftResults) => None,
+        match (self.variant, self.lift_lower, self.async_) {
```
dicej, line 1480: Just realized I should update the comment above; will do.

dicej:
Oof, looks like I broke the codegen tests 🤦 ; will fix.

alexcrichton (approved):

PR #1144, opened 2025-01-24
c#: Use stack when the list is smaller than a given value

### LINKED ISSUE #1012: C#: use `stackalloc` where appropriate for caller-allocated buffers
Per https://github.com/dotnet/runtimelab/pull/2614#discussion_r1699988599, using `stackalloc` would likely be more efficient and avoid the need for GC pinning.

jsturtevant:
fixes #1012

### DIFF
--- a/crates/csharp/src/function.rs
+++ b/crates/csharp/src/function.rs
@@ -853,13 +853,27 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 //TODO: wasm64
                 let align = self.interface_gen.csharp_gen.sizes.align(element).align_wasm32();
 
+                let (array_size, element_type) = crate::world_generator::dotnet_aligned_array(
+                    size,
+                    align,
+                );
+                let ret_area = self.locals.tmp("retArea");
+
                 self.needs_cleanup = true;
                 uwrite!(
                     self.src,
                     "
-                    var {buffer_size} = {size} * (nuint){list}.Count;
-                    var {address} = NativeMemory.AlignedAlloc({buffer_size}, {align});
-                    cleanups.Add(()=> NativeMemory.AlignedFree({address}));
+                    void* {address};
+                    if (({size} * {list}.Count) < 1024) {{
+                        var {ret_area} = stackalloc {element_type}[({array_size}*{list}.Count)+1];
+                        {address} = (void*)(((int){ret_area}) + ({align} - 1) & -{align});
+                    }} 
+                    else
+                    {{
+                        var {buffer_size} = {size} * (nuint){list}.Count;
+                        {address} = NativeMemory.AlignedAlloc({buffer_size}, {align});
+                        cleanups.Add(()=> NativeMemory.AlignedFree({address}));
+                    }}
 
                     for (int {index} = 0; {index} < {list}.Count; ++{index}) {{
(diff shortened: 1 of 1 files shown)


crates/csharp/src/function.rs
```diff
@@ -853,13 +853,27 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
+                    void* {address};
+                    if (({size} * {list}.Count) < 1024) {{
+                        var {ret_area} = stackalloc {element_type}[({array_size}*{list}.Count)+1];
+                        {address} = (void*)(((int){ret_area}) + ({align} - 1) & -{align});
+                    }} 
+                    else
+                    {{
+                        var {buffer_size} = {size} * (nuint){list}.Count;
+                        {address} = NativeMemory.AlignedAlloc({buffer_size}, {align});
+                        cleanups.Add(()=> NativeMemory.AlignedFree({address}));
```
pavelsavara, line 875: Wishful thinking - not blocked for this PR:

It would be awesome if we could see this in context of some generated sample.
Maybe we can have wit->C# integration test in which the C# file is committed into git ?
That would show up as delta in the generated code in the PR.

alexcrichton:
I'm happy to rubber-stamp these (or ask @dicej) to rubber stamp them, but @jsturtevant do you have someone else in mind to review C#-related PRs? I'm happy to add someone else to this repository for reviewing things to assist with assignment/merging/etc.

(mostly want to clarify your expectations around reviewing since I'm not able to review C# code myself)

crates/csharp/src/function.rs
jsturtevant: There are tons of integration tests in https://github.com/bytecodealliance/wit-bindgen/tree/main/tests

Currently we don't check any into source, but having something like the that does have some value.  I typically pull a PR and take a look at the source changes if I am unsure so would find this useful.

@alexcrichton  any thoughts on this? I suppose the Rust version is a macro so you don't really have source to check in?

jsturtevant:
> I'm happy to rubber-stamp these (or ask @dicej) to rubber stamp them, but @jsturtevant do you have someone else in mind to review C#-related PRs? I'm happy to add someone else to this repository for reviewing things to assist with assignment/merging/etc.
> 
> (mostly want to clarify your expectations around reviewing since I'm not able to review C# code myself)

Typically try to get a review from the C# maintainers: @silesmo, @yowl or @dicej. 

@pavelsavara is our upstream .NET contributor that could offer C#/.net advice too. @pavelsavara are you interested in joining the repo for c# reviews?

alexcrichton:
Ok sounds good 👍 

Mostly wanted to make sure you weren't blocking on my and I wasn't aware of it!

pavelsavara:
> @pavelsavara are you interested in joining the repo for c# reviews?

Yes

crates/csharp/src/function.rs
jsturtevant: I was thinking that we didn't need to have every test in `tests/codegen` but maybe something like the `flavorful` example checked in?  I think it might be overwhelming to have to many files changed in a PR. Anyways I will open an issue to track.

pavelsavara (approved):

Issue #1143, opened 2025-01-24
c#: Update the allocation clean up code to avoid using closure which will allocate on the heap


jsturtevant:
> Also note that cleanups.Add(()=> NativeMemory.AlignedFree({address})); will allocate closure.
> 
> Agree. I would consider handling this in a stack allocated `struct` that implements `IDisposable` or a `ref struct` with a `Dispose()` method. Note that Roslyn allows a `ref struct` with a `Dispose()`, not implementing `IDisposable`, to be used in a `using` statement. 

 _Originally posted by @AaronRobinsonMSFT in [#1012](https://github.com/bytecodealliance/wit-bindgen/issues/1012#issuecomment-2610430610)_

jsturtevant:
for some more possible design options: https://github.com/bytecodealliance/wit-bindgen/pull/1145#discussion_r1937031162

PR #1145, opened 2025-01-28
c#: Handle Cabi realloc post return 

### LINKED ISSUE #1116: c#: handle cabi_post properly
Currently we don't free any of the resources that should be freed: https://github.com/bytecodealliance/wit-bindgen/blob/19f259b191537719881c430ff6c5cfda933d653b/crates/csharp/src/interface.rs#L464-L468

jsturtevant:
fixes #1116 

This pull request includes several changes to the `FunctionBindgen` implementation in the `crates/csharp/src/function.rs` file, primarily focusing on memory management and cleanup handling. Additionally, there are modifications in the `InterfaceGenerator` implementation in the `crates/csharp/src/interface.rs` file to support post-return handling for guest exports.

The makes sure memory that was allocated for export functions is removed in the ABI's post return. Prior to this change exports where pinning memory but not releasing the pins in most cases.  It also fixes a case in the List import where the list was being lowered and clean up code was executing but since memory holding it was pinned the memory was still there.  This makes sure that the memory is exported for long enough time and is cleaned up in later calls.


crates/csharp/src/function.rs
```diff
@@ -868,6 +870,10 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                     }}
                     "
                 );
+                if realloc.is_none() {
+                    uwrite!(self.src,
+                    "cleanups.Add(()=> NativeMemory.AlignedFree({address}));")
```
pavelsavara, line 875: Naive question, why do we keep `cleanups` list instead of generating this line after the call to the host ?

crates/csharp/src/function.rs
```diff
@@ -1033,19 +1039,72 @@ impl Bindgen for FunctionBindgen<'_, '_> {
             Instruction::Malloc { .. } => unimplemented!(),
 
             Instruction::GuestDeallocate { .. } => {
-                uwriteln!(self.src, r#"Console.WriteLine("TODO: deallocate buffer for indirect parameters");"#);
+                // the original alloc here comes from cabi_realloc implementation (wasi-libc in .net)
+                uwriteln!(self.src, r#"NativeMemory.Free((void*){});"#, operands[0]);
```
pavelsavara, line 1043: Which test would be executing this new line ?

crates/csharp/src/function.rs
jsturtevant: In the case of a lower during an import, we would still need to clean up the lowered list.  We could re-write this to use an array/stackalloc and fixed in the case of an import.

crates/csharp/src/function.rs
jsturtevant: The runtime tests that exercises this is https://github.com/bytecodealliance/wit-bindgen/tree/main/tests/runtime/resource_aggregates

crates/csharp/src/function.rs
pavelsavara: I suggest that we try to kill as many by using `fixed` fist and then see what scenarios still need another approach.

I think the nested arrays in `public static unsafe void ListParam4(List<List<string>> a)` is the nasty example.
I'm confused why the buffers are not freed by the receiving end ? Could you please explain (or link the spec).

It seems to me that if we really need cleanup with dynamic number of elements, perhaps all those elements are just buffers. At lest current code looks like it. 
And so `cleanups` collection could be just list of pointers instead of more expensive delegates.

```
var cleanups = new List<IntPtr>();

var address3 = NativeMemory.AlignedAlloc(10,4);
cleanups.Add((IntPtr)address3);

/// more code here ...

{
    Marshal.FreeHGlobal(cleanup);
}
```

jsturtevant: > And so cleanups collection could be just list of pointers instead of more expensive delegates.

This one is on me, We had two different clean up code paths a few pr's back and I consolidated both of them and used a delegate because we had two different types and it was simple way to handle it. I hadn't really considered the allocations from the closure. I've got https://github.com/bytecodealliance/wit-bindgen/issues/1143 to track cleaning it up.

jsturtevant: > I'm confused why the buffers are not freed by the receiving end ? Could you please explain (or link the spec).

https://github.com/bytecodealliance/wit-bindgen/blob/b19ef10a8fa96c7f2f65e043b05ae9490d1a9476/crates/core/src/abi.rs#L226-L251

There are couple scenarios describe in that doc how to handle lowering but this particular code change has two scenarios:

1. first one is when its an import (we are calling the host/another component).  We can lower our type and then clean it up after the call to the external component is complete. - i.e `realloc.is_none()`
2. We are in an export (meaning the host or another component is calling us).  In this case, we need to allocate and lower but that lowered value needs to live past our function call and so the ABI will call us with cabi_post to clean it up.  i.e `realloc.is_some()`

This section has some notes too but I find it harder to parse: https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#canonical-abi

jsturtevant: > I suggest that we try to kill as many by using fixed first and then see what scenarios still need another approach.

Agreed, there was a mix different ways this was done originally... slowly trying to consolidate them all as we learn. I've opened https://github.com/bytecodealliance/wit-bindgen/issues/1150 to track that work

pavelsavara: Perhaps we could do `cleanups` even better

```cs
int cleanupCount = ...; // calculate dynamic count on the runtime from the parameters
int cleanupPos = 0;

nint* cleanups = stackalloc IntPtr[cleanupCount];

var address3 = NativeMemory.AlignedAlloc(10, 4);
cleanups[cleanupPos++]= (nint)address3;

/// more code here ...

for(int i = 0; i < cleanupPos; i++)
{
    var ptr = cleanups[i];
    Marshal.FreeHGlobal(ptr);
}
```

this would avoid `new List<IntPtr>()` too, which is again managed data structure.
`List<T>` would also re-allocate internal buffer as we add more than it's initial buffer size.

Also, if the call to the host is returning some errors, which we translate into exceptions, the cleanup needs to happen before we `throw` or inside `finally`. Note that try/finally is not "cheap", especially on WASM.

jsturtevant: add a note on https://github.com/bytecodealliance/wit-bindgen/issues/1143 to track that suggestion.

crates/csharp/src/function.rs
```diff
@@ -859,22 +862,37 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
+                            if (({size} * {list}.Count) < 1024) {{
+                                var {ret_area} = stackalloc {element_type}[({array_size}*{list}.Count)+1];
+                                {address} = (void*)(((int){ret_area}) + ({align} - 1) & -{align});
+                            }} 
+                            else
+                            {{
+                                var {buffer_size} = {size} * (nuint){list}.Count;
+                                {address} = NativeMemory.AlignedAlloc({buffer_size}, {align});
+                            }}
+                            cleanups.Add(()=> NativeMemory.AlignedFree({address}));
```
yowl, line 880:
```suggestion
                            cleanups.Add(() => NativeMemory.AlignedFree({address}));
```
Nit: space

yowl, line 880: This should be for the >= 1024 side of the if I think?

pavelsavara: And unit test for that use-case would be good.

jsturtevant: messed up on my rebase, good catch. I've also added a scenario to the integration tests that force the native alloc to happen.

pavelsavara:
Something went wrong with exports, I see
```
        [UnmanagedCallersOnly(EntryPoint = "cabi_post_test:lists/test#list-result")]
        public static unsafe void cabi_post_wasmExportListResult(nint p0) {
            NativeMemory.Free((void*)BitConverter.ToInt32(new Span<byte>((void*)(p0 + 0), 4)));

        }
```

It doesn't call the implementation.

Maybe we are not unit testing those exports ?

jsturtevant:
> Something went wrong with exports, I see
> 
> ```cs
>         [UnmanagedCallersOnly(EntryPoint = "cabi_post_test:lists/test#list-result")]
>         public static unsafe void cabi_post_wasmExportListResult(nint p0) {
>             NativeMemory.Free((void*)BitConverter.ToInt32(new Span<byte>((void*)(p0 + 0), 4)));
> 
>         }
> ```
> 
> It doesn't call the implementation.
> 
> Maybe we are not unit testing those exports ?

This is expected and what is being added by this PR. It is the ABI's `post-return` call when an alloc happens in the main export since the memory needs to live past the export call.  

from https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#canonical-abi:

```
The (post-return ...) option may only be present in canon lift when async is not present and
specifies a core function to be called with the original return values after they have finished
being read, allowing memory to be deallocated and destructors called. This immediate is
always optional but, if present, is validated to have parameters matching the callee's return tye and empty results.
```


If you look above that section of code you will find the matching export: 

```csharp
       [UnmanagedCallersOnly(EntryPoint = "test:lists/test#list-result")]
        public static unsafe nint wasmExportListResult() {

            byte[] ret;
            ret = TestImpl.ListResult();

            var byteLength = (1) * ret.Length;
            var address = NativeMemory.Alloc((nuint)(byteLength));
            ret.AsSpan().CopyTo(new Span<byte>(address,byteLength));
            BitConverter.TryWriteBytes(new Span<byte>((void*)(ptr + 4), 4), ret.Length);
            BitConverter.TryWriteBytes(new Span<byte>((void*)(ptr + 0), 4), (int)(address));
            return ptr;

        }

        [UnmanagedCallersOnly(EntryPoint = "cabi_post_test:lists/test#list-result")]
        public static unsafe void cabi_post_wasmExportListResult(nint p0) {
            NativeMemory.Free((void*)BitConverter.ToInt32(new Span<byte>((void*)(p0 + 0), 4)));

        }
```

pavelsavara:
> this is expected and what is being added by this PR.

Oh, sorry. I misunderstood this, I didn't know this is per export.

This will do native-> managed transition just to call native `free` from dotnet. Pretty expensive.

> `This immediate is always optional`

Does it mean we could avoid emitting that and host would call `cabi_realloc` export of our `wasi-libc` instead ?
(much cheaper)

What is the benefit of emitting it ?

> ```cs
>             NativeMemory.Free((void*)BitConverter.ToInt32(new Span<byte>((void*)(p0 + 0), 4)));
> ```

Final nit, why we are doing the `BitConverter.ToInt32` dance ?
Is this the same ?

```cs
[UnmanagedCallersOnly(EntryPoint = "cabi_post_test:lists/test#list-result")]
public static unsafe void cabi_post_wasmExportListResult(void **p0)
{
    NativeMemory.Free(*p0);
}
```

jsturtevant:
>This will do native-> managed transition just to call native free from dotnet. Pretty expensive.

I not sure I understand.  Could you suggest another approach?

>This immediate is always optional
> Does it mean we could avoid emitting that and host would call cabi_realloc export of our wasi-libc instead ?
(much cheaper)
> What is the benefit of emitting it ?

I don't think not emitting the post_return is something we have control over. We only generate these exports when `if abi::guest_export_needs_post_return(self.resolve, func)` returns true.

I believe it is already calling wasi-libc `realloc` in some situations (i.e. indirect parameters and then needs us to free it with `Instruction::GuestDeallocate`) but in the other cases like lowering lists, variants, and strings, we need to do the allocation, and then are called latter via `GuestDeallocateList`, `GuestDeallocateVariant`, and `GuestDeallocateString`

> Final nit, why we are doing the BitConverter.ToInt32 dance ?

this is coming from one of the other instructions that are generated via the wit-bindgen abi (https://github.com/bytecodealliance/wit-bindgen/blob/5d8fa6196b6382a90592c42aa63ad73f17760797/crates/csharp/src/function.rs#L400-L407)

 Maybe we don't need `BitConverter.ToInt32` in other cases?  Maybe that something else to clean up (in a separate pr)

pavelsavara:
> > This will do native-> managed transition just to call native free from dotnet. Pretty expensive.
> 
> I not sure I understand. Could you suggest another approach?

- In Mono we generate this as C export and 
- then we use reflection to find managed code (only once I think)
- enter GC boundary
- interpret the managed method
- leave GC boundary
- eventually call to libc `free`
- enter GC boundary on way back
- leave GC when leaving the managed code
- return from the component code

In NAOT, it could be lot better, but I think it will also cross the GC boundary in same way.

>  Could you suggest another approach?

I thought: what would the host do if we didn't generate that at all.

> Maybe we don't need `BitConverter.ToInt32` in other cases? Maybe that something else to clean up (in a separate pr)

OK, different PR. And we could consider how to deal with the other data types for which `void**` parameter is not the right choice.

pavelsavara (approved):

pavelsavara:
> > Could you suggest another approach?

WASI exports require WASI SDK and c-lang compilation anyway. 
We could also generate C code instead, just for this particular export type.

And then `<NativeFileReference Include="wasi_exports.c" />` in csproj

jsturtevant:
> > > Could you suggest another approach?
> 
> WASI exports require WASI SDK and c-lang compilation anyway. We could also generate C code instead, just for this particular export type.
> 
> And then `<NativeFileReference Include="wasi_exports.c" />` in csproj

we did something similar in the past for cabi_realloc originally. It sounds like this would be a good improvement, so I opened #1165

PR #1149, opened 2025-01-30
feat: support error context in stream/error operations

vados-cosmonic:


crates/guest-rust/rt/src/async_support.rs
```diff
@@ -187,26 +190,49 @@ mod results {
     pub const CANCELED: u32 = 0;
 }
 
+/// Result of awaiting a asynchronous read or write
+pub enum AsyncWaitResult {
```
dicej, line 194: Could you add a `#[doc(hidden)]` to this since it's only used by other `#[doc(hidden)]` APIs and shouldn't be used directly by application code?

crates/guest-rust/rt/src/async_support.rs
```diff
@@ -187,26 +190,49 @@ mod results {
...
+}
+
+impl AsyncWaitResult {
+    /// Interpret the results from an async operation that is known to *not* be blocked
+    fn from_nonblocked_async_result(v: u32) -> Self {
+        match v {
+            results::CLOSED | results::CANCELED => Self::End,
+            v => {
+                if v & (1 << 31) != 0 {
+                    Self::Error(v ^ results::CLOSED)
```
dicej, line 210: The use of XOR and `1 << 31` as another way of saying `results::CLOSED` is confusing to me.  I would have expected something like `if v & results::CLOSED != 0 { Self::Error(v & !results::CLOSED) } else { ... }`

dicej (commented):
I haven't had time to look at this super thoroughly yet, but looking good so far.  See comments inline.

crates/guest-rust/rt/src/async_support.rs
vados-cosmonic: Yup, will add -- thanks

crates/guest-rust/rt/src/async_support.rs
vados-cosmonic: ah yeah this is a bit confusing -- yeah that's a bunch simpler -- will go with that, I think when I wrote that I forgot the 30 bits were guaranteed by spec not to touch/overlap

dicej (commented):
LGTM, but let's hold off merging this until we have some end-to-end test coverage in (your or my branch of) Wasmtime, e.g. tests that exercise closing with an error and verifying the reader receives it.  We should test both the "local" case (i.e. when the same component instance has both ends of the stream/future) and the "remote" case (where the reader and writer ends are owned by different component instances).

vados-cosmonic:
Just a note: there are now working tests that use the functionality in here:

https://github.com/bytecodealliance/wasip3-prototyping/pull/3

dicej (approved):

Issue #1159, opened 2025-02-07
Scala.js binding generator


vigoo:
I have implemented a Scala.js binding generator in my fork (https://github.com/vigoo/wit-bindgen-scalajs/pull/1) which genreates Scala code that, when compiled with Scala.js, produces JS code compatible with jco/componentize.js

Is this something that could be merged to the official `wit-bindgen` codebase? If yes, I will open a pull request.

(Otherwise I'm going to publish the fork as a separate binding generator CLI only for scala.js)

alexcrichton:
Hello! I think it'd be reasonable to add here so long as it doesn't add undue burden to developing the other languages. For example it shouldn't introduce large files or new major unconditional runtime deps or something like that. Ideally it also isn't too slow to build/test on CI. 

Additionally for now I'd ask that you'd be ok being pinged about any issues, and ok with how it might be disabled temporarily in CI (or broken in the code) if something else urgent needs to land. In such a situation you'd alway be pinged though and given time to fix things before a new release is made or similar.

Issue #1158, opened 2025-02-06
c#: CI is extremely slow


alexcrichton:
Since C# CI was introduced it's been quite slow, but I realize that I haven't opened an issue about this. CI for C#, specifically on Windows, takes 30+ minutes whereas all other platforms finish in ~10. 

For C# folks @jsturtevant, @yowl, @silesmo, would y'all be able to help optimize this? Is there something that can be done to bring the total time closer to the 10 minute range?

pavelsavara:
I guess the overall C# time is dominated by AOT ? Perhaps we could switch to Mono for CI ?

https://github.com/bytecodealliance/wit-bindgen/issues/1079

I realize that was not easy with Net9, but Net10 preview1 bits are due in few days. We can try again.

yowl:
I'm surprised that for these small programs it is quite so slow.  Are there logs?

alexcrichton:
CI logs are [from the `main.yml`](https://github.com/bytecodealliance/wit-bindgen/actions/workflows/main.yml) workflow. Any green run which ran everything should suffice and the slow builder is "Test (windows-latest, csharp)". The half hour is spent in [one step](https://github.com/bytecodealliance/wit-bindgen/actions/runs/13146113548/job/36684502693#logs) which only has limited logs. 

That being said I would expect that running the same command would reproduce this locally pretty easily. If it doesn't then I'd be curious to understand how a test run is so fast on your machine and not on CI!

I can also clarify that at least from my perspective everything about building C# is a large black box to me, so I can't personally answer many questions myself as I mostly just rubber stamped things around CI integration.

pavelsavara:
It could be also download of the nuget packages `Microsoft.DotNet.ILCompiler.LLVM` and `runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM`. Maybe there is way how to add that to the github action cache ?

yowl:
I would expect that to only affect the first AOT compiled test (unless we have a nuget.config with a local storage specified).  The subsequent tests should then be fast, but there are still many that take over a minute.

yowl:
> unless we have a nuget.config with a local storage specified

We do, trying an alternatvei

yowl:
Didn't make much difference sharing the nuget location.  I would like to see finer timings, I'm not sure how much of each runtime test is rustc, the compiled rust program, the c# compilation, and running the wasm.

yowl:
[...]

PR #1171, opened 2025-02-25
Scala.js binding generator

vigoo:
Resolves https://github.com/bytecodealliance/wit-bindgen/issues/1159

I've only enabled the tests for linux, because on the macos runner it was extremely slow and on windows the `setup-scala` action does not seem to properly setup scala.


tanishiking:
Hi @vigoo! FYI @sjrd @alexcrichton 

I'm one of the developers working on the Scala.js Wasm backend 🙂  
I was wondering—does this code generator assume a frontend Wasm setup using `jco`?  

Currently, Scala.js's Wasm backend only generates modules that depend on JavaScript, meaning they can't be executed in standalone runtimes like Wasmtime. It also doesn't export linear memory or `cabi_realloc`, nor does it generate instructions that directly manipulate linear memory (e.g., `i32.store`/`i32.load`). Because of this, it shouldn't be able to support the Component Model—at least in standalone Wasm runtimes. Does `jco` provide any helpers for these operations? (I'm not very familiar with `jco` 😞)  

Aside from that, there's an ongoing effort to support the Wasm Component Model and WASI in standalone Wasm runtimes: https://github.com/scala-js/scala-js/issues/5121.  

In our implementation, instead of exposing memory manipulation APIs at the source language level and generating Scala code that interacts with them via `wit-bindgen`, we take a different approach: we keep memory operations hidden from the source language and delegate them to the linker backend. This means the conversion between Scala objects and the Canonical ABI is handled by the linker backend, rather than in user code.  
A lot of Scala.js's standard libraries don't work in standalone Wasm runtimes yet, but we're planning to re-implement some of them using pure Scala and WASIp2.  

So, if this implementation is aimed at frontend Wasm using `jco` (and if `wit-bindgen` community accepts it), great work! This would mean Scala.js will have `wit-bindgen` support for both frontend Wasm (`jco`) and server-side Wasm.  

If the goal is to develop a `wit-bindgen` for server-side Wasm, then aligning it with our approach could be great. 🙂  Would love to hear your thoughts!

tanishiking:
Oops, sorry you've already mentioned that!

> genreates Scala code that, when compiled with Scala.js, produces JS code compatible with jco/componentize.js
https://github.com/bytecodealliance/wit-bindgen/issues/1159#issue-2838584935

.github/workflows/main.yml
```diff
@@ -112,12 +112,16 @@ jobs:
       with:
         tinygo-version: 0.31.0
 
+    - uses: olafurpg/setup-scala@v11
```
xuwei-k, line 115: `olafurpg/setup-scala` is deprecated. 


https://github.com/sbt/setup-sbt

vigoo: Ah thanks. I've seen mentioning setup-scala is deprecated _because sbt is now included in the runners_ but turned out that's not the case anymore, that's why I reverted to it (without finding the new action to use).

crates/scalajs/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "wit-bindgen-scalajs"
+authors = ["Daniel Vigovszky <me@vigoo.dev>"]
+version = { workspace = true }
+edition = { workspace = true }
+repository = { workspace = true }
+license = { workspace = true }
+homepage = 'https://github.com/bytecodealliance/wit-bindgen'
+description = """
+Scala.JS bindings generator for WIT and the component model, typically used
```
sjrd, line 10:
```suggestion
Scala.js bindings generator for WIT and the component model, typically used
```
FYI, the recommended spellings are:

* Free text: Scala.js
* PascalCase: ScalaJS
* camelCase: scalaJS
* URL: scala-js

crates/scalajs/scala/wit.scala
```diff
@@ -0,0 +1,685 @@
+package object wit {
+
+  import scala.scalajs.js
+  import scala.scalajs.js.JSConverters._
+  import scala.scalajs.js.|
+  import scala.scalajs.js.annotation.JSName
+
+  sealed trait Nullable[+A] extends js.Any
```
sjrd, line 8: Given the Scala has non-nullable reference types by default, the consensus in Scala.js is to represent "nullable `A`" as `A`. So a `nullable String` is a `String`.

For primitives, we use their boxed variants : a `nullable Int` is a `java.lang.Integer`.

crates/scalajs/scala/wit.scala
```diff
@@ -0,0 +1,685 @@
...
+    def apply(): WitTuple0 = js.Array().asInstanceOf[WitTuple0]
+
+    def unapply(tuple: WitTuple0): Some[Unit] = Some(())
+
+    implicit def fromScalaTuple0(tuple: Unit): WitTuple0 = WitTuple0()
+
+    implicit def toScalaTuple0(tuple: WitTuple0): Unit = ()
+  }
+
+  sealed trait WitTuple1[T1] extends js.Object {
```
sjrd, line 105: Looks like `WitTupleN[...]` could be an alias of `js.TupleN[...]`.

sjrd:
Hi. Lead maintainer of Scala.js here.

I am honestly baffled that I never heard of this before just now. Yet it is presented as the "official" (`bytecodealliance/scalajs` namespace) way to use Scala.js with the component model. It's lucky that we even notice this PR at all.

The solution presented here (Scala.js->JS->Wasm) is better than nothing; that's for sure. It's a good workaround if you desperately want to use Scala in the component model today, rather than in 6-12 months. But I don't think this is *the* solution we want to present under the raw `scalajs` namespace.

As @tanishiking mentioned, direct Scala.js support for the Wasm component model is under construction (we already shipped a fully functional Scala.js->WasmGC-with-JS-host backend). But the source API won't look anything like this, since it won't actually use JS types. Once it's there, I would hope we could claim the `scalajs` namespace for the direct solution. We won't be able to do that in a backward compatible way if we already shipped the current workaround under that namespace.

I suggest to rename the crate so that it is clear that it goes through jco. Perhaps `scalajs-jco`? (no idea whether `-` is a thing Rust people use in their crates' names, sorry)

crates/scalajs/scala/wit.scala
vigoo: Yeah I've copied that from `js.Tuple2` but found there is only `Tuple2` and `Tuple3` in the scala.js library while you can definitely use bigger arities in WIT.

crates/scalajs/scala/wit.scala
vigoo: This is added to not loose the WIT type information of something being `optional<>`. Even though Componentize.js decided to represent optionals as nulls we don't have to expose it like that for Scala, in my opinion.

crates/scalajs/scala/wit.scala
sjrd: The other ones are here: https://github.com/scala-js/scala-js/blob/main/library/src/main/scala/scala/scalajs/js/Tuple.nodoc.scala so that they don't clutter the Scaladoc. But they do exist ;)

crates/scalajs/scala/wit.scala
sjrd: I disagree. The philosophy of Scala.js is to present JS as it is, without artifacts. Then, user-land libraries can build abstractions *on top* of the real thing.

Also, when targeting Scala 3, we can actually write `A | Null`, so that users using `-Yexplicit-nulls` can benefit from it all.

vigoo:
@sjrd I'm fine with not calling it `scalajs` bindgen but something more specific, or to just get rid of it once the real solution is done. For now this provides our users a way to use Scala right now - we have an sbt plugin as well that uses our forked bindgen at the moment, but I wanted to get create this PR upstream to contribute back and make it more accessible for people.

crates/scalajs/scala/wit.scala
vigoo: I understand this philosophy of Scala.js - but in this case I'm not trying to present JS at all - I'm trying to present a component model interface which _has_ optionals. Having JS in between is just a temporary "annoyance" that, as you wrote earlier, will go away.

So in my opinion the users of this would like to define a WIT interface containing optionals, and implement it in Scala. How it is represented in JS is not relevant for them.

About targeting Scala 3, I've added a command line option to set the dialect (2 vs 3) but not really taking advantage of it yet, it can be added easily.

crates/scalajs/scala/wit.scala
vigoo: Thanks, then these types are really not needed. Will remove them.

alexcrichton:
Thanks for the PR here @vigoo and for chiming in @tanishiking and @sjrd!

I can perhaps clarify from my perspective of maintaining wit-bindgen that I think it's reasonable to add this and facilitate it being more visible, but I also definitely would want to make sure that no one feels like their toes are being stepped on. I am no Scala (nor Scala.js) expert myself so I'd lean on y'all to guide how best this can be done.

The suggestion of naming this `scalajs-jco` or something along those lines sounds quite reasonable to me! It might also be worth documenting how this is a temporary solution with pointers to upstream work too?

sjrd:
Thanks for the rename. This looks good to me. (although I'm only judging the Scala.js aspects of it; I haven't looked at the Rust code)

vigoo:
Renamed it to `scalajs-jco` and applied some of the suggested changes.

Also note that for fully working with Componentize.js unfortunately there are two small tweaks in `spidermonkey-embedding-splicer` because of some specifics of what JS Scala.js generates. I believe these are small and generic enough to get included in componentize.js (and they are options so does not affect anything existing) and will open a PR for that there too - just our fork is a bit behind and I did not have time to update it yet.

alexcrichton:
Currently scalajs-jco is ~20 minutes in CI, would it be possible to optimize/cache things to get that more within 10 minutes? Currently C# is [around 30 minutes](https://github.com/bytecodealliance/wit-bindgen/issues/1158) but everything else is 10, and my hope is to stay roughly in the 10 minute time frame for cycle time

vigoo:
I'm not sure how to speed it up keeping the existing test structure - I believe the artifacts are cache and shared between the test runs, it's just starting up a jvm and sbt has a big overhead.

Maybe instead of generating a separate test case for each codegen test, we could generate one big scala project containing all, and compile it once? That would make it harder to quickly test things while working on the codebase, but would definitely speed up CI significantly.

alexcrichton:
I'm unfortunately not familiar enough with Scala.js or the infrastructure here to know how best to solve this. If rearchitecting things works well then that seems like the way to go yeah. I'm basically wary of adding more languages that slow down CI here as it's already quite a drag to wait for C# CI when it's significantly longer than all the others.

PR #1183, opened 2025-03-01
Wit bindgen rust 2024 compat

yoshuawuyts:
I believe this should fix the failing CI in https://github.com/bytecodealliance/cargo-component/pull/372, triggered by the new edition. I'm not entirely sure why the tests seem to be opting into a new edition, but I guess they do? Regardless: this changes the codegen of `wit-bindgen-rust` to respect the new "`unsafe fn` does not imply `unsafe {}`" rule added in the 2024 edition.

This PR is best reviewed commit by commit. The first commit is just the output of `cargo fix --edition`. The second commit are the actual `unsafe` changes.

Thanks!

### DIFF
--- a/crates/rust/src/bindgen.rs
+++ b/crates/rust/src/bindgen.rs
@@ -6,7 +6,7 @@ use wit_bindgen_core::abi::{Bindgen, Instruction, LiftLower, WasmType};
 use wit_bindgen_core::{dealias, uwrite, uwriteln, wit_parser::*, Source};
 
 pub(super) struct FunctionBindgen<'a, 'b> {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Would you be up for adding a test for this? For example leaving the wit-bindgen crate itself as edition 2021 but changing the generated output to be compatible with 2024. I'd ideally like to change `r#gen` to `generator` since raw idents aren't the actual fix to have here (just the quick fix) down the road, and I'd also like to double-check that the generated output is indeed compatible with 2024 since IIRC  we also need to do `unsafe extern "C" { .. }` I think?

yoshuawuyts:
Spoke with @pchickey today who volunteered to take over the implementation

PR #1185, opened 2025-03-03
Store arguments for async params/results on the stack

alexcrichton:
After #1176 there's no longer any need to store these values on the heap, so store them on the stack instead. This also updates to store params/results in an overlapping allocation which should be reasonable as it's local per-import and results are never written before parameters are read. (and params are never read after results are written).

### DIFF
--- a/crates/core/src/abi.rs
+++ b/crates/core/src/abi.rs
@@ -551,19 +551,8 @@ def_instruction! {
             blocks: usize,
         } : [1] => [0],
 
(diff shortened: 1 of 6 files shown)


dicej (approved):
LGTM, assuming this passes the `wasip3-prototyping` tests.

BTW, the phrase "on the stack" is arguably misleading when we may be talking about boxed, self-referential futures generated by the compiler from `async` functions or blocks.  Then "on the stack" may very well mean on the heap, but pedantry aside I understand what you mean.

PR #1187, opened 2025-03-03
rust: Store raw intrinsics in vtables

alexcrichton:
Instead of generating a wrapper-per-intrinsic try to instead store raw functions inside of vtables to push as much code as possible into `mod async_support`, as it's generally easier to grok code in one location rather than generated code.

Note that this does not update the `read` or `write` vtable functions as they're going to be trickier, if even possible, to expose the more raw version.

This shouldn't have any functional change, it's just a reorganizing. The `cancel_*` intrinsics are all marked with a `TODO`, however, to come back and reconsider the return value which is otherwise discarded right now.

### DIFF
--- a/crates/guest-rust/rt/src/async_support/future_support.rs
+++ b/crates/guest-rust/rt/src/async_support/future_support.rs
@@ -22,10 +22,29 @@ use {
 pub struct FutureVtable<T> {
     pub write: fn(future: u32, value: T) -> Pin<Box<dyn Future<Output = bool>>>,
     pub read: fn(future: u32) -> Pin<Box<dyn Future<Output = Option<Result<T, ErrorContext>>>>>,
-    pub cancel_write: fn(future: u32),
-    pub cancel_read: fn(future: u32),
-    pub close_writable: fn(future: u32, err_ctx: u32),
-    pub close_readable: fn(future: u32),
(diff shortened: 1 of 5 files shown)


crates/guest-rust/rt/src/async_support/future_support.rs
```diff
@@ -84,7 +103,11 @@ impl<T> CancelableWrite<T> {
                 Handle::LocalReady(..) => {
                     entry.insert(Handle::LocalOpen);
                 }
-                Handle::Write => (writer.vtable.cancel_write)(writer.handle),
+                Handle::Write => unsafe {
+                    // TODO: spec-wise this can return `BLOCKED` which seems
```
dicej, line 107: It can only return `BLOCKED` if called with the `async` option enabled.  As long as we don't try to import it using the `[async]` prefix, we should be good.

dicej (approved):

crates/guest-rust/rt/src/async_support/future_support.rs
alexcrichton: Ah ok, I'll leave this as a follow-up of https://github.com/bytecodealliance/wit-bindgen/issues/1188 to come back and fill in comments here.

Issue #1161, opened 2025-02-10
Refactor/rewrite this repository's test suite


alexcrichton:
I've personally long lamented tests in this repository. Currently tests fall into roughly three categories:

* `codegen` tests (Rust) - Rust codegen tests use some macro-magic to run a `generate!` macro invocation with various options over all the tests in `tests/codegen/*.wit`. This then compiles the output for the native host (not wasm) and then nothing actually happens at runtime as nothing is actually run, things are only generated.
* `codegen` tests (not Rust) - other languages typically have a `#[test]`-per-`generate!` also using some macro-magic. Each test typically spawns some external process and ensures it passes, for example a C compiler is spawned for the `wit-bindgen-c` crate.
* `runtime` tests - In `tests/runtime/*` there are a number of test that are organized as `#[test]`. The idea is that there's one monolithic host runner based on a custom Wasmtime embedding. Each runtime test can be implemented in an array of languages and each language gets run. Compilation happens in an external process and, like before, compilation of Rust binaries is "special" compared to other languages.

There are a lot of downsides to this, including:

* Rust is "special" and doesn't look like other languages in testing, that's not great.
* Compilation of `runtime` tests takes forever because you have to compile all of Wasmtime and its embedding. 
* Iteration on the `runtime` tests take forever because it takes awhile to compile.
* It's not easy to incrementally bring a language online as it generally needs to support everything or it's not tested at all.
* This repository is very close to a "reference test suite for WIT" but stops just shy of being something that's actually upstreamable anywhere.
* There's no way to benefit from these "runtime" tests in any other component runtime right now.

Overall I would like to move in a new direction that solves all of these problems with a new foundation for running tests. Specifically what I'm imagining is something like this:

1. All tests are organized into two components: a "runner" component and a "test" component. The `runner` imports the `test`.
2. The `runner` component exports the `wasi:cli/run` interface and looks like a command line executable. In other words it's a "command" in terms of WASI.
3. The `test` component looks like a "reactor" in terms of WASI.
4. Both `runner` and `test` can import WASI functions, for example to print to the screen, read environment variables, etc.
5. Tests are then created by composing the `runner` with the `test`, producing a final component that's a "command" and only imports WASI.
6. Tests are then executed with the `wasmtime` CLI, critically not the embedding API.

[...]

PR #1192, opened 2025-03-05
Add a `wit-bindgen test` subcommand

alexcrichton:
As I've read more and more of the Rust async runtime support and other various bits and pieces I've wanted more and more the ability to easily write tests for guest interactions with the host. While I don't think it's feasible to generate arbitrary hosts I do think it's possible to do this much more easily than is done today with the testing support in this repository. In essence this commit is an implementation of #1161.

The goal of this commit is to add a `wit-bindgen test` test suite runner. This test suite will be used to migrate all existing tests in this repository to this new framework. In the limit this is expected to make it easier to write tests (no Rust knowledge necessary), make it more flexible to write tests (now you can use raw `*.wat`), and additionally improve the quality of the test suite by making it more reusable. The reusability isn't the highest priority at this time as it's not clear what else would want to reuse this, but my hope is that this refactoring is at least a large-ish leap forward towards having a component model test suite of some kind eventually.


alexcrichton:
For now I'd definitely considere this in a "hopefully working" but still somewhat draft form. I'm curious to get others' feedback on this design and possible improvements (if anyone has any).

Documentation is available [here](https://github.com/alexcrichton/witx-bindgen/blob/wit-bindgen-tests/crates/test/README.md) in rendered form. 

The remaining work items after this PR are:

* Finish migrating other languages supported in this repository to this new test subcommand: C# and teavm-java.
* Finish migrating Rust codegen tests to this repository
* Migrate all preexisting `tests/runtime/*` tests to `tests/runtime-new/*` (this new subcommand)
* Jettison `crates/test-helpers` and all other preexisting test support.

alexcrichton:
@pchickey would you be willing to take on review of this?

If you're ok with it you can ignore all changes in `test/*` and take my word for it that tests are migrated over. That leaves mostly just `crates/test/*` and some misc changes here and there which should in theory be much smaller than the 7kloc size as posted on the PR itself

pchickey (approved):
This is a huge PR so as described in the comments I really only reviewed the new test crate and spot-checked a translation.

I wasn't immediately bought into the `test` / `runner` nomenclature, but its fine. I can't think of better terms besides (maybe `inner`/`outer` which is also imperfect), and I don't think its really worth changing or painting the shed in general.

alexcrichton:
Thanks for reviewing! I'll note that if in the future we stumble on better names I'm happy to take on the work of renaming things.

PR #1197, opened 2025-03-06
feat(rust): do not yield to the host while guest tasks are ready

rvolosatovs:
cc @dicej 

This fixes https://github.com/bytecodealliance/wasip3-prototyping/issues/66

Note, that eventually we might want go get smarter about this (or not) e.g. see https://docs.rs/tokio/latest/tokio/task/index.html#cooperative-scheduling

### DIFF
--- a/crates/guest-rust/rt/src/async_support.rs
+++ b/crates/guest-rust/rt/src/async_support.rs
@@ -2,6 +2,7 @@
 #![allow(static_mut_refs)]
 
 extern crate std;
+use core::sync::atomic::{AtomicBool, Ordering};
 use std::alloc::{self, Layout};
 use std::any::Any;
 use std::boxed::Box;
@@ -78,26 +79,29 @@ pub fn with_entry<T>(handle: u32, fun: impl FnOnce(hash_map::Entry<'_, u32, Hand
     fun(unsafe { HANDLES.entry(handle) })
 }
 
-fn dummy_waker() -> Waker {
-    struct DummyWaker;
-
-    impl Wake for DummyWaker {
-        fn wake(self: Arc<Self>) {}
-    }
-
-    static WAKER: Lazy<Arc<DummyWaker>> = Lazy::new(|| Arc::new(DummyWaker));
-
-    WAKER.clone().into()
-}
-
 /// Poll the specified task until it either completes or can't make immediate
 /// progress.
 unsafe fn poll(state: *mut FutureState) -> Poll<()> {
+    #[derive(Default)]
+    struct FutureWaker(AtomicBool);
+
+    impl Wake for FutureWaker {
+        fn wake(self: Arc<Self>) {
+            Self::wake_by_ref(&self)
+        }
+
+        fn wake_by_ref(self: &Arc<Self>) {
+            self.0.store(true, Ordering::Relaxed)
+        }
+    }
+
     loop {
         if let Some(futures) = (*state).tasks.as_mut() {
             let old = CURRENT;
             CURRENT = state;
-            let poll = futures.poll_next_unpin(&mut Context::from_waker(&dummy_waker()));
+            let waker: Arc<FutureWaker> = Arc::default();
+            let poll =
+                futures.poll_next_unpin(&mut Context::from_waker(&Arc::clone(&waker).into()));
             CURRENT = old;
 
             if SPAWNED.is_empty() {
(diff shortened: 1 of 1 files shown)


crates/guest-rust/rt/src/async_support.rs
```diff
@@ -107,7 +111,11 @@ unsafe fn poll(state: *mut FutureState) -> Poll<()> {
                         (*state).tasks = None;
                         break Poll::Ready(());
                     }
-                    Poll::Pending => break Poll::Pending,
+                    Poll::Pending => {
+                        if !waker.0.load(Ordering::Relaxed) {
+                            break Poll::Pending;
+                        }
```
alexcrichton, line 117: I forget how all the bits line up here, but this might be a case where it'd be good to yield to the host? E.g. the async export would return a code indicating that it's yielding. (although I know `yield` is also an imported function, so now I'm not sure if there's a return from the top-level function that means yielding is happening... if that's not the case then this looks reasonable to me)

dicej: Yeah, we could definitely yield here if desired, although `wasip3-prototyping` doesn't support guests returning `STATUS_YIELDED` yet due to it being a relatively newer addition to the spec I haven't gotten around to yet.  Shouldn't be hard to support, though.

alexcrichton: Makes sense, mind leaving a comment indicating that we probably want to transition to that?

crates/guest-rust/rt/src/async_support.rs
```diff
@@ -107,7 +111,11 @@ unsafe fn poll(state: *mut FutureState) -> Poll<()> {
                         (*state).tasks = None;
                         break Poll::Ready(());
                     }
-                    Poll::Pending => break Poll::Pending,
+                    Poll::Pending => {
+                        if !waker.0.load(Ordering::Relaxed) {
```
dicej, line 115:
```suggestion
                        // TODO: Return `CallbackCode.YIELD` (see https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canon-lift)
                        // to the host before polling again once a host implementation exists to support it.
                        if !waker.0.load(Ordering::Relaxed) {
```

dicej (approved):
LGTM.  @rvolosatovs have you verified this fixes the issue?  Also, have you had a chance to run the `wasip3-prototyping` tests (e.g. `component-async-tests` and the `p3` tests) with this patch?  If so, we should be good to merge.

rvolosatovs:
> LGTM. @rvolosatovs have you verified this fixes the issue? Also, have you had a chance to run the `wasip3-prototyping` tests (e.g. `component-async-tests` and the `p3` tests) with this patch? If so, we should be good to merge.

I've just pushed a draft PR using this rev in https://github.com/bytecodealliance/wasip3-prototyping/pull/69, let's merge once CI passes on that PR

rvolosatovs:
> > LGTM. @rvolosatovs have you verified this fixes the issue? Also, have you had a chance to run the `wasip3-prototyping` tests (e.g. `component-async-tests` and the `p3` tests) with this patch? If so, we should be good to merge.
> 
> I did locally, but I've just pushed a draft PR using this rev in [bytecodealliance/wasip3-prototyping#69](https://github.com/bytecodealliance/wasip3-prototyping/pull/69), let's merge once CI passes on that PR

well, the CI hit the `new` bug https://github.com/bytecodealliance/wasip3-prototyping/actions/runs/13710666799/job/38346330477?pr=69

I've been using https://github.com/rvolosatovs/wit-bindgen/commit/554b4820ada1cd890981622753f4302266d3da0c for testing, which was based on the last-known-good commit (used in `main` of wasip3-prototyping previously)

e.g. https://github.com/rvolosatovs/wasi/blob/e3e9143d74d8f962c2e5cad49d19af75d3b4b542/Cargo.toml#L27-L28

dicej:
If you tested it locally, that's good enough for me.  If you really want to test `wasip3-prototyping`, though, you could update the draft PR to point to the branch based on the known good commit.

rvolosatovs:
> If you tested it locally, that's good enough for me.  If you really want to test `wasip3-prototyping`, though, you could update the draft PR to point to the branch based on the known good commit.

Tested locally on a Mac and in CI using the rebased version

PR #1198, opened 2025-03-07
Update support for p3 to match the latest wasm-tools

alexcrichton:
Work-in-progress currently as https://github.com/bytecodealliance/wasip3-prototyping/pull/68 is the upstream work in Wasmtime being used to  discover test failures and such.


alexcrichton:
@dicej your commit here looks good to me, although I can't self-approve my own PR

I also threw in a merge with `main` to get the update of wasm-tools out of the diff as well

PR #1199, opened 2025-03-10
feat: Allow variants and records to be ignored by additional_derives

cdmurph32:
This feature allows some variants and records to use types for which adding traits will cause compilation to fail, such as serde::Deserialize on wasi:io/streams.

Variants and records are specified as they are listed in the wit file, i.e. in kebab case.


pchickey (approved):
Thanks!

PR #1200, opened 2025-03-10
[c#] Enable running the Csharp tests on linux

jsturtevant:
Linux is supported but wasn't actually running in CI: https://github.com/bytecodealliance/wit-bindgen/issues/1161#issuecomment-2711475522


pavelsavara:
It seems that macos is running and even passing some tests
https://github.com/bytecodealliance/wit-bindgen/actions/runs/13775463381/job/38523649143?pr=1200

What are the conditions in this PR excluding ?

jsturtevant:
Yup, I was following the same pattern before but it is misleading.  That's a better approach to explicitly skip or not run them at all.

alexcrichton (approved):

PR #1201, opened 2025-03-11
Use `--release` on tests in CI

alexcrichton:
I have a hunch this is why C# is so slow (the modules are big and there's a lot of them and unoptimized Cranelift is very slow).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -117,7 +117,8 @@ jobs:
           -p wit-bindgen-cli \
           -p wit-bindgen-${{ matrix.lang }} \
           --no-default-features \
-          --features ${{ matrix.lang }}
+          --features ${{ matrix.lang }} \
+          --release
 
   test_unit:
     name: Crate Unit Tests


jsturtevant:
The windows runtime tests also build in debug mode, with might slow it down too? https://github.com/bytecodealliance/wit-bindgen/blob/d73c073b0f277db55e4994397931ee00daf47e20/tests/runtime/main.rs#L744

alexcrichton:
Looks like C# on windows was [32m53s](https://github.com/bytecodealliance/wit-bindgen/actions/runs/13772559788/job/38514349093) before this change and [25m37s](https://github.com/bytecodealliance/wit-bindgen/actions/runs/13792497401/job/38575775483?pr=1201) after this change. Good wins, but just building the rust test harness took 8m35s in debug mode and 16m03s in release mode, so looks like this is +8m to rust build time but -15m to C# testing time. Overall seems like a good win!

With https://github.com/bytecodealliance/wit-bindgen/pull/1192 this should only get better since we basically won't have to build Wasmtime in CI any more.

alexcrichton:
> The windows runtime tests also build in debug mode, with might slow it down too? 

Oh also wanted to mention: good point! I'll try evaluating this in a future PR

alexcrichton:
Oh sorry, yes,  given that this shaves off a nontrivial chunk of time with C# I think it'd be good to land (updated title to be more confident)

pchickey (approved):

PR #1207, opened 2025-03-12
Support C# in wit-bindgen test 

alexcrichton:
Migrate various bits of logic to the new test runner. All codegen tests
are migrated and a single "demo" program is available to show the
runtime test at least somewhat works. Migrating more tests is left to
future work.


alexcrichton:
@jsturtevant question for you as well: I'm attempting to migrate the `numbers/wasm.cs` test to the new framework. Can C# produce a "CLI executable" in that it runs the `Main` function defined in C#? If so does the `*.csproj` need to be updated to reflect that? Naively doing this:

```csharp
public class Program
{
    public static void Main(string[] args)
    {
        // ..
    }
}
```

I got no exports from the final component so I assume the compiler needs to be configured to produce a `wasi:cli/run` export and/or run the main function.

silesmo:
@alexcrichton The Output type needs to be changed from Library to Exe in the property group in the .csproj file.

alexcrichton:
Nice, thanks! Pushed up a commit and the "demo" test now has both a runner and a test half as C#.

After this it should now be possible to fully migrate `tests/runtime/*` to the new framework.

jsturtevant (approved):

PR #1209, opened 2025-03-13
fix(moonbit): set correct string length

peter-jerry-ye:
This PR fixes issues when handling unicode string where the length of the string was broken during reallocation.

The bug could be reproduced by running `wasmtime run cli.wasm "中" "🀄️" "⏰"` where cli.wasm would a MoonBit program printing arguments retrieved from environment.

The new implementation has also been tested when composing with another UTF8 component.

### DIFF
--- a/crates/moonbit/src/lib.rs
+++ b/crates/moonbit/src/lib.rs
@@ -95,8 +95,8 @@ pub extern "wasm" fn free(position : Int) =
   #|(func (param i32) local.get 0 i32.const 8 i32.sub call $moonbit.decref)
 
 pub fn copy(dest : Int, src : Int) -> Unit {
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks!

Also if you're curious you might be interested to explore https://github.com/bytecodealliance/wit-bindgen/pull/1192 and add moonbit support to `wit-bindgen test` to help test this eventually

peter-jerry-ye:
Sure, I will try to add tests later. By the way, I may have missed something, but is it possible that we cross-test, i.e. runner from Rust + test from MoonBit?

alexcrichton:
It is indeed!

Issue #1170, opened 2025-02-21
The wit-bindgen-rt crate does not ship with a README.md or a LICENSE file


ephemeralriggs:
The [current version (0.39.0) as of this writing](https://crates.io/crates/wit-bindgen-rt/0.39.0) of wit-bindgen-rt on crates.io does not come with a README.md or LICENSE file.

While the former is more of an inconvenience, the missing of a LICENSE file is an issue for projects that auto-check for licence compatibility during build. It would be helpful of those files could be added to the crate.

alexcrichton:
A README was added in https://github.com/bytecodealliance/wit-bindgen/pull/1162, and for LICENSE can you use the `license` field of `Cargo.toml`?

ephemeralriggs:
It would be the first time that we are doing this for our projects, since every other crate we use ships a LICENSE file. So naturally I would prefer to avoid special handling just for one single crate. May I ask what is special about wit-bindgen-rt that the LICENSE files are included in the source code repository but are not included in the release on crates.io?

alexcrichton:
The BA has a number of repos, such as this, https://github.com/bytecodealliance/wasm-tools, and https://github.com/bytecodealliance/wasmtime, which all are a workspace of many crates. Instead of copying files all over the workspace the license files are only present at the root and all crates are annotated with that license. I don't know how to manage a workspace of many crates without having many copies of a license.

erickt:
Hi Alex! Long time no see.

The two ways to take care of this that we know of is to either use `license-file` in the workspace and crate packages, or to use symlinks, such as https://github.com/serde-rs/serde/tree/master/serde_derive. My personal preference is symlinks since it still allows for SPDX expressions in the `Cargo.toml`. If that'd be appropriate, we could put together a PR for it if you'd like.

alexcrichton:
Hey! Too long time with no see!

Agreed I'd want to stick to `license` given that we also have three license files instead of one to stick in `license-file`. Do you know how the symlink bits work on Windows? I always assumed that solution relied on developers basically not working on Windows or otherwise having turned on that setting of "allow symlinks"

erickt:
What a wonderful question. I checked on a windows machine, and it does do the wrong thing in this case, the `LICENSE-APACHE` file just contains `../LICENSE-APACHE`, rather than the copied contents. However `cargo package` does warn that this appears to be the case:

```
% cargo package
[...]

PR #1213, opened 2025-03-14
fix: Add licenses to published crates

ephemeralriggs:
Symlink the licenses from the root directory to the crates which are publishes, so they include the license files.

Bug: #1170


alexcrichton (approved):
Thanks!

PR #1215, opened 2025-03-15
C# Fixes wasi http header bug and adds a test for it

silesmo:
Fixes a bug where headers returned from a C# component wouldn't be properly set. Adds a test for it and some cleanup.

### DIFF
--- a/crates/csharp/src/function.rs
+++ b/crates/csharp/src/function.rs
@@ -66,7 +66,7 @@ impl<'a, 'b> FunctionBindgen<'a, 'b> {
             resource_drops: Vec::new(),
             is_block: false,
             fixed_statments: Vec::new(),
-            parameter_type: parameter_type,
+            parameter_type,
         }
     }
 
@@ -458,14 +458,14 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 if flags.flags.len() > 32 {
                     results.push(format!(
                         "unchecked((int)(((long){}) & uint.MaxValue))",
-                        operands[0].to_string()
+                        operands[0]
                     ));
                     results.push(format!(
                         "unchecked(((int)((long){} >> 32)))",
-                        operands[0].to_string()
+                        operands[0]
                     ));
                 } else {
-                    results.push(format!("(int){}", operands[0].to_string()));
+                    results.push(format!("(int){}", operands[0]));
                 }
             }
 
@@ -479,8 +479,8 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                     results.push(format!(
                         "({})(unchecked((uint)({})) | (ulong)(unchecked((uint)({}))) << 32)",
                         qualified_type_name,
-                        operands[0].to_string(),
-                        operands[1].to_string()
+                        operands[0],
+                        operands[1]
                     ));
                 } else {
                     results.push(format!("({})({})", qualified_type_name, operands[0]))
@@ -502,7 +502,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 let mut result = format!("new {} (\n", qualified_type_name);
 
                 result.push_str(&operands.join(", "));
-                result.push_str(")");
+                result.push(')');
 
                 results.push(result);
             }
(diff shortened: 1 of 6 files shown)


crates/csharp/src/function.rs
```diff
@@ -754,7 +754,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                                 "
                             );
                         }
-                        results.push(format!("(nint){ptr}"));
+                        results.push(format!("(int){ptr}"));
```
jsturtevant, line 757: why this change?

jsturtevant (commented):
Is the fix here around the alignment? How was the issue presenting?

crates/csharp/src/function.rs
```diff
@@ -889,13 +889,13 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                             "
                             void* {address};
                             if (({size} * {list}.Count) < 1024) {{
-                                var {ret_area} = stackalloc {element_type}[({array_size}*{list}.Count)+1];
-                                {address} = (void*)(((int){ret_area}) + ({align} - 1) & -{align});
```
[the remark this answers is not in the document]
pavelsavara: Refactor `void* AlignStackPtr(void* stackAddress)` helper and use it everywhere.

crates/csharp/src/function.rs
pavelsavara: Because it generates 
```cs
BitConverter.TryWriteBytes(new Span<byte>((void*)(basePtr + 8), 4), (nint)listPtr);
```

And that uses `TryWriteBytes(Span<Byte>, UInt64) ` which is wrong. There is no TryWriteBytes overload for `nint`.
And then it returns `false` but we ignore it.
Let's stop using TryWriteBytes. 

Instead it could be just 
```cs
new Span<nint>((void*)(basePtr + 8), 1)[0] = listPtr;
```

ignatz:
I ran into this issue as well. Are there any plans to land this fix or similar? Thanks :pray:

crates/csharp/src/function.rs
silesmo: Fixed @pavelsavara

crates/csharp/src/function.rs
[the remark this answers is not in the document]
silesmo: Fixed @pavelsavara

tests/runtime/lists/test.wit
```diff
@@ -26,6 +26,8 @@ interface test {
   list-roundtrip: func(a: list<u8>) -> list<u8>;
 
   string-roundtrip: func(a: string) -> string;
+
+  wasi-http-headers-roundtrip: func(a: list<tuple<string, list<u8>>>) -> list<tuple<string, list<u8>>>;
```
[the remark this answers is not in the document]
silesmo: Fixed @jsturtevant

crates/csharp/src/function.rs
silesmo: This is what was causing the issue @jsturtevant

silesmo:
> Had a brief look through and looks like a good improvement! Thanks!

Do you have the time to leave a review so I can merge it in?

yowl:
As I wasn't one of the original reviewers I was going to leave it for them, but I will review it with previous comments in mind,

crates/csharp/src/function.rs
yowl: The `(int)` cast still seems to be there, is that intentional?

crates/csharp/src/world_generator.rs
```diff
@@ -623,13 +624,28 @@ impl WorldGenerator for CSharp {
...
 
+        if self.needs_align_stack_ptr {
+            uwrite!(
+                src,
+                "
+                {access} static class MemoryHelper
+                {{
+                    {access} static unsafe void* AlignStackPtr(void* stackAddress, uint alignment)
+                    {{
+                        return (void*)(((int)stackAddress) + ((int)alignment - 1) & -(int)alignment);
```
yowl: Would `(nint)` be better when we get to wasm64?

yowl:
> Looks good, thanks!

Just the small question on the use of `int` where `nint` might future proof us better for wasm64

crates/csharp/src/world_generator.rs
silesmo: Yea you are right. I missed changing that back

crates/csharp/src/function.rs
silesmo: No, but fixed now

PR #1218, opened 2025-03-17
Align on naming "c-sharp" vs "csharp"

alexcrichton:
This fixes an issue where CI isn't currently running any tests for C# because CI is using "csharp" but the `wit-bindgen test` requires "c-sharp" because that's what `wit-bindgen`-the-CLI requires. This uses "csharp" in more locations and updates the CLI to allow both.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,11 +59,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         # moonbit removed from language matrix for now - causing CI failures
-        lang: [c, rust, c-sharp]
+        lang: [c, rust, csharp]
         exclude:
-          # For now c-sharp doesn't work on macos, so exclude it from testing.
+          # For now csharp doesn't work on macos, so exclude it from testing.
           - os: macos-latest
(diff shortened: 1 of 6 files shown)


jsturtevant:
I guess the crate folder will need to be renamed too or some kind of change to the value coming in at https://github.com/bytecodealliance/wit-bindgen/blob/6541e44c8d10104fa112d7f37fe10bfaf0b9a40a/.github/workflows/main.yml#L107

alexcrichton:
Hm ok I'm going to try to use "csharp" instead for this repository and the CLI will also recognize "c-sharp" as an alias. If it's officially best known as "c-sharp" when "C#" can't be used otherwise I'm happy to change though.

jsturtevant:
> CLI will also recognize "c-sharp" as an alias

 this sounds like a reasonable approach with the least amount of churn.  FWIW I found some markdown in the docs repo that used `csharp` as the [language tag](https://github.com/search?q=repo%3Adotnet%2Fruntime+csharp+language%3AMarkdown&type=code&l=Markdown).

jsturtevant (approved):

PR #1236, opened 2025-03-21
Migrate the `rust_xcrate` test to `wit-bindgen test`

alexcrichton:
This was a tricky test to migrate. Doing this required implementing a number of new features:

* Rust now has a `[lang]`-specific configuration `externs = [...]` which is a list of files to compile as external crates. This tests the cross-crate behavior of the test.
* Tests now support more than just runner/test composed components, now they also support multiple components being composed. This is required because the test was specifically testing behavior where a component simultaneously imports and exports.

This last point was particularly tricky. Namely this required abstracting the definition of a test to a runner plus a list of tests, not just one test. These are still modeled as a list of `world` items in the original WIT document. Additionally I was having trouble getting `wasm-compose` working so I ended up switching to `wac`. This means that tests can now support custom `wac` composition scripts to configure how exactly the composition is created, and that's used in this new test to insert a component as a sandwich between two others.


crates/test/src/rust.rs
```diff
@@ -39,8 +39,9 @@ struct RustConfig {
     /// Space-separated list or array of compiler flags to pass.
     #[serde(default)]
     rustflags: StringList,
-    // TODO: something about auxiliary builds of crates to test cross-crate
-    // behavior.
+    /// TODO
```
jsturtevant, line 42: nit,
```suggestion

```

crates/test/src/lib.rs
```diff
@@ -735,26 +818,84 @@ impl Runner<'_> {
...
+        let document = match &case.config.wac {
+            Some(path) => {
+                let wac_config = case.path.join(path);
+                fs::read_to_string(&wac_config)
+                    .with_context(|| format!("failed to read {wac_config:?}"))?
+            }
+            // Default wac script is to just make the `test:test` component
+            // available to the `runner`.
+            None => "
+                package example:composition;
```
jsturtevant, line 839: will this default wac script work if you rename it with 

```//@ runner = "other-runner"``` 

as described in the docs above `Test Configuration: World Names`?

File: tests/runtime-new/rust/xcrate/test.wit [hunk 3]
```diff
@@ -0,0 +1,75 @@
+//@ tests = ['intermediate', 'leaf']
```
jsturtevant, line 1: does this mean `runner` is run with both `intermediate` and `leaf`?

crates/test/src/lib.rs
alexcrichton: Good point, I've updated this to account for custom names now too

tests/runtime-new/rust/xcrate/test.wit
alexcrichton: Good question, and mostly points to this being a bad name. I've renamed this to `dependencies` which should make it more clear that these are dependencies of the `runner` component, not separate and individual tests. The sum total of all three components are what makes a single test.

Issue #1232, opened 2025-03-17
c#: Resource and method with same name


MendyBerger:
Componentize generates invalid code in cases where a resource and it's method share the same name.

E.G.
```wit
resource foo {
  foo: func();
}
```
Results in the following error:
```
Restore complete (0.5s)
  dotnet_counter failed with 1 error(s) (2.1s)
    [path]/obj/Release/net9.0/wasi-wasm/wit_bindgen/BrowserWorld.wit.imports.webidl.browser.IGlobal.cs(205513,30): error CS0542: 'Foo': member names cannot be the same as their enclosing type

Build failed with 1 error(s) in 2.9s
```

jsturtevant:
the root cause of this is https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0542?f1url=%3FappId%3Droslyn%26k%3Dk(CS0542) which means you cannot have functions of the same name as the class unless it is a constructor.

I am not sure the best way to handle this as it is a restriction on c#.

jsturtevant:
There has been some discussion on what the best fix is for this in https://github.com/bytecodealliance/wit-bindgen/pull/1238

jsturtevant:
The Component Model spec has been updated to address this: https://github.com/WebAssembly/component-model/pull/494.  There are some changes coming that would catch this https://github.com/bytecodealliance/wasm-tools/pull/2138

PR #1238, opened 2025-03-21
c#: Fix same name issue in resources

jsturtevant:
possible fix for #1232 

I am not a huge fan of this but not sure if there is another way

@yowl @silesmo @pavelsavara thoughts?

### DIFF
--- a/crates/csharp/src/function.rs
+++ b/crates/csharp/src/function.rs
@@ -1434,9 +1434,9 @@ struct BlockStorage {
 #[derive(Clone)]
 pub struct ResourceInfo {
     pub(crate) module: String,
-    pub(crate) name: String,
     pub(crate) docs: Docs,
     pub(crate) direction: Direction,
+    pub(crate) wit_name: String,
 }
 
 impl ResourceInfo {
@@ -1451,7 +1451,11 @@ impl ResourceInfo {
                 .strip_prefix("I")
                 .unwrap()
                 .to_upper_camel_case(),
-            self.name.to_upper_camel_case()
+            self.name().to_upper_camel_case()
         )
+    }
+
+    pub(crate) fn name(&self) -> String {
+        format!("{}Resource", self.wit_name)
     }
 }
--- a/crates/csharp/src/interface.rs
+++ b/crates/csharp/src/interface.rs
@@ -744,7 +744,7 @@ impl InterfaceGenerator<'_> {
         let qualified = self.type_name_with_qualifier(&Type::Id(id), true);
         let info = &self.csharp_gen.all_resources[&id];
         let name = info.name.clone();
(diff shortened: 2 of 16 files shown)


pavelsavara:
Perhaps there could be opt-in mapping per type. `--map-resource-name=XYZ:XYZResource` or something like that. 
Or even for that specific method. `--map-function-name=foo.foo:foo.fooFactory`

pavelsavara:
Making C# look differently than wit is not great. Let's keep the problem as localized as possible.

jsturtevant:
>Perhaps there could be opt-in mapping per type. --map-resource-name=XYZ:XYZResource or something like that.
Or even for that specific method. --map-function-name=foo.foo:foo.fooFactory
> Making C# look differently than wit is not great. Let's keep the problem as localized as possible.

💯. thanks for the suggestion, I'll see if I can get it working

pavelsavara:
If we want something that magically always works, we could have `foo._foo` or `foo.foo1` when we detect class and method name match.

silesmo:
I understand the reasoning for allowing to re-map the name of a type. But that will also add edge cases where you would need to manually remap something if they clash? I would expect the tooling to handle this automatically to not introduce additional hurdles for new users to adopt C# with wasi. @pavelsavara @jsturtevant

pavelsavara:
I guess other languages also have naming clashes, for example with keywords or with method names that start with numbers. having same rules would help.

That said naming something `foo.foo` is bad idea and that's why C# has the rule in place. Maybe we could also propagate that rule to wit spec ?

silesmo:
Keywords are generally handled by escaping the keyword. So it wouldn't directly map to this problem.

Yea I wouldn't mind trying to push it to the wit spec if we can get others to agree.

jsturtevant:
dropping a few more notes here:

- c generator does have a some renaming but not exactly the same as we are discussing here
  - https://github.com/bytecodealliance/wit-bindgen/blob/5e7d46c5d2e46866ae96e9592c80f49144f01aee/crates/c/src/lib.rs#L87-L92
- wit has some limited renaming of via use but also not of use here
  - https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#top-level-use

jsturtevant:
something like `foo._foo` might be relatively unobtrusive.  

I've also raised the issue in the component model https://github.com/WebAssembly/component-model/issues/478

jsturtevant:
proposed change to the spec: https://github.com/WebAssembly/component-model/pull/494

jsturtevant:
this is now part of the spec: https://github.com/WebAssembly/component-model/pull/494 with a change going into the tooling: https://github.com/bytecodealliance/wasm-tools/pull/2138

PR #1241, opened 2025-03-24
Refactor some ABI bits in `wit-bindgen-core`

alexcrichton:
Pass some variables around instead of storing them on `Generator`, removing the need to synthesize dummy values in a few contexts.


alexcrichton:
I'll note that I was setting this up for a bigger change but I think I'm going to abandon that change for now instead. Regardless this should be reasonable to land anyway.

PR #1243, opened 2025-03-25
refactor(moonbit): use internal abi

peter-jerry-ye:
This PR made a lot of hand written inline wasm to call the builtin functions, so that:
1. less bugs
2. prepare for migrating the memory layout.

This PR also adapts the new test framework for MoonBit.


peter-jerry-ye:
@alexcrichton The new test framework was great. It helped me find a number of bugs.

I wonder though if we can exclude MoonBit from the default language testing set for now? (same reason as why it's not in the CI, it's not that stable yet) I tried not to add MoonBit to `ALL` but then the codegen test won't run for MoonBit even if I set `--languages moonbit`.

alexcrichton (approved):
Thanks for this! 

I'll work on integrating a better set of defaults into `wit-bindgen test` to get tests working for this with explicit opt-in, but for now seems reasonable to have this as-is

alexcrichton:
Ok in https://github.com/bytecodealliance/wit-bindgen/pull/1246 I'm going to just require that `--language` is always passed to request something explicitly and avoid the "include everything" behavior.

PR #1244, opened 2025-03-25
Implement serde::Deserialize for wit-bindgen-rust::Opts

justingaffney:
To allow deserialisation of bindgen options, for example from Cargo package metadata like in https://github.com/bytecodealliance/cargo-component/pull/384


alexcrichton (commented):
Thanks! Mind making this an optional dependency to avoid adding extra build time to existing deps?

alexcrichton (approved):

PR #1246, opened 2025-03-26
Don't include all languages by default in `wit-bindgen test`

alexcrichton:
Instead switch the `-l`/`--language` argument to a required argument. I'm backtracking on "include everything" being a reasonable default since I don't think most users actually want to test everything. Short of that this requires explicit opt-in for what to test.


pchickey (approved):
Agreed, all languages is not the desirable default.

PR #1249, opened 2025-03-28
Refactor futures/streams support for Rust

### LINKED ISSUE #1152: Rust async support: WIT stream<T> -> Rust stream<Vec<T>>
I am studying the new async support and trying to figure out the rationale behind the WIT stream\<T\> translating into Rust stream\<Vec\<T\>\> instead of stream\<T\>.  

A common use of async streams in WIT is for sockets and http, where a stream of bytes (chunks) is needed.  In this case stream\<u8\> covers the need, but this could also be achieved by defining a stream\<list\<T\>\> in WIT which would translate to stream\<Vec\<u8\>\> in Rust.  On the other hand, a stream of complex records which should be processed individually cannot be expressed with a WIT stream, as this would result in a stream of complex records chunks.  

Wouldn't it be more universal to define as WIT stream\<T\> as a Rust stream\<T\>, to cover all use cases?

### LINKED ISSUE #1174: `<FutureReader<T> as IntoFuture>::Output` should be `Result<T, ErrorContext>`
Currently, `<FutureReader<T> as IntoFuture>::Output` is `Option<Result<T, ErrorContext>>`, but that's due to my earlier, incorrect understanding of the spec.  The spec says the owner of the write end of the future must either write a value before closing it or close it with an error, so the only two possible outcomes for the read end are a value or an error.  And note that if the writer writes a value _and_ then closes it with an error, the error will be silently dropped, just like it would be if the reader had already explicitly closed its end.

### LINKED ISSUE #1177: Cancelled future reads/writes leak state
Currently if you cancel a future read or write (and I assume stream reads or writes) [this is the suspension point](https://github.com/bytecodealliance/wit-bindgen/blob/e0d2b889411acf60738dcd976ed5373f7a81e651/crates/guest-rust/rt/src/async_support.rs#L258) where the future goes away. The work done in that function to update `todo` and update `CALLS` are both leaked and probably need to get cleaned up.

### LINKED ISSUE #1178: Futures/streams with 2+ levels of indirection leak memory
This WIT today:

```
package test:foo;

world the-world {
  import x: func(x: future<list<list<u32>>>);
}
```

generates this `write` function for the vtable:

```rust
        fn write(
            future: u32,
            value: super::super::_rt::Vec<super::super::_rt::Vec<u32>>,
        ) -> ::core::pin::Pin<
            super::super::_rt::Box<dyn ::core::future::Future<Output = bool>>,
        > {
            super::super::_rt::Box::pin(async move {
                #[repr(align(4))]
                struct Buffer([::core::mem::MaybeUninit<u8>; 8]);
                let mut buffer = Buffer([::core::mem::MaybeUninit::uninit(); 8]);
                let address = buffer.0.as_mut_ptr() as *mut u8;
                unsafe {
                    let vec1 = &value;
                    let len1 = vec1.len();
                    let layout1 = super::super::_rt::alloc::Layout::from_size_align_unchecked(
                        vec1.len() * 8,
                        4,
                    );
                    let result1 = if layout1.size() != 0 {
                        let ptr = super::super::_rt::alloc::alloc(layout1).cast::<u8>();
                        if ptr.is_null() {
                            super::super::_rt::alloc::handle_alloc_error(layout1);
                        }
                        ptr
                    } else {
                        ::core::ptr::null_mut()
                    };
                    for (i, e) in vec1.into_iter().enumerate() {
                        let base = result1.add(i * 8);
                        {
                            let vec0 = e;
                            let ptr0 = vec0.as_ptr().cast::<u8>();
                            let len0 = vec0.len();
                            *base.add(4).cast::<usize>() = len0;
                            *base.add(0).cast::<*mut u8>() = ptr0.cast_mut();
                        }
                    }
                    *address.add(4).cast::<usize>() = len1;
                    *address.add(0).cast::<*mut u8>() = result1;
                }
                #[cfg(not(target_arch = "wasm32"))]
                unsafe extern "C" fn wit_import(_: u32, _: *mut u8) -> u32 {
                    unreachable!()
                }
                #[cfg(target_arch = "wasm32")]
                #[link(wasm_import_module = "[import-payload]$root")]
                extern "C" {
                    #[link_name = "[async][future-write-0]x"]
                    fn wit_import(_: u32, _: *mut u8) -> u32;
                }
                match unsafe {
                    wit_bindgen::rt::async_support::await_future_result(
                            wit_import,
                            future,
                            address,
                        )
                        .await
                } {
                    wit_bindgen::rt::async_support::AsyncWaitResult::Values(_) => true,
                    wit_bindgen::rt::async_support::AsyncWaitResult::End => false,
                    wit_bindgen::rt::async_support::AsyncWaitResult::Error(_) => {
                        unreachable!("received error while performing write")
                    }
                }
            })
        }
```

This notably does not include cleanup of `result1` or `layout1`, for example:

```
        if layout1.size() != 0 {
            _rt::alloc::dealloc(result1.cast(), layout1);
        }
```

which is present when generating synchronous bindings for `import x: func(x: list<list<u32>>);` for example

### LINKED ISSUE #1180: futures: Write-then-transfer-read-handle does not work
Currently with the async runtime support for Rust if a future is created it yields a (reader, writer) end. They're both considered "local" until the reader is transferred away. If the writer has an ongoing write when the reader is transferred away then the write needs to be promoted from a local write to an intrinsic-based write. That doesn't happen today which means that the future basically gets stuck.

### LINKED ISSUE #1181: futures: Write-then-receive-read-handle does not work
This is similar to #1180 but sort of the other way around. If a component only has the write handle to a future and initiates a write it uses an intrinsic to do that. If the read end of the future is then received, however, then the future becomes a "local future". If the future is then read from Rust everything will get stuck as the writer is using an intrinsic where the reader is using in-guest facilities.

### LINKED ISSUE #1182: rust async: Why is using a dummy waker ok?
I've got a lot of jumbled thoughts about wakers and async mutexes (a la https://github.com/WebAssembly/component-model/issues/459) but this is an attempt to mind to crystallize at least something into a coherent thought.

Today tasks are polled with a [dummy waker](https://github.com/bytecodealliance/wit-bindgen/blob/e0d2b889411acf60738dcd976ed5373f7a81e651/crates/guest-rust/rt/src/async_support.rs#L100). That seems suspicious to me and I'm not sure how that works. I'll note that despite this wakers are [indeed kept track of](https://github.com/bytecodealliance/wit-bindgen/blob/e0d2b889411acf60738dcd976ed5373f7a81e651/crates/guest-rust/rt/src/async_support.rs#L52) to a certain degree. I think that's required though because if wakers were ignored completely I don't think that `FuturesUnordered` would work at all. This means Rust is in a state of wakers are sort of kept track of but not at full fidelity where the outermost waker doesn't actually do anything. My goal is to build a mental model for why this is ok (or a reason why it's not).

Here's a contrived situation where I don't think this "ok" currently:

* Task T1 has a future and starts a read. That inserts a channel in `CALLS` to receive the completion signal.
* Task T1 then sends this entire future to task T2
* Task T2 re-polls the future, still sees the channel isn't here
* Task T1 receives the notification when the future is done, and sends it along the channel and wakes up T2's waker. Nothing happens.

Ok now that I wrote this down this may end up just being the same as https://github.com/WebAssembly/component-model/issues/459. Basically there are no facilities for cross-task communication right now in the component model, and I'm not sure we can get all of Rust's async bits and bobs working without it. In theory what should happen here is that if task T2 polls a future created by T1 then either (a) something is reparented to route notifications to T2 (maybe canceling the read and resuming it?) or (b) T2 needs to arrange such that when a notification for T1 comes in that gets redirected and re-routed to T2 instead.

Ok but now after writing that down maybe the answer for futures/streams is cancelling the call and resuming it later. But after writing that down I disagree with myself because that's not possible for general import calls. I understand that they can't currently be canceled but even if they could be it wouldn't be semantically different trying to restart on another task.

Maybe what's needed here is a form of "reparenting". I think we have all the necessary bits of information to detect when such reparenting would be necessary in Rust. Or well maybe not, what I'm thinking is other tasks acquire ownership of objects they're given (e.g. in T2's poll above it'd realize the future was originally from T1). That means that T2 could complete T1, but that's not allowed in the component model because then T1 isn't able to return.

I'm hurting myself in confusion...

### LINKED ISSUE #1184: rust async: Closing a reader might raise a trap as it actually needs to close a writer
Today if you create a (reader, writer) pair for streams/futures if the writer is dropped before the reader then when the reader is dropped it'll use the `close-readable` intrinsic. I think this will raise a trap thoug as it's actually a writable end, not a readable end.

### LINKED ISSUE #1188: rust async: Currently cancelling a stream/future read/write discards the intrinsic return value
Currently cancelling a stream/future read/write discards the return value from the imported intrinsic. According to [this](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#-canon-streamfuturecancel-readwrite) it might even be possible for the intrinsic to not actually complete and return that nothing was cancelled. My hunch is that something needs to be handled here, so I wanted to file an issue to come back and take a look at this.

### LINKED ISSUE #1189: rust: ErrorContext ownership isn't modeled in APIs
This input:

```wit
package test:foo;

world the-world {
  import x: func(x: list<error-context>);
}
```

generates:

```rust
pub fn x(x: &[wit_bindgen::rt::async_support::ErrorContext]) -> () {
```

but I believe that error-context values are passed by ownership, so this semantically isn't the right signature for Rust.

### LINKED ISSUE #1190: rust: Stream ownership is cumbersome
Currently streams implement `Sink` and `Stream` for `Vec<T>`, but this has a number of implications:

* For primitives such as `stream<u8>` that means that sending chunks of data requires sending chunks of allocations. There's no chance to reuse buffers or otherwise have more efficient buffer handling.
* For non-primitives such as `stream<list<u8>>` it's possible that a write doesn't fully send everything. There's no way to recover items which weren't sent in a chunk. I realize right now that `write`-as-generated writes everything in a loop, but I think this is probably also something worth changing.
* If a write is cancelled there's no way currently to recover the original chunk of items that were being sent.

### LINKED ISSUE #1191: rust: Cannot recover values from cancelled writes in futures
Currently the `FutureWriter::write` API doesn't enable recovering the value that's being sent after the send is started.

alexcrichton:
Mostly remove the old implementations for futures/streams and replace them with new versions which address a number of preexisting issues and additionally adapt to upcoming ABI changes.

Closes #1152 - there's now a more custom interface for streams which doesn't use sink/stream traits
Closes #1174 - error-context has been removed from futures/streams
Closes #1177 - all "core" futures are now manual state machines, not `async fn`
Closes #1178 - lowering/lifting was adjusted to avoid leaking anything
Closes #1180 - this should now work
Closes #1181 - the "local" case will trap now in the component model ABI 
Closes #1182 - dummy wakers haven't been used for awhile, doubly so here
Closes #1184 - futures/streams are reader/writer pairs with much simpler ownership
Closes #1188 - the intrinsic return value is now handled and accounts for the operation succeeding racily
Closes #1189 - `error-context` integration with streams/futures is removed
Closes #1190 - `Stream{Writer,Reader}` take and give buffers, no allocations for `stream<u8>`
Closes #1191 - `FutureWriter::write::cancel` returns `Result<(), FutureWriteCancel<T>>`


alexcrichton:
This is a bit of a work-in-progress, but opening this up for upstream PRs. Will flesh out more in the future

crates/guest-rust/rt/src/async_support.rs
```diff
@@ -397,19 +344,21 @@ unsafe fn callback_with_state(
...
+
+        EVENT_STREAM_READ | EVENT_STREAM_WRITE | EVENT_FUTURE_READ | EVENT_FUTURE_WRITE => {
+            (*state).remove_waitable(event1 as u32);
+            let (waker, code) = (*state).wakers.remove(&(event1 as u32)).unwrap();
+            *code = Some(event2 as u32);
+            waker.wake();
+
+            let done = poll(state).is_ready();
+            callback_code(state, done)
+        }
```
alexcrichton, line 360: I'll note that in the future I want to fold `EVENT_CALL_RETURNED` into this exact same pattern as well, but that'll require some refactoring to achieve that.

crates/guest-rust/rt/src/async_support/abi_buffer.rs
alexcrichton, line 1: I'll note that there are no tests of this in this repository, I'd like to ideally flesh some out before landing this

crates/guest-rust/rt/src/async_support/future_support.rs
```diff
@@ -1,50 +1,53 @@
-extern crate std;
+//! Runtime support for `future<T>` in the component model.
+//!
+//! TODO:
+//!
+//! * leaking requires owned values
+//! * owned values means we can't return back sent items
+//! * intimately used in implementation details.
```
alexcrichton, line 7: I do want to expand long-form documentation here about all the concerns around leaks/ownership/io_uring-lookalike/etc.

crates/guest-rust/rt/src/async_support/stream_support.rs
```diff
@@ -1,133 +1,109 @@
...
-    ) -> Pin<Box<dyn Future<Output = Option<Result<usize, ErrorContext>>> + '_>>,
-    pub cancel_write: unsafe extern "C" fn(future: u32) -> u32,
-    pub cancel_read: unsafe extern "C" fn(future: u32) -> u32,
-    pub close_writable: unsafe extern "C" fn(future: u32, err_ctx: u32),
-    pub close_readable: unsafe extern "C" fn(future: u32, err_ctx: u32),
-    pub new: unsafe extern "C" fn() -> u32,
+    pub layout: Layout,
+    pub lower: Option<unsafe fn(value: T, dst: *mut u8)>,
+    pub dealloc_lists: Option<unsafe fn(dst: *mut u8)>,
+    pub lift: Option<unsafe fn(dst: *mut u8) -> T>,
```
alexcrichton, line 23: I'll note that these are crucially `Option` which indicates that the raw representation of `T` can be used to avoid lifting/lowering overhead, aka the `stream<u8>` optimization. I should document this.

crates/guest-rust/rt/src/lib.rs
```diff
@@ -116,3 +122,55 @@ pub fn run_ctors_once() {
 /// Support for using the Component Model Async ABI
 #[cfg(feature = "async")]
 pub mod async_support;
+
+/// Cleanup helper used to deallocate blocks of canonical ABI data from
+/// lowerings.
+pub struct Cleanup {
```
alexcrichton, line 128: I'll note that adding this is going to wreak havoc with `cargo component`. I don't know what to do about that.

crates/rust/src/interface.rs
```diff
@@ -2931,21 +2798,3 @@ impl<'a, 'b> wit_bindgen_core::AnonymousTypeGenerator<'a> for AnonTypeGenerator<
         self.interface.push_str(">");
     }
 }
-
-fn stream_direct(ty: &Type) -> bool {
```
alexcrichton, line 2935: I've replaced this with `is_list_canonical` which already existed but was somewhere else, so I refactored it to be available here.

tests/runtime-new-async-disabled/future-cancel-write-then-read/runner.rs
alexcrichton, line 1: I'll note that I've temporarily disabled these tests, I might remove them from this PR, still figuring out what to do here.

crates/guest-rust/rt/src/lib.rs
alexcrichton: Ok did some thinking (aka paused and did a think) -- I think we should go ahead and land this refactoring. This means that `wit-bindgen` basically can't get updated in `cargo-component` without breaking things, but it won't break anything itself. Not being able to update for a bit is I think probably the best tradeoff here for now.

crates/guest-rust/rt/src/async_support/future_support.rs
```diff
@@ -61,227 +238,243 @@ impl<T> fmt::Debug for FutureWriter<T> {
...
+    ///   as well to resume a write in the future if desired.
+    ///
+    /// Note that if this method is called after the write was already cancelled
+    /// then `Ok(())` will be returned.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the operation has already been completed via `Future::poll`,
+    /// or if this method is called twice.
+    pub fn cancel(self: Pin<&mut Self>) -> Result<(), FutureWriteCancel<T>> {
```
dicej, line 420: What's the rationale for using `Result` here rather than adding a `FutureWriteCancel::AlreadyDelivered` variant and having this return `FutureWriteCancel<T>` instead?  It's a little surprising to me that we'd consider `FutureWriteCancel::Cancelled` (or really any of the possible outcomes) to be an error.

crates/guest-rust/rt/src/async_support/stream_support.rs
```diff
@@ -1,133 +1,187 @@
...
-            vtable,
+    pub unsafe fn new(handle: u32, vtable: &'static StreamVtable<T>) -> Self {
+        Self { handle, vtable }
+    }
+
+    /// Initiate a write of the `values` provided into this stream.
+    ///
+    /// This method is akin to an `async fn` except that the returned
+    /// [`StreamWrite`] future can also be cancelled via [`StreamWrite::cancel`]
+    /// to re-acquire intermediate values.
```
dicej, line 100:
```suggestion
    /// to re-acquire undelivered values.
```
This is just a matter of opinion, but "undelivered" is more meaningful to me in this context than "intermediate".

crates/guest-rust/rt/src/async_support/stream_support.rs
```diff
@@ -1,133 +1,187 @@
...
-        });
+    /// This is a higher-level method than [`StreamWriter::write`] and does not
+    /// expose cancellation for example. This will attempt to send `value` on
+    /// this stream.
+    ///
+    /// If the other end hangs up then the value is returned back as
+    /// `Some(value)`, otherwise `None` is returned indicating the value was
+    /// sent.
+    pub async fn write_one(&mut self, value: T) -> Option<T> {
+        // TODO: can probably be a bit more efficient about this and avoid
```
dicej, line 181: FWIW, [this is how I addressed it on the host side](https://github.com/bytecodealliance/wasip3-prototyping/blob/80a85e686a1722469891d07c4246dc26f98dc322/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs#L242-L262), i.e. by making the buffer type generic and implementing it for `Vec`, `Bytes`, and `Single`.  That would require significant refactoring here, though, given the way `AbiBuffer` is currently tied to `Vec`.

crates/guest-rust/rt/src/async_support/waitable.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// status code returned by the `{future,stream}.cancel-{read,write}`
+    /// intrinsic.
+    fn in_progress_cancel(state: &Self::InProgress) -> u32;
+
+    /// Completion callback for when an operation was canceled.
+    ///
+    /// This is invoked after `in_progress_cancel` is used and the returned
+    /// status code indicates that the operation was indeed cancelled and didn't
+    /// racily return some other result.
+    fn in_progress_canceled(state: Self::InProgress) -> Self::Result;
```
dicej, line 109: Some comments and code spell it `cancelled` and some spell it `canceled`.  Apparently [either is correct](https://www.merriam-webster.com/grammar/canceled-or-cancelled), but we should probably choose one and use it consistently.

dicej (approved):

crates/guest-rust/rt/src/async_support/abi_buffer.rs
```diff
@@ -0,0 +1,417 @@
...
+        assert_eq!(vtable.lower.is_some(), vtable.lift.is_some());
+
+        // SAFETY: We're converting `Vec<T>` to `Vec<MaybeUninit<T>>`, which
+        // should be safe.
+        let rust_storage = unsafe {
+            let ptr = vec.as_mut_ptr();
+            let len = vec.len();
+            let cap = vec.capacity();
+            mem::forget(vec);
+            Vec::<MaybeUninit<T>>::from_raw_parts(ptr.cast(), len, cap)
```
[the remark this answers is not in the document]
alexcrichton: AFAIK this is a situation where it works and probably will work for a long time but everyone will still frown at you for doing it, so I've opted to be a bit safer here and sidestep the question of "is it ok?"

crates/guest-rust/rt/src/async_support/stream_support.rs
alexcrichton: Good point yeah, I kind of want to avoid the complexity for now but I  do still want to leave a note about a potential future optimization, so I'm probably going to leave this as is (and might advocate for the host to do the same in the near term...)

dicej: To be clear, the host optimization was primarily motivated by being able to support `bytes::Bytes` (which is what Hyper uses out of the box, and thus `wasmtime-wasi-http`) natively without any conversions.  The single-element optimization was just a nice side-effect of generic-izing the code.

PR #1250, opened 2025-04-01
Migrate resource-alias-redux test to new framework

alexcrichton:


alexcrichton:
@jsturtevant I think that this is exposing a preexisting bug in the C# generator, I've added a test at `_test_disabled.cs` here which fails to compile in the generated code because I don't think it's `using` the right paths, but I could also very well be misunderstanding something about C# here in this respect.

PR #1260, opened 2025-04-04
 Refactor how imported tasks and bindings work 

alexcrichton:
Some more details in the commit messages, but the main changes here are:

* Imported subtasks are managed entirely differently now.
* Imported functions now take all arguments by ownership, nothing is borrowed except for borrowed resources.
* This fixes an invalid-free of a stack pointer when the `block_on` intrinsic is used.
  * Lots of refactoring was needed to achieve this in a "clean" way
* In the future it should be much easier to "slot in" cancellation of imported subtasks.

### DIFF
--- a/crates/core/src/abi.rs
+++ b/crates/core/src/abi.rs
@@ -914,7 +914,7 @@ impl<'a, B: Bindgen> Generator<'a, B> {
                     }
                 } else {
                     // ... otherwise if parameters are indirect space is
-                    // allocated from them and each argument is lowered
+                    // allocated for them and each argument is lowered
                     // individually into memory.
                     let ElementInfo { size, align } = self
                         .bindgen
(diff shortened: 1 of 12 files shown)


crates/core/src/abi.rs
```diff
@@ -894,84 +904,62 @@ impl<'a, B: Bindgen> Generator<'a, B> {
...
+                    // If the parameters for this function aren't indirect
+                    // (there aren't too many) then we simply do a normal lower
+                    // operation for them all.
+                    for (nth, (_, ty)) in func.params.iter().enumerate() {
+                        self.emit(&Instruction::GetArg { nth });
+                        self.lower(ty);
+                    }
+                } else {
+                    // ... otherwise if parameters are indirect space is
+                    // allocated from them and each argument is lowered
```
FrankReh, line 917: "from them" or "for them"? Maybe it's as intended.

dicej: Yeah, I think this should be "for them"

```suggestion
                    // allocated for them and each argument is lowered
```

dicej (approved):

PR #1264, opened 2025-04-08
Implement async task cancellation 

alexcrichton:
This commit implements async task cancellation in Rust for subtasks
created through imported calls. This involved some extra
bindings-related code to drop both lists and owned handles (effectively
a full destructor) along with some refactorings of how waitables are
handled due to new states that can pop up.

### DIFF
--- a/crates/core/src/abi.rs
+++ b/crates/core/src/abi.rs
@@ -554,6 +554,10 @@ def_instruction! {
             blocks: usize,
         } : [1] => [0],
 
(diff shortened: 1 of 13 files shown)


alexcrichton:
Ok with tests I think this is now good to go. I've currently based this on https://github.com/bytecodealliance/wit-bindgen/pull/1287 due to relying on a few refactorings there to get the tests working

dicej (approved):

PR #1268, opened 2025-04-09
Refactor how guest exported tasks work

alexcrichton:
This commit refactors some of the code generation and ABI instructions used for exported tasks. Instead of a future + closure they're represented now only as a single future which internally handles the invocation of `task.return`. No major functional change here, just some refactoring.


crates/rust/src/interface.rs
```diff
@@ -1004,34 +1004,25 @@ unsafe fn call_import(params: *mut u8, results: *mut u8) -> u32 {{
...
-            async_result_name,
             ..
         } = f;
         if async_ {
+            let async_support = self.r#gen.async_support_path();
+            uwriteln!(self.src, "{async_support}::start_task(async move {{");
             if needs_cleanup_list {
-                let vec = self.path_to_vec();
-                uwriteln!(self.src, "let mut cleanup_list = {vec}::new();");
+                uwriteln!(self.src, "let mut cleanup_list = Vec::new();");
```
dicej, line 1013: Why do we not need to use `self.path_to_vec()` here?

dicej (approved):
LGTM, assuming the `wasip3-prototyping` tests pass.

crates/rust/src/interface.rs
alexcrichton: Oh this is my mistake, we do still need that and this just got lost in some refactorings I made along the way and didn't double-check the diff at the end.

PR #1269, opened 2025-04-10
Finish migration from old-style runtime tests to `wit-bindgen test`

alexcrichton:
Copy over `options` and `results` tests and then delete all the old infrastructure.


jsturtevant (approved):

alexcrichton:
Alas I was hoping for more of a speedup in CI but in the end C# on windows is still clocking in at around 17 minutes...

PR #1273, opened 2025-04-15
Change how async stream/future results are encoded

alexcrichton:
Updating to WebAssembly/component-model#498

### DIFF
--- a/crates/guest-rust/rt/src/async_support/stream_support.rs
+++ b/crates/guest-rust/rt/src/async_support/stream_support.rs
@@ -2,7 +2,7 @@
 //! module documentation in `future_support.rs`.
 
 use crate::async_support::waitable::{WaitableOp, WaitableOperation};
-use crate::async_support::{results, AbiBuffer};
+use crate::async_support::{AbiBuffer, ReturnCode};
 use {
     crate::Cleanup,
     std::{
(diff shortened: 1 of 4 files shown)


crates/guest-rust/rt/src/async_support/stream_support.rs
```diff
@@ -480,20 +480,20 @@ where
...
             // When an in-progress read is successfully cancelled then the
             // allocation that was being read into, if any, is just discarded.
             //
             // TODO: should maybe thread this around like `AbiBuffer` to cache
             // the read allocation?
-            results::CANCELED => Ok((StreamResult::Cancelled, buf)),
+            ReturnCode::Cancelled(0) => Ok((StreamResult::Cancelled, buf)),
 
-            amt => {
+            ReturnCode::Completed(amt) | ReturnCode::Closed(amt) | ReturnCode::Cancelled(amt) => {
```
alexcrichton, line 496: I'll note that for now I've opted to just lose information for here as it's, in theory harmless. For example if a stream is closed after transferring N items then future writes should in theory not even hit the host and immediately say "nope it's close", or perhaps the state is reflected in the return value where the stream is consumed. I don't think it's super important to expose these states just yet so I've left that for a future update.

dicej (approved):

PR #1275, opened 2025-04-16
Use fully qualified type names

### LINKED ISSUE #1265: C#: Generated code contains compilation errors
Hello. I've used https://github.com/bytecodealliance/componentize-dotnet for the first time today. I ended up incredibly disappointed.

The main problem is a generated file containing some `Result`-related definitions. In the code, two `ArgumentException`s are thrown. There is no reference to the `System` namespace, such that compiling the library results in `error CS0246: The type or namespace name 'ArgumentException' could not be found`.

---

I have multiple other things I would like to address and discuss. Please let me know if I should take this somewhere else.
1. Naming.
  - The generated code contains namespaces with parts which are all lowercase. This is unacceptable in C# and actually results in errors in all of my codebases.
  - The generated namespaces use the target world's name (`world foo` -> `namespace FooWorld`). This means it does not use the containing library's root namespace. This, along with generating `FooWorld.wit.imports.sample.pkg`, takes a lot of control out of the user's hands. I do not want the code to be generated in these namespaces.
2. Types and declarations.
  - In simple examples, `Result<TOk, TErr>` and `None` (as well as `IFooWorld` and `FooWorld`) are not used at all. Why are these generated in those cases?
  - Why is `Result<TOk, TErr>.Tags` generated? The constants (if required at all) can be declared at the class level in `Result<TOk, TErr>`.
  - Why are these all declared as `public`? I do not want these types to be part of my library. A consumer of my library should never know about internal implementation details. Can I configure this?
    I may even want to wrap the generated imports and exports further. Everything being public is a nuisance for this purpose.
3. `DllImport` module and entry point name.
  - My host app does not export functions in `sample:pkg/foo`, but instead in `env`. The generated code always uses `[DllImport("sample:pkg/foo", ...), ...]`. Can I configure this?
  - My host app does not export functions as `kebab-case`, but `snake_case`. The generated code always uses `kebab-case`. Can I configure this?
4. This should be a source generator.
  - The fact that this is a separate software from `componentize-dotnet` is a huge problem for this point. An incremental source generator using additional files provided by the user would be much more convenient to use. The code would be generated while the user types, not upon executing a `build`. This would require the bindings generation to be done completely in C#, but is a lot more pleasant to use.

just-ero:
Closes #1265.


crates/csharp/src/function.rs
```diff
@@ -948,7 +935,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 uwrite!(
                     self.src,
                     "
-                    var {array} = new List<{ty}>({length});
+                    var {array} = new global::System.Collections.Generic.List<{ty}>({length});
```
pavelsavara, line 938: Does `type_name_with_qualifier` give `ty` the global name too ?

just-ero: I do not believe so. Please lead me in the right direction to make this happen. It should be unnecessary as long as `ty` is either a primitive or a type already contained within the same namespace as this emitted code.

pavelsavara: I don't know much about the generator codebase, @jsturtevant please advise.

I just noticed that different flavors `ty` is sprinkled thru the code.

I'm thinking that we could have built-in hash table (or switch) translating known system types from short name to long name.  All of them for consistency sake. 

For `int` -> `global::System.Int32`

I guess later we will also generate marshaling for `System.Threading.Tasks.Task` which is not in `System` namespace.

And I think WIT can do list of list ?
`global::System.Collections.Generic.List<global::System.Collections.Generic.List<global::System.Int32>>`

just-ero: I don't personally believe that primitives (`int`) should be fully specified, but that's up to you.

just-ero: In general, one requirement of generated code (at least for .NET) is to fulfill the lowest common denominator. In C# that might mean not using interpolated strings, file-scoped namespaces, or other more modern features, as the user may target a language version which does not support those features.

`componentize-dotnet` is basically hard locked to projects which target .NET 10 and up, so this would only be a problem if the user explicitly downgraded using the `LangVersion` MSBuild property. It's not a scenario I would worry too much about, but I thought I'd mention it.

pavelsavara: Let's not worry about below Net10

jsturtevant: > Fully qualified type names are safer, as they leave no ambiguity.

This is compelling to me for generated code.  So I am +1 to this change.

jsturtevant (approved):

just-ero:
Hello, any news on this?

just-ero:
I don't believe this PR was ready for merge yet.

> Does `type_name_with_qualifier` give `ty` the global name too ?  
> No `type_name_with_qualifier` will not add `global::` in-front of the type name.  
> And I think WIT can do list of list ?  
> Yes it can.  

I was hoping for some input to address these problems. I was also waiting for input on the `const` question I asked above.

PR #1277, opened 2025-04-20
Initial fixed-length list implementation

cpetig:
See https://github.com/WebAssembly/component-model/pull/384 for the commit which added fixed size lists to the component model standard documents.


crates/core/src/abi.rs
```diff
@@ -307,6 +307,28 @@ def_instruction! {
...
+            size: u32,
+            id: TypeId,
+        } : [*size as usize] => [1],
+
+        /// Pops an array off the stack, decomposes the elements and then pushes them onto the stack.
+        FixedSizeListLower {
+            element: &'a Type,
+            size: u32,
+            id: TypeId,
+        } : [1] => [*size as usize],
```
alexcrichton, line 323: For this I'll echo some of my thoughts on the sibling PRs: I think we'll want to have this with a different shape where the size of the list isn't affecting the O(...) processing here. For example this is pushing/popping the list items from the operand stack of the ABI code, but I think we're going to want something different more akin to how dynamically sized lists work to avoid generating an exponential amount of code in the code generators

cpetig:
I have a newer version at https://github.com/cpetig/wit-bindgen/tree/work-in-progress which tests more in depth and also supports C++, I need to separate and clean up that one and then push it here for review.

cpetig:
Also this requires https://github.com/bytecodealliance/wasmtime/pull/10619 , and a very recent version of wasm-component-ld for the tests to pass.

crates/core/src/abi.rs
cpetig: Now, lifting from and lowering to memory uses constant code size, only when the list gets flattened in arguments the code remains O(N).

I will now fix the wasmtime implementation as well.

cpetig:
This is ready, but depends on https://github.com/bytecodealliance/wasmtime/pull/10619 for successful tests.

crates/test/src/runner.rs
```diff
@@ -45,7 +45,8 @@ impl TestRunner {
         for arg in self.args.iter() {
             ret.arg(arg);
         }
-        ret.arg("--invoke=run()");
+        ret.arg("--invoke=run()")
+            .arg("-Wcomponent-model-fixed-size-list");
```
alexcrichton: For this, for now, this'll go in `.github/workflows/main.yml`. In retrospect this should go in the test file itself with some sort of `//@ ...` option or something like that. I can investigate poking around with the latter, so if you want to move this line itself to `.github/` that should be sufficient.

alexcrichton (commented):
Looks good to me!

crates/test/src/runner.rs
alexcrichton: In https://github.com/bytecodealliance/wit-bindgen/pull/1497 I'm moving these flags to test files themselves

cpetig:
This still has a limitation: The generated Rust and C++ code can't handle non-Copy/POD types. For Functional Safety this isn't an issue as you need to avoid putting string or list into a fixed size array anyway, resources might still make some sense.

I created https://github.com/bytecodealliance/wit-bindgen/issues/1501 to track this limitation.

cpetig:
TODO: Check for correct usage of "fixed-length" over "fixed-size"

cpetig:
~~I guess I should fix the naming in wasm-tools (fixed *length* list) before merging this MR (as it shouldn't be urgent),~~ but on the other hand that merge will affect all of wit-bindgen languages anyway.

Second thought: The impact of the wasm-tools change will be identical (all code generators in wit-bindgen have to do a search and replace independently of whether this patch is merged or not). But it will create a conflict with this patch, so I guess it can be merged and then the other change can be discussed separately.

cpetig:
The (discussion of the) rename in wasm-tools happens in https://github.com/bytecodealliance/wasm-tools/pull/2421

alexcrichton (approved):
Let's go ahead and get this merged and renames/refactors can land in parallel

PR #1278, opened 2025-04-22
Update how async configuration is processed

alexcrichton:
This commit updates the Rust bindings generator for async to respect the `async` annotations in WIT by default. The `--async` flag and `async: { ... }` configuration have changed as well to be modeled as "this is an override" where the fallback is "what the WIT says".


alexcrichton:
@dicej https://github.com/bytecodealliance/wasip3-prototyping/pull/132 is the corresponding change to p3-prototyping which you may want to peruse as well

dicej (approved):

PR #1281, opened 2025-04-23
Skip rustfmt on the link_section part of the code too

bnjbvr:
As the link section value can be short enough that `cargo-fmt` decides to put it on one line, when prettyplease puts it on two, in some cases.


alexcrichton (approved):
Thanks!

PR #1299, opened 2025-05-19
Document fully qualified worlds in generate!

Xe:
This nudge would have saved me some time in trying to figure out how to share .wit interfaces between components in a monorepo.

### DIFF
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -643,6 +643,14 @@
 ///     // The name of the world that bindings are being generated for. If this
 ///     // is not specified then it's required that the package selected
 ///     // below has a single `world` in it.
+///     //
+///     // Optionally, you can specify the fully qualified path to the world
+///     // to read from the `wit/deps` folder:
+///     //
+///     // "example:test/my-world"
+///     //
+///     // This would read from your `my-world.wit` file in
+///     // `wit/deps/example:test/my-world`.
 ///     world: "my-world",
 ///
 ///     // Path to parse WIT and its dependencies from. Defaults to the `wit`


crates/guest-rust/src/lib.rs
```diff
@@ -643,6 +643,14 @@
...
 ///     // is not specified then it's required that the package selected
 ///     // below has a single `world` in it.
+///     //
+///     // Optionally, you can specify the fully qualified path to the world
+///     // to read from the `wit/deps` folder:
+///     //
+///     // "example:test/my-world"
+///     //
+///     // This would read from your `my-world.wit` file in
+///     // `wit/deps/example:test/my-world`.
```
alexcrichton, line 653: To clarify this slightly, the names of folders within `wit/deps` are insignificant and are only used for deduplicating packages. The `package example:test` can be within `wit/deps/test` or `wit/deps/example` equally. It's the contents of the `*.wit` files within that folder which dictate the package name.

Additionally the name `my-world` doesn't have to match the name of the WIT file itself, for example you could have `world my-world { ... }` in `wit/deps/test/test.wit`. You could even go further with a single-file package and have `wit/deps/test.wit` which contains `package example:test;` and `world my-world { .. }`.

Mostly would you ok relaxing the filename recommendation here? The wording currently reads as if it's required to be in such a location which may cause confusion.

alexcrichton: To confirm, you do not wish to change the wording here? If so I can incorporate this change in a separate PR but personally I won't merge this until the wording no longer implies that such filesystem naming is required.

Xe: I'm okay with changing the wording. I'll think about it overnight and file a suggestion in this thread.

PR #1306, opened 2025-05-27
Support "linear writes" in Rust `future` bindings

### LINKED ISSUE #1304: Ensure `FutureWriter` always writes
Discussed some time ago, forgotten, and then reminded today, the expectation is that at the component model level dropping a writable-end future will be a trap if nothing was actually written. Modeling this in Rust without actually surfacing traps will be done with a few changes:

* When a `FutureWriter<T>` is constructed an additional parameter `default: fn() -> T` will be required. This will be stored within the `FutureWriter<T>` return value.
* When a `FutureWriter<T>` is dropped without recording a successful write then a task will be spawned which actually performs the write (e.g. calls the `default` and then performs the write for you).

This needs to be implemented in the Rust bindings.

alexcrichton:
To account for WebAssembly/wasip3-prototyping#172 and WebAssembly/component-model#521 the Rust bindings need to take into account that a future writer cannot be dropped without actually writing something. To model this futures now must be created with a thunk that creates the default value to be sent. This value is only used if another value was not otherwise written.

Closes #1304

### DIFF
--- a/crates/guest-rust/rt/src/async_support/future_support.rs
+++ b/crates/guest-rust/rt/src/async_support/future_support.rs
@@ -86,6 +86,30 @@
 //! represented the same in Rust as it is in the canonical ABI. That means that
 //! sending `list<T>` into a future might require copying the entire list and
 //! changing its layout. Currently this is par-for-the-course with bindings.
+//!
+//! ## Linear (exactly once) Writes
+//!
+//! The component model requires that a writable end of a future must be written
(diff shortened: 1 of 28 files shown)


alexcrichton:
For now tests are "updated enough" to pass but I suspect that more tests will need to be updated once the host is also enforcing this. Notably I suspect some C tests will need to be updated.

dicej (approved):

PR #1311, opened 2025-06-03
Async trait function error

### LINKED ISSUE #1312: Cannot generate resource with an async function in it
A wit file like this:

```wit
package wha:hoo;

interface iii {
  resource rrr {
    fff: async func();
  }
}

world core {
  export iii;
}
```
will generate an error like this

```
               |
           156 | ...                   { T::fff(RrrBorrow::lift(arg0 as u32 as usize)).await };
               |                         ------ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&T`, found `RrrBorrow<'_>`
               |                         |
               |                         arguments to this function are incorrect
               |
               = note: expected reference `&T`
                             found struct `RrrBorrow<'_>`
```

error is exposed in https://github.com/bytecodealliance/wit-bindgen/pull/1311.

elmerbulthuis:
- fixes #1312


alexcrichton:
Thanks! I'll try to get to this in the near future

crates/rust/src/bindgen.rs
```diff
@@ -897,7 +897,10 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                     // Automatically convert `Borrow<'_, AResource>` to
                     // `&Self` since traits have `&self` as their
                     // first arguments.
-                    if i == 0 && matches!(func.kind, FunctionKind::Method(_)) {
+                    if i == 0
```
elmerbulthuis, line 900: in case of an async method we should also use .get()

elmerbulthuis:
One failing test, maybe this is fixed in https://github.com/bytecodealliance/wit-bindgen/pull/1317 ?

alexcrichton:
Oh thank you for this! The fix here looks right to me, and the failing test there is expected as that's transient where we're trying to nail down exactly what the semantics of various edge-cases are.

PR #1313, opened 2025-06-06
update `future.{write,read}` ABIs

dicej:
This updates the Rust generator to use the new ABI defined in https://github.com/WebAssembly/component-model/pull/524.  The C generator will need a similar update as well.

- `future.write` now accepts its payload value as up to 4 flat parameters before spilling to linear memory.

- `future.read` takes no payload pointer when it has no payload type

- `{stream,future}.close-{readable,writable}` have been renamed to `{stream,future}.drop-{readable,writable}`

I've tested this end-to-end with `wasip3-prototyping` and will soon post a corresponding PR there.

Finally, `rustfmt` was complaining about using `gen` as a field name, so I took this opportunity to address that.


dicej:
I should note that this doesn't cover all of https://github.com/WebAssembly/component-model/pull/524.  Specifically, these items will need work (presumably just runtime test updates in this repo once `wasip3-prototyping` has been updated):

> - There is no "number of elements written" packed in the high 28 bits of the future.{read,write} results.
> - On successful copy, future.{read,write} return COMPLETED instead of CLOSED (which, as noted above, was renamed to DROPPED, making it especially "wrong" as the result code).

dicej:
Alex, Luke, and I just discussed this and decided we're going to scale back these changes given how big they turned out to be implementation-wise.  I'm going to close this and open a new one next week.

PR #1315, opened 2025-06-09
update `future`/`stream` ABIs and names

dicej:
This updates `wit-bindgen-rt` etc. to use (some of) the new ABI defined in https://github.com/WebAssembly/component-model/pull/524.  It covers everything in that PR _except_ the lifting and lowering changes to `future.{read,write}`, which we decided are more trouble than they're worth.

I've tested this end-to-end with `wasip3-prototyping` and will soon post a corresponding PR there.

Finally, `rustfmt` was complaining about using `gen` as a field name, so I took this opportunity to address that.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -201,20 +201,20 @@ impl WorldGenerator for C {
         _files: &mut Files,
     ) -> Result<()> {
         let wasm_import_module = resolve.name_world_key(name);
-        let mut gen = self.interface(resolve, true, Some(&wasm_import_module));
-        gen.interface = Some((id, name));
-        gen.define_interface_types(id);
+        let mut r#gen = self.interface(resolve, true, Some(&wasm_import_module));
+        r#gen.interface = Some((id, name));
+        r#gen.define_interface_types(id);
 
         for (i, (_name, func)) in resolve.interfaces[id].functions.iter().enumerate() {
             if i == 0 {
                 let name = resolve.name_world_key(name);
-                uwriteln!(gen.src.h_fns, "\n// Imported Functions from `{name}`");
-                uwriteln!(gen.src.c_fns, "\n// Imported Functions from `{name}`");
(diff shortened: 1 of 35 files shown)


crates/guest-rust/rt/src/async_support/future_support.rs
```diff
@@ -376,24 +376,24 @@ where
         match ReturnCode::decode(code) {
```
alexcrichton: This should be able to stop using `ReturnCode::decode` I think, right? In that there should be a small handful of codes to test for here and that's it?

crates/c/src/lib.rs
```diff
@@ -201,20 +201,20 @@ impl WorldGenerator for C {
         _files: &mut Files,
     ) -> Result<()> {
         let wasm_import_module = resolve.name_world_key(name);
-        let mut gen = self.interface(resolve, true, Some(&wasm_import_module));
-        gen.interface = Some((id, name));
-        gen.define_interface_types(id);
+        let mut r#gen = self.interface(resolve, true, Some(&wasm_import_module));
```
alexcrichton, line 204: I can do this in a follow-up but I'd prefer to avoid `r#gen`. I understand it's the machine-applicable way to migrate from one edition to another but no one really wants to either read or write `r#gen` over time so I'd prefer to go through and update things to `generator` or similar.

crates/guest-rust/rt/src/async_support/future_support.rs
dicej: You mean e.g. `match code { BLOCKED => { ... } c@(DROPPED | CANCELLED) => { ... } COMPLETED => { ... } }`?  Yeah I can do that.

alexcrichton: Indeed!

dicej:
I'm going to try to fix the `Test Async` CI job before merging this.

PR #1319, opened 2025-06-19
[c] Add documentation for the C binding generator

TartanLlama:
Improves the documentation for the C binding generator:

- Adds a new documentation component to `crates/c/README.md` with extensive documentation on generated bindings
- Links to the documentation from the top-level `README.md`
- Improves the generated binding documentation for `*_string_len`, `*_string_set`, and `*_string_dup`

### DIFF
--- a/crates/c/README.md
+++ b/crates/c/README.md
@@ -26,7 +26,7 @@ This command will generate either two or three files, depending on command line 
 
 In this section, *your component* refers to the component for which you are generating bindings.
 
-In general, your component is responsible for allocating memory for data it produces and freeing memory for data it consumes. `wit-bindgen` automatically generates `*_free` functions for types that require allocation (e.g. strings, lists, records with fields that require allocation) to properly deallocate the memory for the type and any of its nested allocations.
+In general, your component is responsible for allocating memory for data it produces and freeing memory for data it consumes. `wit-bindgen` automatically generates `*_free` functions for types that require allocation (e.g. strings, lists, records with fields that require allocation) to properly deallocate their memory, including any nested allocated types.
 
 There are additional ownership considerations for `string` and `resource`. See the [Strings](#strings) and [Resources](#resources) sections for details.
 
--- a/crates/c/README.md
+++ b/crates/c/README.md
@@ -112,7 +112,7 @@ world cat-registry {
 }
 ```
 
-Here is a possible implementation of the `get-cat-by-name` export that correctly manages the memory of the paremeter and return value:
+Here is a possible implementation of the `get-cat-by-name` export that correctly manages the memory of the parameter and return value:
 
 ```c
 bool exports_cat_registry_cat_registry_api_get_cat_by_name(
--- a/crates/c/README.md
+++ b/crates/c/README.md
@@ -233,7 +233,7 @@ typedef struct my_world_list_u8_t {
 } my_world_list_u8_t;
 ``` 
 
-Helper functions are generated for freeing lists:
+A helper function is generated for freeing lists:
 
 ```c
 // Frees the memory associated with the list
@@ -256,6 +256,12 @@ typedef struct my_world_cat_toy_t {
 #define MY_WORLD_CAT_TOY_BALL 0
 #define MY_WORLD_CAT_TOY_TEDDY 1
 #define MY_WORLD_CAT_TOY_WAND 2
+```
+
+A helper function is generated for freeing variants:
+
+```c
+void my_world_cat_toy_free(my_world_cat_toy_t *ptr);
 ```
 
 #### Results
@@ -293,6 +299,11 @@ extern void my_example_string_getter_get_string_by_index(uint32_t index, my_exam
 extern bool my_example_string_getter_get_string_by_index(uint32_t index, string_getter_user_string_t *ret, my_example_string_getter_error_t *err);
 ```
 
+A helper function is generated to free variants:
+
+```c
+void my_world_result_string_u32_free(my_world_result_string_u32_t *ptr);
+```
 
 #### Options
 
@@ -323,6 +334,12 @@ extern void my_example_string_getter_get_string_by_index(uint32_t index, string_
 
 // Flag not passed, the return value is flattened
 extern bool my_example_string_getter_get_string_by_index(uint32_t index, string_getter_user_string_t *ret);
+```
+
+A helper function is generated to free options:
+
+```c
+void my_world_option_string_free(my_world_option_string_t *ptr);
 ```
 
 #### Enums and Flags
--- a/crates/c/README.md
+++ b/crates/c/README.md
@@ -36,9 +36,9 @@ Ownership rules differ for imported functions and exported functions, so we'll t
 
 Imported functions are functions that are implemented in another component and that your component calls.
 
-**Ownership rules:**
+_Ownership rules_:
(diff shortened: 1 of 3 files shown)


TartanLlama:
Depends on #1320

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+
+## Usage
+
+To generate bindings with this crate, issue the `c` subcommand to `wit-bindgen`:
+
+```bash
+$ wit-bindgen c [OPTIONS] <WIT>
+```
+
+See the output of `wit-bindgen c help` for available options.
```
alexcrichton: I think this'll want to be `wit-bindgen help c` or `wit-bindgen c --help` (as-is `wit-bindgen c help` attempts to read `./help` and look for WIT inside)

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+
+Ownership rules differ for imported functions and exported functions, so we'll tackle these in turn.
+
+#### Imported functions
+
+Imported functions are functions that are implemented in another component and that your component calls.
+
+_Ownership rules_:
+
+- **Arguments**: Your component is responsible for allocating memory for the input arguments and freeing this memory when it is no longer required. Your input data will not be modified or freed by the import's implementation: complex types like strings and records will be copied into the implementation's storage.
```
alexcrichton: > complex types like strings and records will be copied into the implementation's storage

This part is a little subtle and maybe worth expanding? For example below `cat_registry_user_string_dup(&name, "Poptart");` technically isn't necessary since there's no need for the string being passed in to be an alloated string. I think `cat_registry_user_string_set` would work as well in that case.

Put another way the generated import functions should never allocate memory internally, so I'm not sure what you mean here by "the implementation's storage" since in theory it has none. I might be missing something though (or just forgetting how things work!), so I wanted to dig in and/or expand this phrase here

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+```
+
+Here is a possible implementation of the `get-cat-by-name` export that correctly manages the memory of the parameter and return value:
+
+```c
+bool exports_cat_registry_cat_registry_api_get_cat_by_name(
+    cat_registry_string_t *name,
+    exports_cat_registry_cat_registry_api_cat_t *ret) {
+  // We only support the name "Poptart" for this example
+  if (strcmp((const char *)name->ptr, "Poptart") == 0) {
```
alexcrichton: I think this'll want to use `strncmp` or similar instead of `strcmp` here because `name->ptr` isn't nul-terminated. Or I forget which memory-comparing function in C is the best for that...

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+    // Allocate memory for each nickname and copy the strings in
+    cat_registry_string_dup(&ret->nicknames.ptr[0], "Poppy");
+    cat_registry_string_dup(&ret->nicknames.ptr[1], "Popster");
+  }
+
+  // We own the memory for input parameter, so we must free it
+  cat_registry_string_free(name);
+
+  // We allocated memory for the return value, but freeing it is handled by the generated bindings
+  return false;
```
alexcrichton: I think this may want to dynamically return `true` or `false` depending on if `name->ptr` was `"Poptart"`, right?

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+}
+```
+
+Importantly, the generated `*_post_return` function that frees the return value assumes that lists and strings in the return value reference dynamically-allocated memory that can be freed with `free`. As such, if you call `*_string_set` with a string literal to set one of the fields of the return value, the bindings will try to free the string literal. Unless you want to replace the generated `*_post_return` function with a version that only frees some of the data, always dynamically allocate the contents of your strings and lists in your return values.
```
alexcrichton: One thing perhaps worth mentioning here is that `*_post_return` is defined as a weak symbol, so applications can override it by defining their own if there are different ownership rules in use

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+my_example_string_getter_get_string_by_index(
+    uint32_t index, my_example_string_getter_result_string_error_t *ret);
+
+// Flag not passed, the return value is flattened
+extern bool
+my_example_string_getter_get_string_by_index(
+    uint32_t index, string_getter_user_string_t *ret, my_example_string_getter_error_t *err);
+```
+
+A helper function is generated to free variants:
```
alexcrichton: s/variants/results/

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+  adopter_string_free(&name);
+
+  if (got_cat) {
+    // When taking a borrow of a handle we own, we must not drop
+    // the borrowed handle.
+    cat_example_adoption_authority_api_notify_adoption(
+        cat_example_registry_api_borrow_cat(cat));
+
+    // We must drop the owning handle when we are done with it
+    if (got_cat) {
```
alexcrichton: The `got_cat` condition here may be redundant given the enclosing condition?

crates/c/README.md
```diff
@@ -0,0 +1,730 @@
...
+    registry_string_t *name,
+    exports_cat_example_registry_api_own_cat_t *ret) {
+  bool found = false;
+  for (size_t i = 0; i < g_cat_count; i++) {
+    exports_cat_example_registry_api_own_cat_t *cat = &g_cats[i];
+    // Get the internal representation for the owning handle
+    exports_cat_example_registry_api_cat_t *cat_rep =
+        exports_cat_example_registry_api_cat_rep(*cat);
+    // If the name matches, we found the cat to adopt
+    if (strcmp((const char *)cat_rep->name.ptr, (const char *)name->ptr) == 0) {
```
alexcrichton: Similar comment to above about how this may want to use a different comparison function as these aren't guaranteed to be nul-terminated

crates/c/README.md
TartanLlama: By "the implementation's storage" I was meaning that the arguments will be copied into the linear memory of the component that implements the imported function. Agreed on the wording being not great. Rephrasing to 

> Your input data will not be modified or freed when you call the imported function; the data will be copied into the linear memory of the component that implements the function.

Agreed that the string doesn't need to be allocated, I did it this way just to demonstrate that if the string *is* allocated, it's the caller's responsibility to free it. I'm clarifying by adding the following comment:

> // (This could call cat_registry_user_string_set to avoid allocating,
> // the allocation is just for demonstration purposes)

crates/c/README.md
TartanLlama: Agreed, changing to `strncmp`

crates/c/README.md
TartanLlama: Good point, adding
> The `*_post_return` functions are defined as [weak symbols](https://en.wikipedia.org/wiki/Weak_symbol) so that you can simply define your own versions to override the defaults and they will be selected by the linker.

crates/c/README.md
TartanLlama: Indeed, thanks!

crates/c/README.md
TartanLlama: Whoops, thanks

TartanLlama:
I think that's all the feedback resolved, thanks!

PR #1321, opened 2025-06-22
[cpp] Use value types for asymmetric API

TartanLlama:
Changes the API style model. It used to be:

- `old`: uses rvalue reference types on guest exports for strings and lists
- `new`: uses view types (`string_view`, and `span`) on guest exports for strings and lists
- Default API style is `new` 
- API style is set by mutually exclusive `--new-api` and `--old-api` flags

This updates it to this state model:

- `asymmetric`: uses value types on guest exports for strings and lists
- `symmetric`: uses view types (`string_view`, and `span`) on guest exports for strings and lists
- Default API style is `asymmetric`
- API style is set by `--api-style` argument with `symmetric` or `asymmetric` as arguments

### DIFF
--- a/crates/cpp/DESIGN.md
+++ b/crates/cpp/DESIGN.md
@@ -46,9 +46,9 @@
 | HIA | v | string | | string_view | a,l |
 | HIR | t | string | | wit::string[^3] | &(a,l) |
 | HEA | t | string | | 🌓 wit::string[^4] | a,l |
-| | | | | 🪞 string_view [^6] | |
+| | | | | ⚖️ string_view [^6] | |
 | HER | p | string | | 🌓 wit::guest_owned<string_view> | 📘 -> &(a,l) |
-| | | | | 🪞 wit::string [^6] | 🪞 &(a,l) |
+| | | | | ⚖️ wit::string [^6] | 🪞 &(a,l) |
 
 [^1]: The host never frees memory (is never passed ownership)!
 


cpetig:
We discussed this today in depth on Zulip, I was unaware of the (in hindsight obvious) guarantee removing the copy even if you pass a string or vector by value (from a temporary (prvalue), or a move'd object (xvalue)).

Now Rust and C++ have matching semantics with respect to `&[T]` vs `Vec<T>` for arguments and results, when the `&&` qualifier always felt clumsy. So this canonical (asymmetric) API became default again, making the less efficient symmetric API a conscious choice.

crates/cpp/DESIGN.md
```diff
@@ -18,8 +18,8 @@
 
 | API | | | ABI | |
 | --- | --- | --- | --- | --- |
-| 🕸 | old | | 📘 | canonical |
-| 💎 | new | | 🪞 | symmetric |
+| 🌓 | asymmetric | | 📘 | canonical |
+| 🪞 | symmetric | | 🪞 | symmetric |
```
cpetig: I think having a separate symbol for symmetric API and symmetric ABI is still worth it. I could imagine 🌑 for ABI and 🌕 for API (or vice versa). ABI on the other hand clearly puts you into more distance with the wasm component model, so the dark moon is probably fitting.

crates/cpp/src/lib.rs
```diff
@@ -2672,9 +2710,7 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
                 }
 
                 let op0 = &operands[0];
-                let flavor = if self.gen.gen.opts.new_api
-                    && matches!(self.variant, AbiVariant::GuestImport)
-                {
+                let flavor = if matches!(self.variant, AbiVariant::GuestImport) {
```
cpetig, line 2713: This is a logic change for canonical. It could make sense.

tests/runtime/lists/test.cpp
```diff
@@ -98,30 +102,30 @@ wit::vector<wit::string> exports::test::lists::to_test::ListResult3() {
     return wit::vector<wit::string>::from_view(wit::span<wit::string>(std::vector<wit::string>{wit::string::from_view("hello,"), wit::string::from_view("world!")}));
 }
 
-wit::vector<uint8_t> exports::test::lists::to_test::ListRoundtrip(wit::span<const uint8_t> x) {
-    return wit::vector<uint8_t>::from_view(x);
+wit::vector<uint8_t> exports::test::lists::to_test::ListRoundtrip(wit::vector<uint8_t> x) {
+    return x;
```
cpetig, line 106: :heart:

tests/runtime/options/test.cpp
```diff
@@ -3,12 +3,11 @@
 #include <math.h>
 #include <test_cpp.h>
 
-template <class T>
-static bool equal(T const& a, T const& b) {
-    return a==b;
+static bool equal(std::optional<wit::string> const& a, std::optional<std::string_view> const& b) {
+    return a->get_view()==b.value();
```
cpetig: This assumes the non-optional case.

tests/runtime/options/test.cpp
```diff
@@ -17,17 +16,17 @@ std::optional<wit::string> exports::test::options::to_test::OptionNoneResult() {
...
     assert(equal(a, std::optional<std::string_view>("foo")));
 }
 
 std::optional<wit::string> exports::test::options::to_test::OptionSomeResult() {
     return std::optional<wit::string>(wit::string::from_view("foo"));
 }
 
-std::optional<wit::string> exports::test::options::to_test::OptionRoundtrip(std::optional<std::string_view> a) {
+std::optional<wit::string> exports::test::options::to_test::OptionRoundtrip(std::optional<wit::string> a) {
     if (!a.has_value()) return std::optional<wit::string>();
```
cpetig: If I an not mistaken it could just return a;

tests/runtime/strings/test.cpp
```diff
@@ -27,8 +24,7 @@ wit::string exports::test::strings::to_test::ReturnEmpty() {
   return wit::string((char const*)1, 0);
 }
 
-// new API: Identical for guest import and export
-wit::string exports::test::strings::to_test::Roundtrip(std::string_view str) {
+wit::string exports::test::strings::to_test::Roundtrip(wit::string str) {
   assert(str.size() > 0);
```
cpetig: This guard is no longer necessary with the new API as it doesn't use from_view

crates/cpp/src/lib.rs
TartanLlama: I think it's needed for things like `option<string>` so that it generates `std::string_view payload...` rather than `wit::string payload...` for imports, which would fail to compile

crates/cpp/DESIGN.md
TartanLlama: I switched to `⚖️` for API

crates/cpp/DESIGN.md
```diff
@@ -35,20 +35,20 @@
...
 | | | result<string,list> | Result<String, Vec> | std::expected<wit::string, wit::vector> | 📘 -> &(d,a,l) cabi_post |
 | --S | ? | string | String | wit::string | addr, len |
 | HIA | v | string | | string_view | a,l |
 | HIR | t | string | | wit::string[^3] | &(a,l) |
-| HEA | t | string | | 🕸 wit::string[^4] | a,l |
-| | | | | 💎 string_view [^6] | |
-| HER | p | string | | 🕸 wit::guest_owned<string_view> | 📘 -> &(a,l) |
-| | | | | 💎 wit::string [^6] | 🪞 &(a,l) |
+| HEA | t | string | | 🌓 wit::string[^4] | a,l |
+| | | | | 🪞 string_view [^6] | |
```
cpetig: mirror in but last column should be a scale (as well as the first on the next line)

cpetig (approved):
Minor symbol mismatch left, all other comments addressed.

TartanLlama:
Fixed!

alexcrichton (approved):

alexcrichton:
Thanks @TartanLlama! Also thanks  @cpetig for the review!

PR #1322, opened 2025-06-22
[cpp] Merge wit-common and wit-guest, install them when generating bindings

TartanLlama:
Currently, the C++ bindings have two headers that are required by the generated bindings in some cases. This PR merges these two headers into one, and installs them at generation time.

The required header is embedded into the binary, resulting in an increase of 8KB, or 0.04% of the binary size on release builds. I think that this will still be reasonable even if the header increases in size by an order of magnitude or so (which may well happen if we decide to ship an `expected` implementation).


crates/cpp/helper-types/wit-guest.h
```diff
@@ -9,7 +18,69 @@
 #include <stdlib.h> // free
 #include <new>
 
+
 namespace wit {
+#ifdef __cpp_lib_span
+using std::span;
+#else
+/// Minimal span (vector view) implementation for older C++ environments
+template <class T> class span {
```
cpetig: as we are targeting wasi-sdk by default C++20 should become the default and this is no longer necessary

crates/cpp/helper-types/wit-guest.h
```diff
@@ -9,7 +18,69 @@
...
+      resources.erase(iter);
+    }
+    return std::move(result);
+  }
+};
+
+/// @brief Replaces void in the error position of a result
+struct Void {};
+
+template<class To, class From>
```
cpetig: similarly this can be replaced by std::bit_cast (or at least import the std::bit_cast into the wit namespace).

crates/cpp/src/lib.rs
```diff
@@ -684,6 +684,13 @@ impl WorldGenerator for Cpp {
             .unwrap()
             .as_slice(),
         );
+
+        if self.dependencies.needs_wit {
+            files.push(
+                &format!("wit-guest.h"),
```
cpetig: I still believe this should no longer be called wit-guest -- as long as we don't plan a wit-host.

cpetig (commented):
I guess the publish failure could be related to not increasing the version number.

TartanLlama:
Thanks, I think these are all addressed now

crates/cpp/helper-types/wit-guest.h
TartanLlama: Done

crates/cpp/helper-types/wit-guest.h
TartanLlama: Done

crates/cpp/src/lib.rs
TartanLlama: Changed to `wit.h`

alexcrichton:
Deferring to @cpetig for review (if that's ok with you as well @cpetig, if you're busy I can also skim this).

Lucky1313:
Not sure if this is actually related to these changes, but when attempting to use the bindings from this branch, I wasn't able to get the component to link successfully without changing [this line](https://github.com/bytecodealliance/wit-bindgen/pull/1322/files#diff-db66db8663e3c8ea5889b5941928ac59942da2477e9853bf8e8127b09c232ad5L628) to use `extern "C"`, when compiling using clang++ 19 (the symbol was always not found). Is there another way to prevent this from being mangled? 

Is there somewhere we have a full invocation of the compiler showing what flags to use? For reference I was running with:
```-std=c++20 -fno-exceptions -Wall -Wextra -Wc++-compat -g --target=wasm32-wasip2 -O2 -Wl,--export-all -Wl,--no-entry```

**Comment by TartanLlama** [comment]
@Lucky1313 huh, it does look like that should be there, I'm not sure why it links for me and not you. I'll add that to this PR when I get a min

**Comment by TartanLlama** [comment]
@Lucky1313  Does compiling with the [WASI SDK](https://github.com/WebAssembly/wasi-sdk) make it link? On the subject of compiler flags, you can find the ones passed for the runtime tests [here](https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/test/src/cpp.rs#L123).

**Comment by TartanLlama** [comment]
Ah, the difference is the `-Wl,--export-all`; without it the `*_public_use_in_this_compilation_unit` function isn't kept around, so no undefined symbol in the linker

**Comment by Lucky1313** [comment]
I'm actually compiling using the WASI SDK [docker image](https://github.com/WebAssembly/wasi-sdk?tab=readme-ov-file#docker-image).

So it seems like it actually shouldn't be linking without that, because we're dropping the public use function without making sure it's used?

**Comment by TartanLlama** [comment]
Indeed, the extern "c" is definitely needed

**Comment by cpetig** [comment]
> Deferring to @cpetig for review (if that's ok with you as well @cpetig, if you're busy I can also skim this).

Yes, I am quite busy this week, but the small time needed to review this should be doable.

**Comment by cpetig** [comment]
The conflicts (can you please take a look @TartanLlama ?) and the linker symbol confuse me, though. My guess would be that the .c source file is compile as C++ to need extern "C" - I will need to take a closer look later this evening.

**Comment by TartanLlama** [comment]
@cpetig I've resolved the conflicts, so long as the tests pass and you're happy with everything, we should be good! 

On the `extern "C"` issue, that function gets generated into the `world_name.cpp` file, which gets compiled as C++, so the `__component_type_object_force_link_plugin` symbol gets mangled. The matching symbol in the `world_name_component_type.o` is not mangled, so linking fails as long as the `__component_type_object_force_link_plugin_public_use_in_this_compilation_unit` function is not discarded by the linker. Marking `__component_type_object_force_link_plugin` as `extern "C"` will disable mangling for it so that it links successfully, but the fact that the linker can just discard  `__component_type_object_force_link_plugin_public_use_in_this_compilation_unit` breaks the whole point of those functions. Perhaps we could also add `__attribute__((used))` to that function to prevent it from being discarded. This part affects the C bindings as well. I'll make a separate PR to address both of these.

**Comment by cpetig** [comment]
I took a closer look and the unification to std::span/bit_cast, the inclusion of wit.h and the removal of the ancient-C++ test cases in favor of making c++20 the default feels so much cleaner that I want this patch quickly merged even if I have to rework some long standing habits on my symmetric branch.

Short: Looks more than good to me, looking forward to see how it feels to use this in practice.

File: tests/runtime/cpp/param-ownership/runner-coarse-borrowing.cpp [hunk 4]
```diff
@@ -0,0 +1,43 @@
...
+  std::array<std::string_view, 2> a2 = {"value3", "value4"};
+  std::array<std::span<std::string_view const>, 2> as = {
+      std::span<std::string_view const>(a1.data(), a1.size()),
+      std::span<std::string_view const>(a2.data(), a2.size())};
+  std::span<std::span<std::string_view const> const> input(as.data(),
+                                                           as.size());
+  auto res = test::ownership::Foo(input);
+  assert(res.size() == 2);
+  assert(res[0].size() == 2);
+  assert(res[0][0].get_view() == "VALUE1");
```
cpetig, line 16: I guess this hand inlining is more readable than my previous equal overloading. I like it.

tests/runtime/demo/runner.cpp
```diff
@@ -1,7 +1,5 @@
-//@ args = '--language=Cpp'
-
-#include <runner.h>
+#include <runner_cpp.h>
 
 int main() {
-  a_b_the_test_x();
+  a::b::the_test::X();
```
cpetig, line 4: This also removes the icky feeling from this test. :heart:

tests/runtime/lists/test.cpp
```diff
@@ -12,28 +12,28 @@ static bool equal(T const&a, S const& b) {
...
 template<class R>
-static bool equal(wit::span<const R> const&a, std::vector<R> const& b) {
-    return equal(a, wit::span<R>(b));
+static bool equal(std::span<const R> const&a, std::vector<R> const& b) {
+    return equal(a, std::span<const R>(b));
 }
 template<class R>
 static bool equal(wit::vector<R> const&a, std::vector<R> const& b) {
-    return equal(a.get_view(), wit::span<R>(b));
+    return equal(a.get_view(), std::span<R>(b));
```
cpetig, line 36: more "std::", less language fragmentation (as long as you can use a new enough compiler ... I will ask people working on MISRA :wink:)

cpetig (approved):

PR #1335, opened 2025-07-04
Clarify host/guest language support

ShalokShalom:
This fixes the potential confusion in the wording of the README.

Before, it suggests that the languages supported by this repo are host languages, as they are named after the paragraph that speaks about host support.

It moves this section next to the guest support paragraph, to avoid any confusion.

I know that text itself said there is no host support here, but the placement made it ambiguous in my mind.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,13 @@ between bindings definitions.
 [component model]: https://github.com/WebAssembly/component-model
 
 The `wit-bindgen` repository is currently focused on **guest** programs which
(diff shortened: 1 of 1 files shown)


ShalokShalom:
I generally found it out of date, and posted something about that into the Discord channel. As an example, Java is not supported anymore. 

Also, C++ is a host now.

alexcrichton:
Indeed! I've updated some words in https://github.com/bytecodealliance/wit-bindgen/pull/1336

PR #1336, opened 2025-07-08
Update intro blurb about supported languages

alexcrichton:
Bringing some documentation up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ between bindings definitions.
 [component model]: https://github.com/WebAssembly/component-model
 
 The `wit-bindgen` repository is currently focused on **guest** programs which
-are those compiled to WebAssembly.  
-Languages developed in this repository are Rust, C, Java (TeaVM
-Java), Go (TinyGo), and C#.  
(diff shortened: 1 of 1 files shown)


ShalokShalom:
Thanks. The C++ implementation is host and guest compatible, as I think to recall from reading some commit messages?

Or do I remember that incorrectly? And do I need support for C++ host and TinyGo guest, to embed a TinyGo runtime into my C++ application?

The docs don't really seem to be clear now, when to use host, and when guest bindings. 

I mean, do host bindings provide bindings for all guest languages, or what are they for, please? 🤷🏻‍♂️ 

Thanks 🙂

alexcrichton:
Host bindings are a term used for when you're running components and the bindings necessary to run those components. Guest bindings are a term used for when you're building a component to run elsewhere on a host. The interop layer between a host and guest is WIT and there's no need to have any other sort of connection between hosts/guests in terms of languages.

ShalokShalom:
Sorry, I didnt quite get that. 

Use case wise: When do I need host bindings, please?

alexcrichton:
If you want to run components you need host bindings.

Issue #1332, opened 2025-07-02
[cpp] List in a variant generates invalid function binding


Lucky1313:
Given a WIT definition which includes a function returning a variant that includes a list:
```
package local:test@0.1.0;

interface task {
  variant value {
    a(bool),
    b(list<u8>),
  }
  
  do-thing: func() -> value;
}

world test {
  import task;
}
```

And generating C++17 bindings using latest master (`7f92598`):
```
wit-bingen cpp test.wit
```
`test.cpp`:
```
...snip
extern "C" __attribute__((import_module("local:test/task@0.1.0")))
__attribute__((import_name("do-thing")))
void localX3AtestX2FtaskX400X2E1X2E0X00do_thing(uint8_t *);
local::test::task::Value local::test::task::DoThing()
{
  uintptr_t ret_area[((3*sizeof(void*))+sizeof(uintptr_t)-1)/sizeof(uintptr_t)];
  uint8_t* ptr0 = (uint8_t*)(&ret_area);
  localX3AtestX2FtaskX400X2E1X2E0X00do_thing(ptr0);
  Value variant2;
  switch ((int32_t) (*((uint8_t*) (ptr0 + 0)))) {
    case 0: { 
      variant2.variants = Value::A{(bool((int32_t) (*((uint8_t*) (ptr0 + sizeof(void*))))))};
    } break;
    case 1: { auto len1 = *((size_t*) (ptr0 + (2*sizeof(void*))));

    variant2.variants = Value::B{wit::vector<uint8_t>((uint8_t*)(*((uint8_t**) (ptr0 + sizeof(void*)))), len1)};
  } break;
}
auto result3 = variant2;
return result3;
}
...snip
```
The bindings are unable to compile, I think due to the copy constructor for the variant that includes the `wit::vector` being deleted:

```
/opt/wasi-sdk/bin/clang++ -std=c++20 -fno-exceptions -Wall -Wextra -Wc++-compat -g --target=wasm32-wasip2 -O2 -I/bindings -c /bindings/test.cpp -o test.o
/bindings/test.cpp:44:6: error: call to implicitly-deleted copy constructor of 'Value'
   44 | auto result3 = variant2;
      |      ^         ~~~~~~~~
/bindings/test_cpp.h:11:22: note: copy constructor of 'Value' is implicitly deleted because field 'variants' has a deleted copy constructor
   11 |   std::variant<A, B> variants;
...snip
```

[...]

PR #1338, opened 2025-07-15
[cpp] Add move_if_necessary in C++ binding flush instruction

Lucky1313:
This addresses https://github.com/bytecodealliance/wit-bindgen/issues/1332.

Not completely sure this is the right way to do it, would appreciate a double check @cpetig or @TartanLlama when you get the chance.

Also not sure where I would add a test for this, some guidance would be helpful.

### DIFF
--- a/crates/cpp/src/lib.rs
+++ b/crates/cpp/src/lib.rs
@@ -3063,7 +3063,7 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
                 for i in operands.iter().take(*amt) {
                     let tmp = self.tmp();
                     let result = format!("result{}", tmp);
-                    uwriteln!(self.src, "auto {result} = {};", i);
+                    uwriteln!(self.src, "auto {result} = {};", move_if_necessary(i));
                     results.push(result);
                 }
             }
--- a/crates/cpp/src/lib.rs
+++ b/crates/cpp/src/lib.rs
@@ -2639,7 +2639,7 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
                         let case = format!("{elem_ns}::{}", case.name.to_pascal_case());
                         uwriteln!(
                             self.src,
-                            "const {} &{} = std::get<{case}>({}.variants).value;",
+                            "{} &{} = std::get<{case}>({}.variants).value;",
                             ty,
                             payload,
                             operands[0],
--- a/tests/runtime/cpp/variant-with-data/runner.cpp
+++ b/tests/runtime/cpp/variant-with-data/runner.cpp
@@ -1,0 +1,55 @@
+#include <assert.h>
+#include <runner_cpp.h>
+
+template <class T>
+static bool equal(T const& a, T const& b) {
+    return a == b;
+}
+
+template <class T>
+static bool equal(wit::vector<T> const& a, wit::vector<T> const& b) {
+    if (a.size() != b.size()) return false;
+    for (size_t i = 0; i < a.size(); i++) {
+        if (!equal(a[i], b[i])) return false;
+    }
+    return true;
+}
+
+static bool equal(wit::string const& a, wit::string const& b) {
+    return a.get_view() == b.get_view();
+}
+
+static bool equal(::test::variant_with_data::to_test::DataVariant const& a, ::test::variant_with_data::to_test::DataVariant const& b) {
+    if (a.variants.index() != b.variants.index()) return false;
+    switch (a.variants.index()) {
+        case 0: return equal(std::get<::test::variant_with_data::to_test::DataVariant::Bytes>(a.variants).value, std::get<::test::variant_with_data::to_test::DataVariant::Bytes>(b.variants).value);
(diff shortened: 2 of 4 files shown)


TartanLlama:
Seems reasonable! I'd add a test to the `tests/runtime/cpp` folder. You can see, for instance, this small test for an example of how they look: https://github.com/bytecodealliance/wit-bindgen/tree/main/tests/runtime/c/rename

cpetig:
I think https://github.com/bytecodealliance/wit-bindgen/tree/main/tests/runtime/numbers is a good template for a test (look at the cpp files).

Lucky1313:
I put together a test that fails before with the expected error, and succeeds after.

In writing the test, I discovered another issue related to the exported bindings, which was used in the test.

When an exported function returns a variant that contains data, it creates the payload for the variant that's `const`, but then when it tries to leak that value, the `leak` function can't run on a `const` variable:
```cpp
extern "C" __attribute__((__export_name__("test:variant-with-data/to-test#get-data")))
uint8_t * testX3Avariant_with_dataX2Fto_testX23get_data(int32_t arg0)
{
  auto result0 = exports::test::variant_with_data::to_test::GetData((uint8_t(arg0)));
  static uintptr_t ret_area[((3*sizeof(void*))+sizeof(uintptr_t)-1)/sizeof(uintptr_t)];
  uint8_t* ptr1 = (uint8_t*)(&ret_area);
  switch ((int32_t) (result0).variants.index()) {
    case 0: {
      const wit::vector<uint8_t> &payload2 = std::get<test::variant_with_data::to_test::DataVariant::Bytes>(result0.variants).value;
      *((int8_t*)(ptr1 + 0)) = (int32_t(0));
      auto const&vec3 = payload2;
      auto ptr3 = (uint8_t*)(vec3.data());
      auto len3 = (size_t)(vec3.size());
      payload2.leak();
...snip
```

```
error: 'this' argument to member function 'leak' has type 'const wit::vector<uint8_t>' (aka 'const vector<unsigned char>'), but function is not marked const
           39 |       payload2.leak();
              |       ^~~~~~~~
```

I added another [commit](https://github.com/bytecodealliance/wit-bindgen/pull/1338/commits/4440264ece93ba4b70d8db3cb9aa35d4c06c00f3) to remove this `const` in the bindings, but not sure if this is the right thing to do unconditionally.

crates/cpp/src/lib.rs
```diff
@@ -2639,7 +2639,7 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
                         let case = format!("{elem_ns}::{}", case.name.to_pascal_case());
                         uwriteln!(
                             self.src,
-                            "const {} &{} = std::get<{case}>({}.variants).value;",
+                            "{} &{} = std::get<{case}>({}.variants).value;",
```
cpetig, line 2642: I agree that it feels wrong to remove this const unconditionally but with leak (which is necessary for correct ownership) it can't be const. I will probably give it a try to reintroduce const later.

tests/runtime/cpp/variant-with-data/runner.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+static bool equal(wit::string const& a, wit::string const& b) {
+    return a.get_view() == b.get_view();
+}
+
+static bool equal(::test::variant_with_data::to_test::DataVariant const& a, ::test::variant_with_data::to_test::DataVariant const& b) {
+    if (a.variants.index() != b.variants.index()) return false;
+    switch (a.variants.index()) {
+        case 0: return equal(std::get<::test::variant_with_data::to_test::DataVariant::Bytes>(a.variants).value, std::get<::test::variant_with_data::to_test::DataVariant::Bytes>(b.variants).value);
+        case 1: return equal(std::get<::test::variant_with_data::to_test::DataVariant::Number>(a.variants).value, std::get<::test::variant_with_data::to_test::DataVariant::Number>(b.variants).value);
```
cpetig, line 26: Nothing preventing this from merging but I realize that long term we will want to add a `using namespace ::test::variant_with_data::to_test;` to keep this function more readable.

tests/runtime/cpp/variant-with-data/test.cpp
```diff
@@ -0,0 +1,23 @@
...
+      variant.variants = DataVariant::Bytes(wit::vector<uint8_t>::from_view(std::span<uint8_t>(bytes)));
+    }
+    case 1:
+      variant.variants = DataVariant::Number(42);
+    case 2:
+      variant.variants = DataVariant::Text(wit::string::from_view("hello"));
+    default:
+      variant.variants = DataVariant::Number(0);
+  }
+  auto result = std::move(variant);
```
cpetig, line 21: seems unnecessary, directly returning variant should work as well?

tests/runtime/cpp/variant-with-data/runner.cpp
Lucky1313: I just followed the examples of the other tests in the repo, which all seemed to have things fully expanded where they were used.

tests/runtime/cpp/variant-with-data/test.cpp
Lucky1313: Turns out it does, I had just used the bindings as an example of the export function since I needed it for the test, and copied exactly how they do it haha

PR #1331, opened 2025-06-30
Test the C++ generator in CI

alexcrichton:
Updating CI configuration to ensure that `cpp` is listed under the languages tested.

PR #1340, opened 2025-07-20
Test the C++ generator in CI (2nd version: with more fixes included and more upstream changes)

cpetig:
Extends https://github.com/bytecodealliance/wit-bindgen/pull/1331 by Alex

@TartanLlama would love to hear from you about these changes as well as you wrote the original test case.

There are still 26 codegen tests failing:
 - [ ] `import_export_func.wit`: confuses function foo and namespace foo :cry: 
 - [ ] `multiversion`: we don't support versioned namespaces, yet
 - [ ] `variants-unioning-types.wit`: I guess this is import/export type incompatibility as these types are part of a world and not an interface
 - [ ] `resource-borrow-in-record.wit`: we don't include `<span>` here
 - [ ] 20 times TBD


cpetig:
- [ ] `lists.wit`: We receive a const argument and create non-const references in lowering.
- [ ] `async-trait-function.wit`: `todo!()` in `crates/cpp/src/lib.rs:1795`
- [ ] `resource-alias.wit`: Likely still incomplete resource support
- [ ] `futures.wit`, `streams.wit`: `todo!()` in `crates/cpp/src/lib.rs:1579`
- [ ] `ret-areas.wit`: Multiple ret-areas? The first should be an argument buffer (rename).
- [ ] `keywords.wit`: Escape some more names
- [ ] `simple-http.wit`: Oh, we need `optional` to construct an `expected` even outside of autosar types (wrapping expected in optional gives you an unconditional default constructor)
- [ ] `variants.wit`: string_view/span vs string/vector confusion and expected 

I guess with this amount of significant effort I will introduce an "expected codegen failure" feature and mask them for now.

cpetig:
I just had to deploy the already existing infrastructure for expected codegen test failures, still fixing these some day would be nice :wink:

cpetig:
This should be ready to merge, as it successfully turns on CI testing of C++, but the remaining expected test failures require more work.

alexcrichton (approved):

PR #1341, opened 2025-07-21
c/lib.rs: dtor symbols should use name, not snake

james-rms:
I noticed this discrepancy between the symbols generated by the rust and C binding generators - this change should (hopefully) make it so that rust and C/C++ implementers can build components that are compatible with eachother.

### DIFF
--- a/crates/c/src/lib.rs
+++ b/crates/c/src/lib.rs
@@ -1382,7 +1382,7 @@ extern int32_t __wasm_import_{ns}_{snake}_rep(int32_t);
     return ({ns}_{snake}_t*) __wasm_import_{ns}_{snake}_rep(handle.__handle);
 }}
 
-__attribute__((__export_name__("{module}#[dtor]{snake}")))
+__attribute__((__export_name__("{module}#[dtor]{name}")))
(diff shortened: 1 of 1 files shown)


cpetig:
I guess it would be possible to extend the runtime tests to cover this error (fail before, pass after) by implementing a resource in C which has a more complex name.

james-rms:
I'm not really familiar enough with the test framework to make that change. Note that this bug case only comes up when we generate bindings using C, and try to consume the resulting component with bindings generated in some other language.

alexcrichton:
Tests are located [here](https://github.com/bytecodealliance/wit-bindgen/tree/main/tests/runtime) with a basic readme located [here](https://github.com/bytecodealliance/wit-bindgen/tree/main/tests) and you can find some C-specific tests [here](https://github.com/bytecodealliance/wit-bindgen/tree/main/tests/runtime/c). With those bits would you be up for adding a test?

james-rms:
I took a look at that for a few hours and couldn't figure out how to add a test that compares the output of two different generators. I'm not going to put more work into this patch, feel free to close or merge as needed.

PR #1346, opened 2025-08-02
[C#] Simple Async support

yowl:
This PR add the very minimal changes to enable the first async functions tests to pass.  These tests return immediately which means a lot of the callback, waitable, changes are not done yet.  There are no TaskCompletionSources created, we just set the return value on the task.  User code will look like

https://github.com/bytecodealliance/wit-bindgen/pull/1346/files#diff-f34015927bd91d8f86e20e1a9258891caad407de6a499fcf977cb1165c89b3a5

and 

https://github.com/bytecodealliance/wit-bindgen/pull/1346/files#diff-756d40996fd1204ae5975dd4da40e815b827d804c27b13dd36f69b6612ced825

Which should be idiomatic.

Thanks @dicej  for the help in Zulip.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,8 +69,9 @@ jobs:
     - uses: actions/checkout@v4
       with:
         submodules: true
+    # FIXME(rust-lang/rust#148347) current rust (1.91) is broken
     - name: Install Rust
-      run: rustup update stable --no-self-update && rustup default stable
+      run: rustup update 1.90.0 --no-self-update && rustup default 1.90.0
     - run: rustup target add wasm32-wasip1 wasm32-wasip2
 
     - run: rustup target add wasm32-unknown-unknown


crates/csharp/src/function.rs
```diff
@@ -1027,6 +1114,24 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
+                    uwriteln!(self.src, r#"if (ret.IsCompletedSuccessfully) 
+                    {{
+                        {name}TaskReturn({ret_param});
+                        return (uint)CallbackCode.Exit;
+                    }}
+                    
+                    return (uint)CallbackCode.Yield;
+                    "#);
+                }
+
```
dicej, line 1159: Might want to add a "TODO" comment here to remember to defer dropping borrowed resources until a result is returned (i.e. just before calling `{name}TaskReturn`).

crates/csharp/src/interface.rs
```diff
@@ -496,10 +586,40 @@ impl InterfaceGenerator<'_> {
...
+            [global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "[callback][async-lift]{import_module_name}#{wasm_func_name}")]
+            public static uint {camel_name}Callback(uint eventRaw, uint waitable, uint code)
+            {{
+                // TODO: decode the parameters
+                return (uint)CallbackCode.Exit;
+            }}
+            "#
+            );
+
+            let task_return_param = match &sig.results[..] {
```
dicej, line 605: The task return function could accept up to 16 core parameters (e.g. 2 for a `string`, 5 for a `tuple<string, list<u8>, u32>`, etc.).

dicej (approved):
Looks like a solid start, thanks!

crates/csharp/src/function.rs
```diff
@@ -956,11 +1003,24 @@ impl Bindgen for FunctionBindgen<'_, '_> {
             }
 
             Instruction::CallWasm { sig, .. } => {
+                let is_async = InterfaceGenerator::is_async(self.kind);
+
+                let requires_async_return_buffer_param = is_async && sig.results.len() >= 1;
+                let async_return_buffer = if requires_async_return_buffer_param {
+                    let buffer = self.emit_allocation_for_type(&sig.results);
+                    uwriteln!(self.src, "//TODO: store somewhere with the TaskCompletionSource.");
```
pavelsavara: I think you can use `new TaskCompletionSource(object? state)` -> `Task.AsyncState`
Probably with some (internal?) helper struct/class to hold allocations and handles that you need to cleanup.

pavelsavara:
Could you please create and link a gist with the generated code ? Thank you.

yowl:
> Could you please create and link a gist with the generated code ? Thank you.

https://gist.github.com/yowl/72b406c9604b5ff4c170e441fb3e88c8

Has the code gen and copies of the harness code from this repo.

crates/csharp/src/function.rs
yowl: Thanks, done

crates/csharp/src/interface.rs
yowl: I'm hoping I'll be forced to address all these shortcuts as I go through enabling the tests.  If not, then I've added a TODO here and I can address the todo's at the end.

crates/csharp/src/function.rs
yowl: Thanks, I've updated the comment here, as per the remark above I'm hoping to be forced to address all these as I go through enabling the tests.

yowl:
Have reverted the code gen interop output as that is a bigger change than can be snuck in here.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -0,0 +1,9 @@
+/**
+ * Helpers for the async support.
+ */
+
+ public enum CallbackCode
```
jsturtevant, line 5: Looks like another type would eventually want to pull out to make them sharable.

jsturtevant (commented):
Looks like a good start to me.  Thanks for getting this started! I assume eventually there will need to be some changes to the runtime to support Async?

yowl:
> Looks like a good start to me. Thanks for getting this started! I assume eventually there will need to be some changes to the runtime to support Async?

I don't think so, we should have everything in place.  As I flesh out more tests we will see, but haven't seen anything yet, either here on in the futures PR that requires a change.

crates/csharp/src/AsyncSupport.cs
yowl: Yes, there is some cleanup/sharing for size that we should do at some point.

PR #1350, opened 2025-08-13
Migrate this workspace to using trusted publishing

alexcrichton:
Similar to bytecodealliance/wasm-tools#2281 but for this repository. The main benefit is removal of a long-lived token, and the main consequence is that crates will need to be published to crates.io with a placeholder before being merged in here.


alexcrichton:
I plan on publishing a new release after this to ensure all the automation works.

pchickey (approved):

PR #1357, opened 2025-08-20
C# future simple codegen

yowl:
This PR adds enough code gen to support the simple-future wit runtime test.  As for the async PR, this is pretty much the minimum PR that I could do to pass the test.  I've not tackled the typed canonical methods except to add a "void" implementation which is hard coded as the one to use.

Have followed the c test cases rather than the rust ones.

WIP as behind https://github.com/bytecodealliance/wit-bindgen/pull/1346


yowl:
cc @pavelsavara @dicej @jsturtevant

dicej:
Thanks for doing this, @yowl!  I'm planning to review it by the end of the week.

crates/csharp/src/function.rs
```diff
@@ -1375,15 +1401,27 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
             Instruction::AsyncTaskReturn { name: _, params: _ } => {
                 uwriteln!(self.src, "// TODO_task_cancel.forget();");
             }
 
-            Instruction::FutureLift { .. }
-            | Instruction::StreamLower { .. }
+            Instruction::FutureLift { payload: _, ty: _ } => {
+                // TODO get the prefix for the type
+                let sig_type_name = "Void";
+                uwriteln!(self.src, "var reader = new {}.FutureReader{}({});", self.interface_gen.name, sig_type_name, operands[0]);
```
dicej: I'd recommend using e.g. `self.locals.tmp("reader")` here to generate a unique variable name so as to avoid clashes.

crates/csharp/src/function.rs
```diff
@@ -1375,15 +1401,27 @@ impl Bindgen for FunctionBindgen<'_, '_> {
 
             Instruction::FutureLower { .. } => {
                 let op = &operands[0];
+                self.interface_gen.add_future(self.func_name);
+
                 results.push(format!("{op}.Handle"));
```
dicej: `FutureReader` should probably have a `TakeHandle()` method that zeros out the handle field (and asserts that it wasn't already zero) before returning the original value so that the application won't accidentally try to use the no-longer-valid handle.

crates/csharp/src/interface.rs
```diff
@@ -168,14 +196,211 @@ impl InterfaceGenerator<'_> {
...
+        );
+
+        uwrite!(
+            self.csharp_interop_src,
+            r#"
+                [global::System.Runtime.InteropServices.DllImportAttribute("$root", EntryPoint = "[waitable-set-drop]"), global::System.Runtime.InteropServices.WasmImportLinkageAttribute]
+                internal static unsafe extern void WaitableSetDrop(int waitable);
+                "#
+        );
+
```
dicej: These could be moved to a standalone e.g. `AsyncSupport.cs` file that gets copied into the output since they don't depend on the WIT in any way.

crates/csharp/src/interface.rs
```diff
@@ -168,14 +196,211 @@ impl InterfaceGenerator<'_> {
...
+        let future_type_name = "Void";
+
+        self.csharp_gen
+            .interface_fragments
+            .entry(self.name.to_string())
+            .or_insert_with(|| InterfaceTypeAndFragments::new(false))
+            .interface_fragments
+            .push(InterfaceFragment {
+                csharp_src: format!(r#"
+                    public class FutureReader{future_type_name} : FutureReader
```
dicej: Have you considered making `FutureReader` and `FutureWriter` generic classes rather than generating separate classes for each payload type?  That's what we do in Rust and in [Go](https://github.com/bytecodealliance/wit-bindgen/pull/1447), for example.  Note how we use "vtable"-style structure in both Rust and Go to specialize the implementations internally.  Happy to discuss further if this isn't clear.

crates/csharp/src/interface.rs
```diff
@@ -787,9 +1028,9 @@ impl InterfaceGenerator<'_> {
                             .map(|ty| self.type_name_with_qualifier(ty, qualifier))
                             .unwrap_or_else(|| "".to_owned());
                         if name.is_empty() {
-                            return "Task".to_owned();
+                            return "FutureReader".to_owned();
                         } else {
-                            return format!("Task<{name}>");
+                            return format!("FutureReader<{name}>");
```
dicej: This suggests that `FutureReader` is generic, but it isn't, is it?

crates/csharp/src/interface.rs
```diff
@@ -168,14 +196,211 @@ impl InterfaceGenerator<'_> {
...
+        for future in &self.futures {
+            let camel_name = future.to_upper_camel_case();
+
+            //TODO: we need these per future type.
+            // Create a hash map..
+            uwrite!(
+                self.csharp_interop_src,
+                r#"
+                [global::System.Runtime.InteropServices.DllImportAttribute("{import_module_name}", EntryPoint = "[future-new-0][async]{future}"), global::System.Runtime.InteropServices.WasmImportLinkageAttribute]
+                internal static extern ulong {camel_name}VoidNew();
```
dicej: Using `camel_name` (which is derived from the name of an arbitrary function in which this future type happened to appear) is confusing IMO.

It's true that we use that function name as part of the Wasm import name because that's the only practical way to uniquely communicate to `wit-component` which future type we're referring to.  However, that's purely an implementation detail that shouldn't be exposed to users.  Instead, we should generate names like `FutureNew` for a `future`, `FutureStringNew` for a `future<string>`, etc.  And if the same future type shows up in more than one function, we should only emit one copy.  See the Rust, C, Go, and/or [Python](https://github.com/bytecodealliance/componentize-py) generators for examples.

dicej (commented):
Thanks for doing this!  Looks like a good start; please see my comments inline.

crates/csharp/src/interface.rs
yowl: Thanks, done.

crates/csharp/src/interface.rs
yowl: Correct, removed the generic.

crates/csharp/src/function.rs
yowl: RIght, done.

crates/csharp/src/function.rs
yowl: I've done this, but not sure I understand in what situation it would be used.

crates/csharp/src/interface.rs
yowl: Done.[](url)

crates/csharp/src/interface.rs
yowl: Is there a test that exercises the generic types with more than one concrete implementation?

yowl: nevermind, I've found what I was missing

crates/csharp/src/function.rs
dicej: Sorry, I meant not only that we should add a `FutureReader.TakeHandle()` method but that we should use it here (instead of just reading the `Handle` field) for the `Instruction::FutureLower { .. }` case.

crates/csharp/src/interface.rs
yowl: Latest commit has the generic versions of `FutureReader` and `FutureWriter` with a vtable concept.  In c# we dont have generics with no type, so we need an extra version for that.

yowl:
@dicej Hi, I think I've covered the points from the review now, was a bit of a change for the generic refactoring, sorry about that churn.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -2,8 +2,300 @@
...
+        [global::System.Runtime.InteropServices.DllImportAttribute("$root", EntryPoint = "[waitable-set-wait]"), global::System.Runtime.InteropServices.WasmImportLinkageAttribute]
+        internal static unsafe extern int WaitableSetWait(int waitable, int* waitableHandlePtr);
+
+        [global::System.Runtime.InteropServices.DllImportAttribute("$root", EntryPoint = "[waitable-set-drop]"), global::System.Runtime.InteropServices.WasmImportLinkageAttribute]
+        internal static unsafe extern void WaitableSetDrop(int waitable);
+    }
+
+    public static WaitableSet WaitableSetNew() 
+    {{
+        var waitable = Interop.WaitableSetNew();
```
dicej: Nit: consider naming this `waitableSet` or `handle`.  The name `waitable` is a bit confusing since that term means something else in the component model.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -2,8 +2,300 @@
...
+        }
+        var handle = Handle;
+        Handle = 0;
+        return handle;
+    }
+
+    // TODO: Generate per type for this instrinsic.
+    public Task Read()
+    {
+        // TODO: Generate for the interop name and the namespace.
```
dicej, line 133: Should check here to see if `Handle` is zero and throw an exception if so.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -2,8 +2,300 @@
...
+            return Task.CompletedTask;
+        }
+
+        throw new NotImplementedException();
+    }
+
+    void Dispose(bool _disposing)
+    {
+        // Free unmanaged resources if any.
+        vTable.DropReader(Handle);
```
dicej: Should only call this if `Handle != 0`

crates/csharp/src/AsyncSupport.cs
```diff
@@ -2,8 +2,300 @@
...
+        }
+        var handle = Handle;
+        Handle = 0;
+        return handle;
+    }
+
+    // TODO: Generate per type for this instrinsic.
+    public Task Read()
+    {
+        // TODO: Generate for the interop name and the namespace.
```
dicej, line 195: Should throw exception if `Handle == 0` here.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -2,8 +2,300 @@
...
+            return Task.CompletedTask;
+        }
+
+        throw new NotImplementedException();
+    }
+
+    void Dispose(bool _disposing)
+    {
+        // Free unmanaged resources if any.
+        vTable.DropReader(Handle);
```
dicej: Should only call this if `Handle != 0`

crates/csharp/src/AsyncSupport.cs
```diff
@@ -2,8 +2,300 @@
...
+
+public class FutureWriter<T>(int handle, FutureVTable vTable) // : TODO Waitable
+{
+    public int Handle { get; } = handle;
+    public FutureVTable VTable { get; private set; } = vTable;
+
+    // TODO: Generate per type for this instrinsic.
+    public Task Write()
+    {
+        // TODO: Generate for the interop name.
```
dicej, line 295: Should throw exception if `Handle == 0` here.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -2,8 +2,300 @@
...
+            return tcs.Task;
+        }
+
+        throw new NotImplementedException();
+    }
+
+    void Dispose(bool _disposing)
+    {
+        // Free unmanaged resources if any.
+        VTable.DropWriter(Handle);
```
dicej: Should only call this if `Handle != 0`.

crates/csharp/src/function.rs
```diff
@@ -166,7 +168,7 @@ impl<'a, 'b> FunctionBindgen<'a, 'b> {
             .drain(self.blocks.len() - cases.len()..)
             .collect::<Vec<_>>();
         let ty = self.interface_gen.type_name_with_qualifier(ty, true);
-        //let ty = self.gen.type_name(ty);
+        //let ty = self.r#gentype_name(ty);
```
dicej:
```suggestion
        //let ty = self.interface_gen.type_name(ty);
```

Or can we just get rid of this line?

crates/csharp/src/function.rs
dicej:
```suggestion
                results.push(format!("{op}.TakeHandle()"));
```

crates/csharp/src/world_generator.rs
```diff
@@ -87,6 +94,50 @@ impl CSharp {
...
+
+    // We can share some types to save some code size and allow a more intuitive user experience.
+    pub(crate) fn type_is_bidirectional(resolve: &Resolve, ty: &TypeId) -> bool {
+        let type_def = &resolve.types[*ty];
+        let kind = &type_def.kind;
+
+        return match kind {
+            TypeDefKind::Flags(_) => true,
+            TypeDefKind::Enum(_) => true,
+            _ => false,
```
dicej, line 106: If desired, you can return `true` here for any type that doesn't refer transitively to an exported resource.  For Go, we use [this function](https://github.com/bytecodealliance/wit-bindgen/blob/85369ad5a2cf45bf4ac11d65e68aeba303e7e0e5/crates/go/src/lib.rs#L1403-L1414) to determine whether the type can be shared or not.

dicej (commented):
Thanks for continuing to iterate on this, @yowl, and sorry for the delayed review.

Looks good overall; just a few more comments inline.

crates/csharp/src/world_generator.rs
yowl: Thanks for the tip, I had a go at this, no pun intended, but I have a question.  Types are added from the `import_types` method, and that is where the resources are added in C#, is that where Go adds the resources as well?  I ask because the `has_exported_resource`` function checks for the resource being in the export direction.

dicej: My recollection is that `import_types` is used for types imported at the world level, but that interface-level types are added in `import_interface` and `export_interface`, respectively:

https://github.com/bytecodealliance/wit-bindgen/blob/85369ad5a2cf45bf4ac11d65e68aeba303e7e0e5/crates/go/src/lib.rs#L680-L689
https://github.com/bytecodealliance/wit-bindgen/blob/85369ad5a2cf45bf4ac11d65e68aeba303e7e0e5/crates/go/src/lib.rs#L730-L751

Note the logic that sorts types according to `has_exported_resource`.

Anyway, don't worry about this for now; this PR is big enough as it is, so we can follow up with type sharing tweaks later.

yowl: ok, it was this wit that I had problem with, but will leave it for now.
```
package test:test;

world foo {
  resource x;

  export return-resource: func() -> x;
}
```

yowl:
I think that's all the latest comments addressed.  Thanks for the help!

crates/csharp/src/AsyncSupport.cs
dicej: Looks like we still need a `Handle == 0` check here.

yowl:
Thanks you!

PR #1358, opened 2025-08-20
Prioritize `inline` worlds, not external ones

alexcrichton:
When multiple `path`s are used the `generate!` macro will fail if they all contain a `world` and no world is otherwise specified. This then additionally affected when using `inline` plus loading auxiliary WIT from a `paths` directory (either explicitly or the default `wit`). This commit switches the logic to prioritizing worlds in the `inline` block because if that's being specified it's pretty likely the exact one that wants to be used.

### DIFF
--- a/crates/guest-rust/macro/src/lib.rs
+++ b/crates/guest-rust/macro/src/lib.rs
@@ -266,6 +266,7 @@ fn parse_source(
                     }
                 }
             }
+            pkgs.truncate(0);
             pkgs.push(resolve.push_group(UnresolvedPackageGroup::parse("macro-input", s)?)?);
         }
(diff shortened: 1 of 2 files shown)


pchickey (approved):

alexcrichton:
cc @bytecodealliance/dotnet-core do y'all know what's going on with the C# failures in this PR? I've retried a few times over a day or so and so it doesn't seem spurious

yowl:
Looks all green, where can I see the failures?

alexcrichton:
This is failing when going through the merge queue, so you'll need to hit "View details" buttons on the right. An example is https://github.com/bytecodealliance/wit-bindgen/actions/runs/17137088638/job/48615613853

PR #1360, opened 2025-08-22
Support fallible constructors

badeend:
This implements https://github.com/WebAssembly/component-model/pull/550 for the Rust generator. C worked out of the box without additional changes. Support for other language generators needs more thought (see below).

What's the best way to proceed with this PR from here on?
Get all languages working first? I'll need help with that then.
Or merge with this initial set of languages, and leave the rest for future PRs?

### C++
AFAICS, there isn't any exception throwing/handling support in the cpp generator today and everything uses `std::expected`. I have near-zero experience with C++, though I've heard exception handling is a bit of a contentious topic in the C++ community. So I suppose this was done for a reason?

If the cpp generator wants to continue using C++'s native constructor syntax it'll require exceptions.
Alternatively, it'll need to switch to a static factory method. But, at least for _exported_ resources, there already exists a static `New` method. Not sure why this is needed when there's already a constructor too.

### C#
The csharp generator has exception support already. But it also has a `with_wit_results` mode. What should happen in that case?
- Ignore the flag and generate a throwing constructor anyway? Not very intuitive.
- Leave the native constructor syntax behind and generate a static factory method?
- Generate both versions? Similar to the common [`Parse`](https://learn.microsoft.com/en-us/dotnet/api/system.int32.parse?view=net-9.0)+[`TryParse`](https://learn.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=net-9.0) pattern.

Maybe worth noting: [properties](https://github.com/WebAssembly/component-model/issues/235) will likely end up in the same boat.

### Moonbit
No experience with this language.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -816,7 +816,7 @@ name = "test-helpers"
 version = "0.0.0"
 dependencies = [
  "codegen-macro",
- "wasm-encoder 0.236.0",
+ "wasm-encoder 0.237.0",
  "wit-bindgen-core",
  "wit-component",
  "wit-parser",
@@ -999,9 +999,9 @@ checksum = "ddbd7f2a9e3635abe5d4df93b12cadc8d6818079785ee4fab3719ae3c85a064e"
 
 [[package]]
 name = "wasm-compose"
-version = "0.236.0"
+version = "0.237.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "98b5290a0aca685aab16c936f682b85e8e4e3a0bfe1843afd43372eb82e34f47"
+checksum = "4691ea85473e81ea13f91088854186eb792710d7c1af8b07bc1ad1ab3a02404f"
 dependencies = [
  "anyhow",
  "heck 0.4.1",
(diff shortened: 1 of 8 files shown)


alexcrichton:
It's fine to leave this test as not-working on some generators, no need to take on implementing this everywhere yourself. Flagging the test as failing in `crates/test/*` is fine for some generators and just having this implemented in others is ok

badeend:
Ah, alright. I'll exclude them for now then.

@cpetig & @jsturtevant: going off git blame I see you're the primary authors of the cpp & csharp generators. Could you take a look at the top post in this PR?

Edit: looks like CI  is broken for C# in general. Other PRs have the same error. [Example1 ](https://github.com/bytecodealliance/wit-bindgen/actions/runs/17137088638), [example 2](https://github.com/bytecodealliance/wit-bindgen/actions/runs/17148890830/job/48650486975)

yowl:
I'll look at the C# problem.

yowl:
One possibility is the update to wasi-sdk-27.  Do you want to just disable C# in the CI for now as it will take me a few days to look into this.

alexcrichton:
Ok in that case I'll disable C# for now and file an issue, I don't trust myself to update wasi-sdk-27 alas

PR #1364, opened 2025-08-22
fix(msrv): add rust-version based on ground truth

billf:
use of `iter::repeat_n` (and `Cargo.lock` version 4?) cause these
packages to depend on Rust 1.82.0 at minimum


pchickey (approved):
Looks good. Thank you for adding this

PR #1368, opened 2025-08-26
Support lists of WIT directories on the CLI.

sunfishcode:
Accept multiple WIT paths in the wit-bindgen CLI, using the same logic as the multiple WIT paths accepted by the `path` parameter in the Rust bindgen macro.

Also, document that list-of-WIT interfaces, in the Rust macro and now the wit-bindgen CLI, have a topological ordering requirement which may be lifted in the future.

### DIFF
--- a/src/bin/wit-bindgen.rs
+++ b/src/bin/wit-bindgen.rs
@@ -85,14 +85,18 @@ struct Common {
     #[clap(long = "out-dir")]
     out_dir: Option<PathBuf>,
 
-    /// Location of WIT file(s) to generate bindings for.
+    /// Locations of WIT file(s) to generate bindings for.
     ///
-    /// This path can be either a directory containing `*.wit` files, a `*.wit`
-    /// file itself, or a `*.wasm` file which is a wasm-encoded WIT package.
-    /// Most of the time it's likely to be a directory containing `*.wit` files
-    /// with an optional `deps` folder inside of it.
+    /// These paths can be either directories containing `*.wit` files, `*.wit`
+    /// files themselves, or `*.wasm` files which are wasm-encoded WIT packages.
+    /// Most of the time they're likely to be directories containing `*.wit`
+    /// files with optional `deps` folders inside of them.
+    ///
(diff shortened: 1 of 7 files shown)


src/bin/wit-bindgen.rs
```diff
@@ -219,8 +219,12 @@ fn gen_world(
             resolve.features.insert(feature.to_string());
         }
     }
-    let (pkg, _files) = resolve.push_path(&opts.wit)?;
-    let world = resolve.select_world(pkg, opts.world.as_deref())?;
+    let mut main_packages = Vec::new();
+    for wit in &opts.wit {
+        let (pkg, _files) = resolve.push_path(wit)?;
+        main_packages.push(pkg);
+    }
```
alexcrichton, line 230: Ah this reminds me one of the reasons I found this tricky to do before. The ideal implementation in my head is such that there's no order dependency of arguments on the CLI and the implementation internally would figure out how to slurp up all the WITs. As-implemented here, however, while later arguments could refer to WITs of earlier arguments it's not possible the other way around due to the explicit order of `push_path`.

Internally `Resolve` has everything necessary to do all the topo-sorting and everything, but it's not exposed currently through a `push_many` support or anything like that.

This is probably fine to have as a TODO, however, as this equally affects the `paths` argument in the Rust macro. Perhaps worth documenting though in the meantime and having an issue on wasm-tools to fill out at some point?

alexcrichton:
I think you should be able to fix most CI by switching to a git dep instead of using `[patch]` (the test framework makes its own workspace effectively), but that'll probably cause the publish-related steps to fail.

Would you like to have a release of wasm-tools to keep on the path of integrating this?

src/bin/wit-bindgen.rs
sunfishcode: Ah. In my current use cases, this ordering requirement isn't a problem, but I agree we should think about lifting this requirement. I've now added documentation to this PR, and filed https://github.com/bytecodealliance/wasm-tools/issues/2292.

alexcrichton (approved):

PR #1369, opened 2025-08-26
Refactor the `wit-bindgen` crate for being suitable again to include with the Rust standard library

alexcrichton:
Over time with WASIp2 and WASIp3 the `wit-bindgen` crate has become unsuitable for inclusion in libstd. The goal of this commit is to fix all of these issues in favor of an eventual update of Rust's `wasm32-wasip2` target to use the `wasi` crate directly. The high-level changes here are:

* Support for a different `wit-bindgen`-runtime-crate and `wit-bindgen`-macro-crate is removed. This means old compat hacks are removed and additionally `wit-bindgen-rt` is merged back into `wit-bindgen`.
* Support for disabling `bitflags`-the-dep is added by adding a dummy `bitflags!` definition which is used when the crate dependency is disabled. The bare-bones version gives just enough for bindings generation and basic usability.

This is tested with libstd and it at least gets to the point of building libstd itself when the `wasi` crate is regenerated with these changes.


sunfishcode:
Noting this here as I just looked it up: bitflags used to support being built as a dependency of std, but it appears they recently [removed that code](https://github.com/bitflags/bitflags/pull/454). I think the minimal version here is fine for now, though in the future if this lands in upstream Rust, we may want to ask bitflags to re-introduce that code.

alexcrichton:
Oh good find! I can't really blame them though as it's weird esoteric code for one user, and for this it'd be for just one user of one platform. Given that this only affects the Rust standard library's use of the `wasi` crate itself I don't think it's too worthwhile to have more than the basics for `bitflags!` -- basically I'd be ok never asking bitflags to re-add the code. That being said I certainly wouldn't stop anyone as it's a weird thing here for sure to carry.

PR #1373, opened 2025-09-03
[C#] Change class to readonly struct for records

andreakarasho:
By design the `record` wit type is just a set of fields (https://component-model.bytecodealliance.org/design/wit.html#records) This PR aims to reduce the GC work changing from a `class` to a `readonly struct`.

### DIFF
--- a/crates/csharp/src/interface.rs
+++ b/crates/csharp/src/interface.rs
@@ -995,7 +995,7 @@ impl<'a> CoreInterfaceGenerator<'a> for InterfaceGenerator<'a> {
         uwrite!(
             self.src,
             "
-            {access} class {name} {{
+            {access} readonly struct {name} {{
                 {fields}
 
                 {access} {name}({parameters}) {{


GerardSmit:
Additional: the design document states the following

> Records are similar to a `struct` in many languages.
> https://github.com/WebAssembly/component-model/blob/2c17516179f99accdafb01d8e4affcc0f58184cc/design/mvp/WIT.md?plain=1#L1514-L1516

It's also not possible to recursively reference a record. `wit-bindgen` already checks for this. I tried to reference `point` in `point`:
> ```
>    0: failed to parse package: wit
>    1: type `point` depends on itself
> ``` 
> https://github.com/bytecodealliance/wasm-tools/blob/544ac54d0d9d43bcfed2597af4293729dfb89854/crates/wit-parser/src/ast/toposort.rs#L169-L171

yowl:
Thanks, I need to fix the CI first.

PR #1381, opened 2025-09-06
fix(async): rust-rt async depends on std

billf:
presumably after #1369, the following build feature combo fails to build:

```bash
cargo build -p wit-bindgen --no-default-features --features async
```


billf:
a more... aggressive.. version of sits in a local branch as https://github.com/bytecodealliance/wit-bindgen/commit/223acef1a08433c668f5f78a656a73b9ff7f74b8 but, having not educated myself on the `rustc` side of `rustc-dep-of-std` builds, i did not send it.

alexcrichton (approved):

alexcrichton:
Eventually we'll want to add no_std support to the async bits, but that's ok to defer to later. That'll I think require some careful handling to integrate into libstd

PR #1382, opened 2025-09-06
chore(nostd): no-op eprintln! earlier in pipeline

billf:


alexcrichton (commented):
Thanks! Could you expand on the rationale though? This is an intentional design to avoid dead code warnings when this cfg is disabled

billf:
> Thanks! Could you expand on the rationale though? This is an intentional design to avoid dead code warnings when this cfg is disabled

i realize that `default-features = false` is a warranty-breaking maneuver, but `feature="async"` now depends indirectly depends on `feature="std"`. this upgrade, for instance:

```diff
-wit-bindgen = { version = "0.44", default-features = false, features = [
+wit-bindgen = { version = "0.45", default-features = false, features = [
   "macros",
 ]}
```
i may have misconstrued the intent behind the `if false`. beyond developer printf, i thought it was avoiding pulling in a macro (`std::eprintln!`) that may not exist in the guest environment. 

the fix for `no_std+async` is more complex (as noted elsewhere), so i was deferring it separate from this caller in case it escaped containment from async into rt at large.

there's value in keeping code inside the`rtdebug` macro verifiably compilabile and this change may be too aggressive
e.g. `rtdebug!(this::must::exist())`

alexcrichton:
Ah ok makes sense! It's expected that `async` depends on `std` for now, and I think it's ok to lump this in with fixing that eventually too

billf:
I will bring this out of draft with the simpler, reasonable fix and open an issue to track & discuss the larger `no_std` issue.

alexcrichton:
Personally I'd prefer to defer handling this to handling no_std entirely all at once rather than piecemeal. I want to ask, though, do you have a use case for no_std? The assumption so far is that no_std only matters for this crate being depeneded on by the Rust standard library itself, but I'm assuming you have a use case that's otherwise different

billf:
> piecemeal

#1381 is as far as I plan to take it for now.

> use case

professionally, I am building components out of the `wasm32-unknown-unknown` (or `wasm32v1-none`) targets and `wit_component::ComponentEncoder`. they use WIT for well-defined component-to-component API. the host gives guests no access to WASI. `no_std` prevents pulling in any WASI-based dependencies at the codegen stage.

non-professionally, i'm curious about the compiler pipeline of using external dependencies from within a language's (target's) stdlib. i can take that curiosity offline/zulip/etc or onto an issue documenting what `no_std` would look like here.

alexcrichton:
Ah ok makes sense! This is orthogonal to this PR at this point, but it may be worth pointing out though that there's some subtle mismatches between what you're doing an your goals. For example `#![no_std]` does not prevent you from doing `extern crate std`, nor does it prevent other crates from doing `extern crate std`. Basically so long as a target has `std` there's no way to guarantee it's not linked. Furthermore there's nothing preventing code from doing `extern crate wasi` and using WASI APIs even on `wasm32-unknown-unknown`. Mostly it's just conventions/social norms for these two points.

If your goal is to avoid using WASI APIs I'd recommend using `wasm32-unknown-unknown` and using `std`. It'll make things much easier to use `std` and that target won't have any use of WASI from libstd, and most crates conventionally won't use WASI on that target too.

For using async bindings from Rust's `std` crate itself, I think we're going to want to indeed use no_std for that but we're also going to want to get rid of most of the `async`, for example binding everything as a synchronous function and supporting synchronous stream reads/writes and future reads/writes. Effectively libstd won't be using the nonblocking support AFAIK, although wasi-libc will to emulate non-blocking I/O.

billf:
thanks for the explanation. will do any follow up on zulip or an issue.

Issue #1377, opened 2025-09-05
C#: parameters are leaking


GerardSmit:
Assume the following WIT:
```wit
package tests:component@0.1.0;

world test {
    export accept-string: func(s: string);
}
```
<details>
  <summary>Click here for the generated code</summary>

```cs
// Generated by `wit-bindgen` 0.41.0. DO NOT EDIT!
// <auto-generated />
#nullable enable

namespace TestWorld {

    public interface ITestWorld {
        static abstract void AcceptString(string s);

    }

    namespace exports {
        using System.Text;
        using System.Runtime.InteropServices;
        public static class TestWorld
        {

            [UnmanagedCallersOnly(EntryPoint = "accept-string")]
            public static unsafe void wasmExportAcceptString(nint p0, int p1) {

                TestWorldImpl.AcceptString((Encoding.UTF8.GetString((byte*)p0, p1)));

            }

        }
    }

}
```
</details>

In the exported `[UnmanagedCallersOnly]`-method, it creates a new string from the UTF-8 string pointer:
```csharp
TestWorldImpl.AcceptString((Encoding.UTF8.GetString((byte*)p0, p1)));
```
This string is allocated in a new memory block (and managed by the GC). So, the parameter pointer is not used anymore.
However, the pointer is also not being freed.

If I look at the C wit-bindgen docs:

https://github.com/bytecodealliance/wit-bindgen/blob/ca55f12b7cab252797203d0508473b1d93e1d139/crates/c/README.md?plain=1#L39-L42

Here is stated that the guest _must_ free the parameter.

Because this is not being done in the C# wit-bindgen, the parameter memory is kept and thus is building up after each call. After _a lot_ of calls, you will run into an out-of-memory exception. This is what I ran into in https://github.com/dotnet/runtimelab/issues/3161.

There is no way for the user to clean this memory, since the string (created by Encoding.UTF8) is not pointed to the parameter pointer.

I think the generate code is missing `NativeMemory.Free` after the string is created.

PR #1389, opened 2025-09-20
[C#] Fixes mem leak: https://github.com/bytecodealliance/wit-bindgen/issues/1377

andreakarasho:
The PR aims to fix https://github.com/bytecodealliance/wit-bindgen/issues/1377
Beside that other languages still leaking because of missing `free` calls in this case.

wit file:
```wit
package tests:component@0.1.0;

world test {
    export accept-string: func(s: string, b: string);
}
````

Generated code:

```csharp
// Generated by `wit-bindgen` 0.46.0. DO NOT EDIT!
// <auto-generated />
#nullable enable

namespace TestWorld {

    public interface ITestWorld {
        static abstract void AcceptString(string s, string b);

    }

    namespace exports {
        public static class TestWorld
        {

            [global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "accept-string")]
            public static unsafe void wasmExportAcceptString(nint p0, int p1, nint p2, int p3) {

                var str = global::System.Text.Encoding.UTF8.GetString((byte*)p0, p1);
                global::System.Runtime.InteropServices.NativeMemory.Free((void*)p0);

                var str0 = global::System.Text.Encoding.UTF8.GetString((byte*)p2, p3);
                global::System.Runtime.InteropServices.NativeMemory.Free((void*)p2);

                TestWorldImpl.AcceptString((str), (str0));

            }

        }
    }

}

```

### DIFF
--- a/crates/csharp/src/function.rs
+++ b/crates/csharp/src/function.rs
@@ -838,10 +838,23 @@ impl Bindgen for FunctionBindgen<'_, '_> {
             }
 
             Instruction::StringLift { .. } => {
-                results.push(format!(
+                let str = self.locals.tmp("str");
+                let op0 = &operands[0];
+
+                let get_str = format!(
                     "global::System.Text.Encoding.UTF8.GetString((byte*){}, {})",
-                    operands[0], operands[1]
-                ));
+                    op0, operands[1]
+                );
+
+                uwriteln!(
+                    self.src,
+                    "
+                    var {str} = {get_str};
+                    global::System.Runtime.InteropServices.NativeMemory.Free((void*){op0});
+                    "
+                );
+
+                results.push(str);
             }
 
             Instruction::ListLower { element, realloc } => {


andreakarasho:
So the final solution is to check for `size > 0`.
Marked the PR ready.

crates/csharp/src/function.rs
```diff
@@ -838,10 +838,26 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
-                ));
+                    op0, op1
+                );
+
+                uwriteln!(
+                    self.src,
+                    "
+                    var {str} = {get_str};
+                    if ({op1} > 0) {{
+                        global::System.Runtime.InteropServices.NativeMemory.Free((void*){op0});
```
pavelsavara, line 859: Should `Instruction::ListLift` also free ? Which other types pass memory ownership ?

andreakarasho: From what I can see a `list<string>` generates the following code:

```csharp
[global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "accept-string")]
public static unsafe void wasmExportAcceptString(nint p0, int p1) {

    var array = new global::System.Collections.Generic.List<string>(p1);
    for (int index = 0; index < p1; ++index) {
        nint basePtr = p0 + (index * 8);

        var str = global::System.Text.Encoding.UTF8.GetString((byte*)global::System.BitConverter.ToInt32(new global::System.Span<byte>((void*)(basePtr + 0), 4)), global::System.BitConverter.ToInt32(new global::System.Span<byte>((void*)(basePtr + 4), 4)));
        if (global::System.BitConverter.ToInt32(new global::System.Span<byte>((void*)(basePtr + 4), 4)) > 0) {
            global::System.Runtime.InteropServices.NativeMemory.Free((void*)global::System.BitConverter.ToInt32(new global::System.Span<byte>((void*)(basePtr + 0), 4)));
        }

        array.Add(str);
    }
    TestWorldImpl.AcceptString((array));

}
```  

I suppose a `free` is needed here and in `Instruction::ListCanonLift` case too.

andreakarasho:
Covered cases `ListCanonLift` and `ListLift` too

pavelsavara:
LGMT, @yowl , @jsturtevant I would leave the approval with one of you

PR #1393, opened 2025-09-29
feat: async support in MoonBit

peter-jerry-ye:
@BigOrangeQWQ has been working hard on adding async support in MoonBit. Right now we have a working implementation, except for the issue in [wasm-tools](https://github.com/bytecodealliance/wasm-tools/issues/2301) which will make the encoding incorrect when passing strings.


peter-jerry-ye:
It is ready since it's working. There are some other improvements and refactors planned but I might put them in some later PRs.

PR #1406, opened 2025-10-30
extend the lifetime of borrowed resource `get`

chenyan2002:
The original get function elides the lifetime, so we have `fn get(&'b self) ->  &'b T`. But since `Borrow<'a>` is valid for the whole `'a` duration, we should be able to borrow `&'a T` instead. The new signature is also semantically more accurate.

### DIFF
--- a/tests/runtime/resource-borrow/test.rs
+++ b/tests/runtime/resource-borrow/test.rs
@@ -11,7 +11,7 @@ pub struct MyThing {
 }
 
 fn get_val<'a>(v: &ThingBorrow<'a>) -> &'a u32 {
-  v.get::<MyThing>().val
+  &v.get::<MyThing>().val
 }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! For this I'd prefer to have one of (a) leave them private and users use the `&`-returning methods, or (b) make them public but document them and explain why they might be used. I don't mind either one, but I'd prefer to not be in the current state where they're public-but-hidden as that feels a bit too brittle to me (e.g. easy to forget why they were public in the first place)

chenyan2002:
I have a better fix now. The original `get` function elides the lifetime, so we have `fn get(&'b self) ->  &'b T`. But since `Borrow<'a>` is valid for the whole `'a` duration, we should be able to borrow `&'a T` instead. The new signature is also semantically more accurate.

PR #1420, opened 2025-11-14
fix(cpp): escape cpp keywords

ricochet:
Fixes #1419

Signed-off-by: Bailey Hayes <behayes2@gmail.com>

PR #1424, opened 2025-11-19
variant case

ricochet:
Builds on #1420 and must be merged after.

Adds __wasm_export_ prefix to all C ABI export function names
    
When a WIT world exports a function whose name matches the world/package
name (e.g., world "foo" with export function "foo"), the generated C ABI
function name foo() conflicted with the C++ namespace namespace foo.

This generates functions like:
```c
extern "C" __attribute__((__export_name__("foo")))
uint8_t * __wasm_export_foo()  // C++ function name
```

The public C++ API is `exports::foo::Foo()`.
Users never see the __wasm_export_foo() function in the bindings.

### DIFF
--- a/crates/cpp/src/lib.rs
+++ b/crates/cpp/src/lib.rs
@@ -1338,7 +1338,8 @@ impl CppInterfaceGenerator<'_> {
                     f.needs_dealloc = needs_dealloc;
                     f.cabi_post = None;
                     abi::call(f.gen.resolve, variant, lift_lower, func, &mut f, false);
-                    let code = String::from(f.src);
+                    let ret_area_decl = f.emit_ret_area_if_needed();
+                    let code = format!("{}{}", ret_area_decl, String::from(f.src));
                     self.gen.c_src.src.push_str(&code);
                 }
             }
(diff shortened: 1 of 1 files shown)


crates/cpp/src/lib.rs
```diff
@@ -3190,11 +3221,11 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
         };
         uwriteln!(
             self.src,
-            "{static_var}{tp} ret_area[({size_string}+sizeof({tp})-1)/sizeof({tp})];"
+            "{static_var}{tp} ret_area{tmp}[({size_string}+sizeof({tp})-1)/sizeof({tp})];"
```
cpetig: I am just curious where individual ret_area instances are necessary, normally the superset of alignment and size should be sufficient, shouldn't it?

ricochet: I can see where c and rust do a single allocation here. I was unnecessarily paranoid about using something shared with co-operative threads on the horizon. Since the pointers are used sequentially and never simultaneously, it should be fine. Thank you for catching.

PR #1425, opened 2025-11-19
cpp resources

ricochet:
Depends on #1424

This adds resource support for wit-bindgen cpp + clippy cleanups across the generator.

I think the remaining missing feature is covering multiversion. The rest all appear to be async wasip3.

### DIFF
--- a/crates/test/src/cpp.rs
+++ b/crates/test/src/cpp.rs
@@ -49,16 +49,9 @@ impl LanguageMethods for Cpp {
             "async-trait-function.wit"
             | "error-context.wit"
             | "futures.wit"
-            | "lists.wit"
             | "multiversion"
-            | "resource-alias.wit"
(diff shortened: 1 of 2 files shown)


cpetig:
Looks very good, I see a lot of code quality improvements and unifications (which even out the historically grown code base with newer additions like "move_if_necessary", and as usual the clippy warnings are significant quality improvements)

crates/test/src/cpp.rs
```diff
@@ -49,27 +49,9 @@ impl LanguageMethods for Cpp {
             "async-trait-function.wit"
             | "error-context.wit"
             | "futures.wit"
-            | "import_export_func.wit"
-            | "import-func.wit"
-            | "issue573.wit"
-            | "issue929-no-export.wit"
```
cpetig: This file alone shows the quality improvement by this patch ;-)

PR #1436, opened 2025-11-25
Simplify extern definitions for intrinsics

alexcrichton:
Deduplicate host panicking shim definitions with the wasm definitions.

### DIFF
--- a/crates/guest-rust/src/rt/async_support.rs
+++ b/crates/guest-rust/src/rt/async_support.rs
@@ -21,7 +21,40 @@ macro_rules! rtdebug {
             std::eprintln!($($f)*);
         }
     }
+}
 
+/// Helper macro to deduplicate foreign definitions of wasm functions.
+///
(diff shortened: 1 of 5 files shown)


crates/guest-rust/src/rt/async_support.rs
```diff
@@ -21,7 +21,40 @@ macro_rules! rtdebug {
...
         }
     }
+}
 
+/// Helper macro to deduplicate foreign definitions of wasm functions.
+///
+/// This automatically imports when on wasm targets and then defines a dummy
+/// panicking shim for native targets to support native compilation but fail at
+/// runtime.
+macro_rules! extern_wasm {
```
dicej, line 31: Might want to move this to the crate root and make it public some day since it's generally useful.

alexcrichton: Agreed yeah, it'd be useful even for the `generate!` macro to use this I think

Issue #1431, opened 2025-11-24
rust: make `FutureWaker` component-model-task-aware


dicej:
Currently `wit_bindgen::rt::async_support` contains a `FutureWaker` type used to determine whether to return `CALLBACK_CODE_YIELD` or `CALLBACK_CODE_WAIT` when the `Future` corresponding to a given task returns `Poll::Pending`.  However, it ought to be a bit smarter with regards to inter-task communication.

Consider a scenario involving two concurrent tasks: *A* and *B*:

1. *A* creates a oneshot channel and stores the write end in a global variable, then `await`s the read end, returning `CALLBACK_CODE_WAIT` to the host
1. *B* writes to the write end, waking *A*'s waker

In this case, waking *A*'s waker should have the effect of telling the host to resume *A*; currently it does not.  I imagine the waker could create a new payload-less `future`, call `future.read` on the read end (which would return `RETURN_CODE_BLOCKED`), add the read end to *A*'s `waitable-set`, then call `future.write` on the write end, causing the host to eventually call the callback for *A*.  Alternatively, we could either proactively or lazily create a payload-less `stream` for each task and reuse it as needed.  Once we have cooperative multithreading support, we could simplify that to just calling e.g. `thread.switch-to` -- no `future` or `stream` needed.

PR #1440, opened 2025-11-26
rust: Add support for inter-task wakeups

alexcrichton:
This commit fixes a preexisting issue in the async support in the rust
bindings where inter-component-model-task wakeups/communication did not
work. Previously `wit-bindgen` would end up either panicking or wedging
itself if this situation arose, effectively not handling it at all.

Inter-task wakeups are supported in the component model with unit
streams and futures (e.g. `future` and `stream` with no type parameter).
For these types the component model allows the same instance to both
read/write from the stream or future. The support here is implemented
with `stream` as a task might receive many wake up notifications over
its lifetime.

Implementing this is a bit tricky, so the highlights of the
implementation are:

* This relies on new "magic function imports" recognized by
  `wit-component` specifically related to unit-streams and unit-futures.
  This is a very new feature of `wit-component` and thus hasn't had time
  to propagate throughout the ecosystem yet. Due to the niche nature of
  needing cross-task wakeups this commit disables cross-task wakeups by
  default and adds support for it behind a Cargo feature. The thinking
  is that if anyone runs into this there's now a clear panic/error
  message indicating that the feature can be enabled. Otherwise the hope
  is that no one runs into this (as it's relatively niche) and when the
  time is right we can enable this by default.

* All tasks started by wit-bindgen will need this new inter-task-wakeup
  stream for the reason that we don't actually know what will wake a
  task up when it blocks. To avoid adding a stream allocation on the
  fast path, the allocation of the inter-task stream is deferred until
  a task blocks for the first time.

* Right now a read of the inter-task stream starts when a task blocks.
  The read is cancelled when the task wakes up. A future optimization
  would be to keep the read-in-progress and only restart it if
  necessary. For now that's deferred, however.

Closes https://github.com/bytecodealliance/wit-bindgen/issues/1431

### DIFF
--- a/crates/guest-rust/src/rt/async_support.rs
+++ b/crates/guest-rust/src/rt/async_support.rs
@@ -198,7 +198,7 @@ impl FutureState<'_> {
                 .sleep_state
                 .store(SLEEP_STATE_WOKEN, Ordering::Relaxed);
 
-            // With all of our context now configured, deliver the evnet
+            // With all of our context now configured, deliver the event
             // notification this callback corresponds to.
             //
             // Note that this should happen under the reset of
--- a/crates/guest-rust/src/rt/async_support/inter_task_wakeup.rs
+++ b/crates/guest-rust/src/rt/async_support/inter_task_wakeup.rs
@@ -1,6 +1,7 @@
 use super::FutureState;
 use crate::rt::async_support::{BLOCKED, COMPLETED};
 use crate::{RawStreamReader, RawStreamWriter, StreamOps, UnitStreamOps};
+use std::ptr;
 use std::sync::Mutex;
 
 #[derive(Default)]
@@ -40,7 +41,7 @@ impl FutureState<'_> {
         if !self.inter_task_wakeup.stream_reading {
             let stream = self.inter_task_wakeup.stream.as_mut().unwrap();
             let handle = stream.handle();
-            let rc = unsafe { UnitStreamOps.start_read(handle, 1 as *mut _, 1) };
+            let rc = unsafe { UnitStreamOps.start_read(handle, ptr::null_mut(), 1) };
             assert_eq!(rc, BLOCKED);
             self.inter_task_wakeup.stream_reading = true;
             self.add_waitable(handle);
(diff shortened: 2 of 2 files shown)


alexcrichton:
This is temporarily built on https://github.com/bytecodealliance/wit-bindgen/pull/1439 and ~~relies on unpublished versions of wasm-tools so CI will be broken~~.

crates/guest-rust/src/rt/async_support/inter_task_wakeup.rs
```diff
@@ -0,0 +1,97 @@
...
+        // If there's not already a pending read then schedule a new read here.
+        //
+        // Note that this should always return `BLOCKED` since as the only task
+        // running it's not possible for a read to be anywhere else in the
+        // system. Additionally we keep the read end alive, so this shouldn't
+        // ever returned dropped/closed either.
+        if !self.inter_task_wakeup.stream_reading {
+            let stream = self.inter_task_wakeup.stream.as_mut().unwrap();
+            let handle = stream.handle();
+            let rc = unsafe { UnitStreamOps.start_read(handle, 1 as *mut _, 1) };
```
dicej: What's the reason for using `1 as *mut _` here (and below) vs. `ptr::null_mut()`?

crates/guest-rust/src/rt/async_support.rs
```diff
@@ -175,37 +190,40 @@ impl FutureState<'_> {
...
-        self.poll()
-    }
+        self.with_p3_task_set(|me| {
+            // Transition our sleep state to ensure that the inter-task stream
+            // isn't used since there's no need to use that here.
+            me.waker
+                .sleep_state
+                .store(SLEEP_STATE_WOKEN, Ordering::Relaxed);
+
+            // With all of our context now configured, deliver the evnet
```
dicej:
```suggestion
            // With all of our context now configured, deliver the event
```

crates/guest-rust/src/rt/async_support/inter_task_wakeup.rs
alexcrichton: Just my own paranoia, I'll switch to null

PR #1442, opened 2025-11-26
Support custom Rust types for `list`

### LINKED ISSUE #1441: rust: Support custom `list` types
It would be very helpful for me to be able to use a custom type for a `list<T>` type, but this crate seems to assume that `list<T>` always maps directly to `Vec<T>`.

Ex:

```
wit_bindgen::generate!({
    inline: r#"
        package foo:bar;

        interface baz {
            type bytes = list<u8>;
            read: func() -> bytes;
        }

        world bzt {
            use baz.{bytes};
        }
    "#,
    with: {
        "foo:bar/baz/bytes": crate::Bytes,
    }
});

struct Bytes(Vec<u8>);
```

```
error[E0308]: mismatched types
  --> ~/Projects/wit-bindgen-custom-list-type/target/debug/build/wit-bindgen-rust-macro-4f3b8d44afc291c2/out/bzt0.rs:43:21
   |
12 |             pub fn read() -> crate::Bytes {
   |                              ------------ expected `Bytes` because of return type
...
43 |                     result5
   |                     ^^^^^^^ expected `Bytes`, found `Vec<_>`
   |
   = note: expected struct `Bytes`
              found struct `Vec<_>`
help: try wrapping the expression in `Bytes`
   |
43 |                     Bytes(result5)
   |                     ++++++       +
```

My usecase is a wasm component that runs Lua scripts where I want to be able pass wit-bindgen's generated structs directly to `mlua` using its support of `serde::{Serialize, Deserialize}`. This works great for everything except `list<u8>` which would most logically map to a lua string (which hold arbitrary binary data), but it is being pulled in as an array like any other `Vec<T>` would. The way to solve this on the `mlua` side is to wrap it in a newtype with a different `IntoLua` implementation. I would really like to avoid having to have a full set of parallel types and conversions for everything that uses my `type bytes = list<u8>;`.

I think a reasonable way to implement this would be to just require that the type implements `Into<Vec<T>>`, `From<Vec<T>>`, and `Deref<Target = T>` insert some `.into()`s if the list type has a type defined in the `with` section. Though I haven't figured out how well that would fit into how wit-bindgen currently works. Thoughts on if that is a good approach or if there's a better option? I'm happy to do the implementation if this is something you'd accept.

Here's a quick hack just to prove to myself that this could reasonably work: https://github.com/azdle/wit-bindgen/commit/4a5b8d11db0850586516d0d1b9039bed92f9e78c

azdle:
Any type that implements `From<Vec<T>>`, `Into<Vec<T>>` ~~and `Deref<Target = Vec<T>>`~~ can be used for a WIT type `list<T>` via the `with` option.

---

This was simpler than I originally expected, by omitting the type names, I can just let the compiler fall back to calling `<Vec<T> as Into<Vec<T>>::into` when no custom type is used, allowing the codegen to not need to track the custom type.

Resolves #1441

### DIFF
--- a/crates/rust/Cargo.toml
+++ b/crates/rust/Cargo.toml
@@ -34,6 +34,7 @@ wit-bindgen = { path = '../guest-rust', features = ['async'] }
 test-helpers = { path = '../test-helpers' }
 # For use with the custom attributes test
 serde_json = "1"
+bytes = "1"
 
 [features]
 serde = ['dep:serde', 'wit-bindgen-core/serde']
--- a/crates/rust/src/bindgen.rs
+++ b/crates/rust/src/bindgen.rs
@@ -670,11 +670,15 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 let val = format!("vec{}", tmp);
                 let ptr = format!("ptr{}", tmp);
                 let len = format!("len{}", tmp);
+                let vec = self.r#gen.path_to_vec();
                 if realloc.is_none() {
                     self.push_str(&format!("let {} = {};\n", val, operands[0]));
                 } else {
                     let op0 = operands.pop().unwrap();
-                    self.push_str(&format!("let {} = ({}).into_boxed_slice();\n", val, op0));
+                    self.push_str(&format!(
+                        "let {} = <_ as Into<{vec}<_>>>::into({}).into_boxed_slice();\n",
+                        val, op0
+                    ));
                 }
                 self.push_str(&format!("let {} = {}.as_ptr().cast::<u8>();\n", ptr, val));
                 self.push_str(&format!("let {} = {}.len();\n", len, val));
@@ -691,7 +695,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 self.push_str(&format!("let {} = {};\n", len, operands[1]));
                 let vec = self.r#gen.path_to_vec();
                 let result = format!(
-                    "{vec}::from_raw_parts({}.cast(), {1}, {1})",
+                    "<_ as From<{vec}<_>>>::from({vec}::from_raw_parts({}.cast(), {1}, {1}))",
                     operands[0], len
                 );
                 results.push(result);
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -432,7 +432,8 @@ impl RustWasm {
             }
 
             self.src.push_str("mod _rt {\n");
-            self.src.push_str("#![allow(dead_code, clippy::all)]\n");
+            self.src
+                .push_str("#![allow(dead_code, unused_imports, clippy::all)]\n");
             let mut emitted = IndexSet::new();
             while !self.rt_module.is_empty() {
                 for item in mem::take(&mut self.rt_module) {
--- a/crates/rust/tests/codegen.rs
+++ b/crates/rust/tests/codegen.rs
@@ -31,3 +31,140 @@ mod inline_and_path {
         generate_all,
     });
 }
+
+#[allow(unused)]
+mod newtyped_list {
+    use std::ops::Deref;
+
+    wit_bindgen::generate!({
+        inline: r#"
+        package test:newtyped-list;
+
+        interface byte {
+            type typed-list-of-byte = list<u8>;
+            type newtyped-list-of-byte = list<u8>;
+
+            record rec-of-lists {
+                l: list<u8>,
+                tl: typed-list-of-byte,
+                nl: newtyped-list-of-byte,
+            }
+
+            use-list-of-byte: func(l: list<u8>) -> list<u8>;
+            use-typed-list-of-byte: func(tl: typed-list-of-byte) -> typed-list-of-byte;
+            use-newtyped-list-of-byte: func(nl: newtyped-list-of-byte) -> newtyped-list-of-byte;
+            use-rec-of-lists: func(t: rec-of-lists) -> rec-of-lists;
+        }
+
+        interface noncopy-byte {
+            // this will be new-typed by a non-copy struct
+            type noncopy-byte = u8;
+
+            type newtyped-list-of-noncopy-byte = list<noncopy-byte>;
+            type typed-list-of-noncopy-byte = list<noncopy-byte>;
+
+            record rec-of-lists-of-noncopy-byte {
+                ntl: newtyped-list-of-noncopy-byte,
+                tl: typed-list-of-noncopy-byte,
+                l: list<noncopy-byte>,
+            }
+
+            use-newtyped-list-of-noncopy-byte: func(nl: newtyped-list-of-noncopy-byte) -> newtyped-list-of-noncopy-byte;
+            use-typed-list-of-noncopy-byte: func(tl: typed-list-of-noncopy-byte) -> typed-list-of-noncopy-byte;
+            use-list-of-noncopy-byte: func(l: list<noncopy-byte>) -> list<noncopy-byte>;
+            use-rec-of-lists-of-noncopy-byte: func(t: rec-of-lists-of-noncopy-byte) -> rec-of-lists-of-noncopy-byte;
+        }
+
+        world test {
+            import byte;
+            export byte;
+            import noncopy-byte;
+            export noncopy-byte;
+        }
+        "#,
+        with: {
+            "test:newtyped-list/byte/newtyped-list-of-byte": crate::newtyped_list::NewtypedListOfByte,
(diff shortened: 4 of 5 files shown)


crates/rust/src/lib.rs
```diff
@@ -432,7 +432,8 @@ impl RustWasm {
             }
 
             self.src.push_str("mod _rt {\n");
-            self.src.push_str("#![allow(dead_code, clippy::all)]\n");
+            self.src
+                .push_str("#![allow(dead_code, unused_imports, clippy::all)]\n");
```
azdle, line 436: I added an `allow(unused_imports)` here.

This applies to this module in the generated code:

```rust
mod _rt {
    #![allow(dead_code, clippy::all, unused_imports)]
    pub use alloc_crate::vec::Vec;
    extern crate alloc as alloc_crate;
}
```

I'm pretty sure this should be okay, I just wanted to call this out specifically because it was failing with an "unused import" error on a test called "unused-import.wit", but I'm fairly certain those are talking about different unused imports.

azdle:
Status (mostly as a reminder to myself):

* I ended up removing the impl for `Deref<Target = Vec<T>>` and the test passes, but I do actually need that impl in my real application. The test probably needs to test more, I'm just not sure exactly what yet.
* When I don't have the deref impl the errors the compiler spits out are not the most helpful, I want to see if I can find some way to require that trait explicitly to get better errors.

azdle:
> but sending a Bytes to the outside world would unnecessarily require a copy since the shared-reference of Bytes would be copied into a uniquely owned Vec<u8>

I'm not sure what shared reference you're talking about here.

Just to make sure we're on the same page, the `Bytes` type was talking about in the issue was `struct Bytes(Vec<u8>);` in the first code block, not a `Bytes` from the bytes crate. ~~I didn't use the bytes crate because exactly for the reason that it would cause a fully copy of the data.~~ [edit: Actually, this may not be true.]

azdle:
> I didn't use the bytes crate because exactly for the reason that it would cause a fully copy of the data.

I was assuming this, but looking at the bytes crate. I'm not sure this is actually true. It sounds like it's capable of converting a `Bytes` into a `BytesMut` (https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.from) and `BytesMut` into a `Vec<u8>` (https://docs.rs/bytes/latest/src/bytes/bytes_mut.rs.html#1727-1760) without copying, as long as it's all unique.

I tried swapping out my `Bytes` type for the bytes crate `Bytes` and it all just worked. Though, I can't be sure if it's doing a full copy or not, but by my reading it shouldn't be.

alexcrichton:
I can ask perhaps a different question, are you looking for zero-copy when you pass a `Bytes` to an imported function? This PR, for example, does not enable that. Conversion from `Bytes` to `Vec<u8>` I believe always performs a copy. I'm less sure about `BytesMut`, but reading the code it looks like the answer is "sometimes".

azdle:
Sorry if I added to the confusion by looking into the bytes crate types. But my original intent wasn't specifically about the `Bytes` and/or `BytesMut` types from the bytes crate. It was only about supporting newtype wrappers. The `Bytes` type in #1441 was just a newtype wrapper around a `Vec<u8>`.

When I started implementing this I just figured there was no reason to limit this to straight newtype wrappers and support anything that impl `From<Vec<u8>>` and `Into<Vec<u8>>`.

azdle:
At the risk of adding to the confusion, I decided to dig into what the bytes crate does in this case.

As long as you're not doing anything 'interesting' with the `Bytes` you can roundtrip through it without causing a full copy of the data.

When converting from `Vec<u8>` to `Bytes`, it will:

1. [Transform the `Vec` into a boxed slice.](https://github.com/tokio-rs/bytes/blob/v1.11.0/src/bytes.rs#L970)
2. [Turn that into a `Bytes` with a 'promotable' vtable.](https://github.com/tokio-rs/bytes/blob/v1.11.0/src/bytes.rs#L1013)


Then, as long as the `Bytes` hasn't been cloned (or otherwise turned into a `KIND_ARC`), it will:

1. [Call the `into_vec` from the promotable vtable.](https://github.com/tokio-rs/bytes/blob/v1.11.0/src/bytes.rs#L1057)
2. [Call the inner promote function.](https://github.com/tokio-rs/bytes/blob/v1.11.0/src/bytes.rs#L1263) ([Or here](https://github.com/tokio-rs/bytes/blob/v1.11.0/src/bytes.rs#L1302))
3. [Rebuild the original Vec from its parts.](https://github.com/tokio-rs/bytes/blob/v1.11.0/src/bytes.rs#L1229)

Though there is [a call to `ptr::copy`](https://github.com/tokio-rs/bytes/blob/v1.11.0/src/bytes.rs#L1227), it's [called with `ptr` and `buf` equal to each other](https://github.com/user-attachments/assets/a493dce5-d1bc-4c10-b0d0-2fe1036540e9) since our `Bytes` is pointing to the start of the original `Vec`, which I have to imagine doesn't cause any actual copying to happen.

[Edit: Because I was curious, my assumption is essentially correct, at least on my system / version of glibc. As long as there's more than 512 bytes of data it'll do a check on `src` and `dst` and realize it doesn't need to do anything. At 512 or below it just copies into registers and back. Which could be not ideal in some cases, but probably isn't that big of a deal, though that's getting beyond my experience to know for sure.]

But again, this PR isn't specifically about the bytes crate's `Bytes`, I just happened to name my newtype struct `Bytes` in the original issue.

---

I also pushed more test cases that I wrote to prove to myself that you weren't implying something else, so that's unrelated and I plan to remove the changes around non-canonical lists because that definitely _does_ have a copying problem that I'm not sure how I'd solve.

crates/rust/Cargo.toml
```diff
@@ -34,6 +34,7 @@ wit-bindgen = { path = '../guest-rust', features = ['async'] }
 test-helpers = { path = '../test-helpers' }
 # For use with the custom attributes test
 serde_json = "1"
+bytes = "*"
```
alexcrichton: Mind putting a version requirement here instead of just using `*`?

crates/rust/src/bindgen.rs
```diff
@@ -810,7 +814,7 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 uwriteln!(self.src, "let e{tmp} = {body};");
                 uwriteln!(self.src, "{result}.push(e{tmp});");
                 uwriteln!(self.src, "}}");
-                results.push(result);
+                results.push(format!("<_ as From<Vec<_>>>::from({result})"));
```
alexcrichton: From CI I believe this is giong to want to use `{vec}` instead of `Vec`

crates/rust/src/lib.rs
alexcrichton: Yeah this is fine no worries

crates/rust/Cargo.toml
azdle: Done

crates/rust/src/bindgen.rs
azdle: This line was what I was referring to when I said I was going remove the changes for non-canonical lists.

Reverted completely.

crates/rust/tests/codegen.rs
```diff
@@ -31,3 +31,140 @@ mod inline_and_path {
...
+        }
+    }
+
+    impl Deref for NewtypedListOfByte {
+        type Target = Vec<u8>;
+
+        fn deref(&self) -> &Self::Target {
+            &self.0
+        }
+    }
```
azdle, line 113: Just to note `Debug`, `Clone`, and `Deref` are needed because `NewtypedListOfBytes` appears in a record and the generated code for records requires them. As far as I can tell, `Clone` isn't strictly necessary, the generated code just derives `Clone` on the struct. Commenting out the clone derive here doesn't break anything for me: https://github.com/bytecodealliance/wit-bindgen/blob/f7634e9c544dda6574c5f905ad9f03e8203379e2/crates/rust/src/interface.rs#L2080

Ideally that clone would be derived only if all the children implement it, but as far as I can tell the only way to do that requires a bit of a hack using [some syntax that seems to be in stable by accident](https://github.com/rust-lang/rust/issues/48214#issuecomment-2799836786) (as well as generating the clone impl without derive). Though this is getting a bit off topic...

azdle:
I removed the changes around non-canonical lists and polished up the tests covering more generated code, making the naming more descriptive, and deduplicating a lot of things.

I think this should be ready to go now.

alexcrichton (approved):

PR #1443, opened 2025-11-27
Feat/update resources

JADSN1894:
1. Update the dependencies
2. Change the [tag definition implementation to the new api](https://github.com/JADSN1894/wit-bindgen/blob/0327b32ed25ea645681605cccef1c255e709f388/crates/markdown/src/lib.rs#L212)


alexcrichton:
Is there a need to update dependencies here? That's ideally separated out into a different PR to ensure changes aren't entangled.

PR #1444, opened 2025-11-30
Add new functions to generate rust bindings in `build.rs`

### LINKED ISSUE #1423: rust: Generating bindings in `build.rs`
I know there are two supported ways of generating Rust bindings - `generate!` macro and `wit-bindgen` CLI. 

When using the macro, I find myself regularly restarting `rust-analyzer` after every change in the `wit` folder, otherwise the RA does not see the changes.

Using the CLI has its own challenges, e.g. the need to specify `default_bindings_module` as discussed [in this issue](https://github.com/bytecodealliance/wit-bindgen/issues/973#issuecomment-2165950478)

The `build.rs` approach avoids both of those pitfalls.

```rs
use anyhow::Result;
use std::path::Path;
use wit_bindgen_rust::Opts;
use wit_parser::Resolve;

fn main() -> Result<()> {
    println!("cargo:rerun-if-changed=wit/");

    let opts = Opts {
        generate_all: true,
        ..Default::default()
    };
    let mut generator = opts.build();
    let mut resolve = Resolve::default();
    let (pkg, _files) = resolve.push_path("wit")?;
    let main_packages = vec![pkg];
    let world = resolve.select_world(&main_packages, None)?;
    let mut files = Default::default();
    generator.generate(&resolve, world, &mut files)?;

    let out_dir = std::env::var("OUT_DIR").unwrap();
    let dst = Path::new(&out_dir).join("generated.rs");
    let (_name, contents) = files.iter().next().unwrap();
    std::fs::write(&dst, contents)?;
    Ok(())
}
```
```rs
mod generated {
    #![allow(clippy::empty_line_after_outer_attr)]
    include!(concat!(env!("OUT_DIR"), "/generated.rs"));
}
```

Is this is a supported approach? If so, could it be simplified and documented?

tomasol:
Add `RustWasm::generate_to_out_dir` that parses the `wit` folder, selects a world, generates bindings and places them in a file in the `OUT_DIR`. `generate_to_out_dir_modify` allows modifying the generated file.

Sample `build.rs`
```rs
use anyhow::Result;
use std::borrow::Cow;
use wit_bindgen_rust::Opts;

fn main() -> Result<()> {
    Opts {
        generate_all: true,
        additional_derive_attributes: vec![
            "serde::Serialize".to_string(),
            "serde::Deserialize".to_string(),
        ],
        ..Default::default()
    }
    .build()
    .generate_to_out_dir_modify(Some("any"), |contents| {
        let contents = String::from_utf8(contents.to_vec()).unwrap();
        let re = regex::Regex::new(r"(pub\s+enum\s+\w+)").unwrap();
        Cow::Owned(
            re.replace_all(&contents, "#[serde(rename_all = \"kebab-case\")]\n$1")
                .into_owned()
                .into_bytes(),
        )
    })
}
```
Then the following snipped would include it:
```rs
mod generated {
    include!(concat!(env!("OUT_DIR"), "/wit_bindgen_generated.rs"));
}
```
Although this should be simplified as well.

Closes #1423

### DIFF
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -3,11 +3,10 @@ use anyhow::{Result, bail};
 use core::panic;
 use heck::*;
 use indexmap::{IndexMap, IndexSet};
-use std::borrow::Cow;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt::{self, Write as _};
 use std::mem;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use wit_bindgen_core::abi::{Bitcast, WasmType};
 use wit_bindgen_core::{
@@ -287,33 +286,31 @@ impl Opts {
     }
 }
 
-const GENERATED_FILE_NAME: &str = "wit_bindgen_generated.rs";
 impl RustWasm {
-    pub fn generate_to_out_dir(self, world: Option<&str>) -> Result<()> {
-        self.generate_to_out_dir_modify(world, |s| Cow::Borrowed(s))
-    }
-
-    pub fn generate_to_out_dir_modify(
-        mut self,
-        world: Option<&str>,
-        modify: impl FnOnce(&[u8]) -> Cow<[u8]>,
-    ) -> Result<()> {
+    /// Generates Rust bindings from the `wit/` directory and writes
+    /// the result into Cargo’s `OUT_DIR`. Intended for use in `build.rs`.
+    ///
+    /// The `world` parameter specifies the world name to select.
+    /// It must be provided unless the main package contains exactly one world.
+    ///
+    /// Returns the full path to the generated bindings file.
+    pub fn generate_to_out_dir(mut self, world: Option<&str>) -> Result<PathBuf> {
         let mut resolve = Resolve::default();
         println!("cargo:rerun-if-changed=wit/");
         let (pkg, _files) = resolve.push_path("wit")?;
         let main_packages = vec![pkg];
         let world = resolve.select_world(&main_packages, world)?;
+
         let mut files = Files::default();
         self.generate(&resolve, world, &mut files)?;
         let out_dir = std::env::var("OUT_DIR").expect("cargo sets OUT_DIR");
-        let dst = Path::new(&out_dir).join(GENERATED_FILE_NAME);
-        let (_name, contents) = files
+        let (name, contents) = files
             .iter()
             .next()
             .expect("exactly one file should be generated");
-        let contents = modify(contents);
+        let dst = Path::new(&out_dir).join(name);
         std::fs::write(&dst, contents)?;
-        Ok(())
+        Ok(dst)
     }
 
     fn new() -> RustWasm {


crates/rust/src/lib.rs
```diff
@@ -277,15 +279,43 @@ pub struct Opts {
...
+    ) -> Result<()> {
+        let mut resolve = Resolve::default();
+        println!("cargo:rerun-if-changed=wit/");
+        let (pkg, _files) = resolve.push_path("wit")?;
+        let main_packages = vec![pkg];
+        let world = resolve.select_world(&main_packages, world)?;
+        let mut files = Files::default();
+        self.generate(&resolve, world, &mut files)?;
+        let out_dir = std::env::var("OUT_DIR").expect("cargo sets OUT_DIR");
+        let dst = Path::new(&out_dir).join(GENERATED_FILE_NAME);
```
alexcrichton: Instead of using a hardcoded name could all the files be written with the name from the generator itself?

crates/rust/src/lib.rs
```diff
@@ -277,15 +279,43 @@ pub struct Opts {
...
     }
 }
 
+const GENERATED_FILE_NAME: &str = "wit_bindgen_generated.rs";
 impl RustWasm {
+    pub fn generate_to_out_dir(self, world: Option<&str>) -> Result<()> {
+        self.generate_to_out_dir_modify(world, |s| Cow::Borrowed(s))
+    }
+
+    pub fn generate_to_out_dir_modify(
```
alexcrichton: Mind adding some doc comments for these methods?

crates/rust/src/lib.rs
```diff
@@ -277,15 +279,43 @@ pub struct Opts {
...
+        mut self,
+        world: Option<&str>,
+        modify: impl FnOnce(&[u8]) -> Cow<[u8]>,
```
alexcrichton: This isn't necessarily tightly integrated into the generation step, so could this be omitted and performed as part of the top-level build itself? For example the caller of `generate_to_out_dir` would read the output file, perform the transformation, and then write back out the contents.

crates/rust/src/lib.rs
tomasol: Sure, I've replaced it with the selected world name.

crates/rust/src/lib.rs
tomasol: Agreed, removed `generate_to_out_dir_modify`

alexcrichton (approved):
Thanks!

Was there anything else you'd like to address before merging?

tomasol:
Thanks for the quick review! Additionally I would like to add a macro to include the generated file:
```rs
#[cfg(feature = "macros")]
#[macro_export]
macro_rules! include_wit {
    ($name:literal) => {
        include!(concat!(env!("OUT_DIR"), "/", $name, ".rs"));
    };
}
```
so that the bindings can be included simply with `wit_bindgen::include_wit!("myworld");`.
Is it a good idea to add it to `wit-bindgen` ?

alexcrichton:
Personally I'd say that should be left out of the crate and documented in the usage of this method. I (subjectively) don't feel that clears the threshold for inclusion just yet, but if it becomes a common/frequent way of using wit-bindgen it seems ok to add.

tomasol:
Understood, in that case the PR is done.

tschneidereit:
Looks like an intermittent failure, so let's try again

tschneidereit:
*Sigh*, I think Github infra might be having a bad day. I'll try again tomorrow

PR #1447, opened 2025-12-03
reboot `wit-bindgen-go`

dicej:
This re-introduces Go support, with a new implementation written from scratch. While the previous incarnation targeted TinyGo and reused the C generator for lifting and lowering, this one targets "big" Go and handles lifting and lowering itself.  In addition, it supports idiomatic, goroutine-based concurrency on top of the component model async ABI, including streams and futures.

This implementation is also distinct from the
[go-modules](https://github.com/bytecodealliance/go-modules) project, which has a number of limitations (e.g. no async support, not safe for use with "big" Go) and is no longer being actively maintained.

Note that the async support currently requires [a small patch](https://github.com/dicej/go/commit/a1c83220fc9576cdb810e9624366cb998e69b22b) to the Go `runtime` library.  I plan to work with the upstream project to make that patch unecessary in the future.  Components that don't use async features should work with stock, unpatched Go.

One of the tricky parts about lowering values and passing pointers to e.g. `{stream,future}.{read,write}` is that we must tell the Go garbage collector to pin the pointers (i.e. mark the pointee as both immovable and uncollectable) for as long as the host has access to them, but then unpin them promptly afterward to avoid leaks.  We do this using `runtime.Pinner` instances, most of which are locally scoped and easy to reason about.  However, we must use a couple of globally-scoped pinners as well: one for `cabi_realloc` allocations by the host when it lowers parameters (which we unpin after lifting those parameters) and another for returning values from sync-lifted exports (which we unpin in post-return functions).

There are a couple of known missing features which I'll open GitHub issues for:

1. Resource handles are not being restored for unwritten items in the case of an incomplete stream or future write.
2. Importing and/or exporting multiple versions of the same package will cause name clashes.

In addition, I plan to expand the test coverage beyond the basics covered in this commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,11 @@ Now you can build with:
 dotnet publish
 ```
 
-Checkout out [componentize-dotnet](https://github.com/bytecodealliance/componentize-dotnet) for a simplified experience.
+Check out [componentize-dotnet](https://github.com/bytecodealliance/componentize-dotnet) for a simplified experience.
+
+### Guest: Go
+
+See the [wit-bindgen-go README.md](crates/go/README.md) for details and generating and using Go bindings.
 
 ### Guest: Java
 
--- a/crates/go/README.md
+++ b/crates/go/README.md
@@ -1,0 +1,113 @@
+# `wit-bindgen` Go Bindings Generator
+
+This tool generates Go bindings for a chosen WIT world.
+
+## Usage
+
+To generate bindings with this crate, issue the `go` subcommand to `wit-bindgen`:
+
+```bash
+$ wit-bindgen go [OPTIONS] <WIT>
+```
+
+See the output of `wit-bindgen help go` for available options.
+
+This command will generate a variable number of files, depending on the WIT
+world provided:
+
+- `go.mod`: defines a minimal Go module with the name `wit_component`
+    - You can replace this with your own version (e.g. referencing third party dependencies) if desired
+- `wit_bindings.go`: defines the `main` package for the module, including low-level, `//go:export`-annotated entrypoint functions corresponding to exported functions
+    - These entrypoint functions in turn call high-level functions which must be provided by the application developer
+- `wit_runtime/wit_runtime.go`: defines low-level functions for supporting the component model ABI
+- `<name>/wit_bindings.go`: defines any types generated for the interface named `<name>` (or `wit_world` for WIT types defined at the world level), plus any imported functions
+    - Note that the types placed in these files include all types for both imported and exported interfaces, except for exported resource types and any types which depend on exported resource types
+- `export_<name>/wit_bindings.go`: defines intrinsics for use with any exported resource types generated for the interface named `<name>` (or `wit_world` for WIT types defined at the world level), plus any types which depend on those exported resource types, plus any exported functions
+    - The exported resource type definitions must be provided by the application developer
+    - The `export_<name>` package is also the place to define any exported functions
+- (if needed) `wit_types/wit_tuples.go`: defines `Tuple<N>` types as required by the WIT world
+- (if needed) `wit_types/wit_async.go`: defines low-level functions for integrating the Go scheduler with the component model async ABI
+- (if needed) `wit_types/wit_option.go`: defines an `Option` type if required by the WIT world
+- (if needed) `wit_types/wit_result.go`: defines an `Result` type if required by the WIT world
+- (if needed) `wit_types/wit_unit.go`: defines an `Unit` type if required by the WIT world
+- (if needed) `wit_types/wit_stream.go`: defines a `StreamReader` and `StreamWriter` types if required by the WIT world
+- (if needed) `wit_types/wit_future.go`: defines a `FutureReader` and `FutureWriter` types if required by the WIT world
+
+## Example
+
+### Prerequisites
+
+- `wit-bindgen`
+- `wasm-tools`
+- `wasmtime`
+- `curl`
+- `bash` or similar
+
+Given the following WIT file, we can generate bindings for it, write code to
+target the two worlds, and finally compose, build, and run the components.
+
+```shell
+cat >world.wit <<EOF
+package test:test;
+
+interface foo {
+  record thing {
+    a: s32,
+    b: string,
+  }
+  
+  echo: func(x: thing) -> tuple<s32, string>;
+}
+
+world test {
+  export foo;
+}
+
+world runner {
+  import foo;
+  export run: func() -> tuple<s32, string>;
+}
+EOF
+
+curl -OL https://github.com/bytecodealliance/wasmtime/releases/download/v39.0.1/wasi_snapshot_preview1.reactor.wasm
+mkdir test runner
+
+pushd test
+wit-bindgen go -w test ../world.wit
+mkdir export_test_test_foo
+cat >export_test_test_foo/test.go <<EOF
+package export_test_test_foo
+
+import . "wit_component/test_test_foo"
+
+func Echo(x Thing) (int32, string) {
+    return x.A, x.B
+}
+EOF
+GOARCH="wasm" GOOS="wasip1" go build -o core.wasm -buildmode=c-shared -ldflags=-checklinkname=0
+wasm-tools component embed -w test ../world.wit core.wasm -o core-with-wit.wasm
+wasm-tools component new --adapt ../wasi_snapshot_preview1.reactor.wasm core-with-wit.wasm -o component.wasm
+popd
+
+pushd runner
+wit-bindgen go -w runner ../world.wit
+mkdir export_wit_world
+cat >export_wit_world/runner.go <<EOF
+package export_wit_world
+
+import . "wit_component/test_test_foo"
+
+func Run() (int32, string) {
+    return Echo(Thing{42, "hello, world!"})
+}
+EOF
+GOARCH="wasm" GOOS="wasip1" go build -o core.wasm -buildmode=c-shared -ldflags=-checklinkname=0
+wasm-tools component embed -w runner ../world.wit core.wasm -o core-with-wit.wasm
+wasm-tools component new --adapt ../wasi_snapshot_preview1.reactor.wasm core-with-wit.wasm -o component.wasm
+popd
+
+wasm-tools compose -d test/component.wasm runner/component.wasm -o component.wasm
+wasmtime run --invoke 'run()' component.wasm
+```
+
+If all goes well, you should see `(42, "hello, world!")`.


dicej:
I just realized I need to make CI run the go tests.  I'll push an update.

crates/go/src/wit_future.go
```diff
@@ -0,0 +1,131 @@
...
+
+	pinner := runtime.Pinner{}
+	defer pinner.Unpin()
+
+	buffer := wit_runtime.Allocate(&pinner, uintptr(f.vtable.Size), uintptr(f.vtable.Align))
+
+	code, _ := wit_async.FutureOrStreamWait(f.vtable.Read(handle, buffer), handle)
+
+	switch code {
+	case wit_async.RETURN_CODE_COMPLETED, wit_async.RETURN_CODE_DROPPED:
```
alexcrichton: `RETURN_CODE_DROPPED` here should be a panic, right? (it shouldn't be possible anyway to drop the writer without writing)

crates/go/src/wit_future.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func (f *FutureReader[T]) Drop() {
+	handle := f.handle
+	if handle != 0 {
+		f.handle = 0
+		f.vtable.DropReadable(handle)
+	}
+}
```
alexcrichton: I'd recommend pairing this with a finalizer installed via the GC on `f` to auto-call this if it's not otherwise called. Calling this explicitly would still be supported but it'd optionally be cleaned up if forgotten or otherwise not called.

crates/go/src/wit_future.go
```diff
@@ -0,0 +1,131 @@
...
+	f.handle = 0
+	defer f.vtable.DropWritable(handle)
+
+	pinner := runtime.Pinner{}
+	defer pinner.Unpin()
+
+	var buffer unsafe.Pointer
+	if f.vtable.Lower == nil {
+		buffer = unsafe.Pointer(unsafe.SliceData([]T{item}))
+		pinner.Pin(buffer)
```
alexcrichton, line 93: To confirm, it's not possible for the Go GC to run between these two calls? Or is there something that needs to be done to say either "do this atomically" or "don't GC while I'm in here"?

crates/go/src/wit_option.go
```diff
@@ -0,0 +1,41 @@
+package wit_types
```
alexcrichton, line 1: Question on these packages: how are you envisioning that they're going to be distributed? For example should these be published as a standalone package that wit-bindgen refers to? Do these otherwise play well if there's multiple versions of the runtime linked into one binary?

Or is this basically "works at the app level nowhere else" for now?

crates/go/src/wit_runtime.go
```diff
@@ -0,0 +1,83 @@
...
+
+	switch align {
+	case 1:
+		return unsafe.Pointer(unsafe.SliceData(make([]uint8, size)))
+	case 2:
+		return unsafe.Pointer(unsafe.SliceData(make([]uint16, ceiling(size, align))))
+	case 4:
+		return unsafe.Pointer(unsafe.SliceData(make([]uint32, ceiling(size, align))))
+	case 8:
+		return unsafe.Pointer(unsafe.SliceData(make([]uint64, ceiling(size, align))))
```
alexcrichton: Did you find the calls to `ceiling` necessary or were those copy/pasted from elsewhere? I would otherwise expect them to not be necessary as the canonical ABI should guarantee that `size % align == 0` so `size / align` would be an appropriate calculation here

crates/go/src/wit_runtime.go
```diff
@@ -0,0 +1,83 @@
...
+func sbrk(n uintptr) unsafe.Pointer
+
+var useGCAllocations = false
+
+func init() {
+	useGCAllocations = true
+}
+
+func offset(ptr, align uintptr) uintptr {
+	newptr := (ptr + align - 1) &^ (align - 1)
```
alexcrichton, line 86: Is `&^` a typo? If so I'm curious, what does it do?

crates/go/src/wit_stream.go
```diff
@@ -0,0 +1,179 @@
...
+	} else {
+		result := make([]T, 0, count)
+		for i := 0; i < int(count); i++ {
+			result = append(result, s.vtable.Lift(unsafe.Add(buffer, i*int(s.vtable.Size))))
+		}
+		return result
+	}
+}
+
+func (s *StreamReader[T]) ReadInto(dst []T) uint32 {
```
alexcrichton: Could this be the only interface that this module supports? And depending on `Lift` it either reads directly into the provided slice or otherwise allocates temporary storage and then deserializes from there?

tests/runtime-async/async/simple-call-import/runner.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	test "wit_component/a_b_i"
+	"wit_component/wit_async"
+)
+
+//go:wasmimport [export]wasi:cli/run@0.3.0-rc-2025-09-16 [task-return]run
+func taskReturn(result int32)
+
+//go:wasmexport [async-lift]wasi:cli/run@0.3.0-rc-2025-09-16#run
```
alexcrichton: I'm a bit surprised by these directives, is this not something we'd expect the bindings generator to take care of?

tests/runtime-async/async/simple-future/runner.go
```diff
@@ -0,0 +1,57 @@
...
+	"wit_component/wit_async"
+	"wit_component/wit_types"
+)
+
+//go:wasmimport [export]wasi:cli/run@0.3.0-rc-2025-09-16 [task-return]run
+func taskReturn(result int32)
+
+//go:wasmexport [async-lift]wasi:cli/run@0.3.0-rc-2025-09-16#run
+func run() uint32 {
+	return wit_async.Run(func() {
```
alexcrichton: Similar to above, I'd naively expect this `wit_async.Run(func() { ... })` dance to be part of the bindings generator

alexcrichton (commented):
Agreed we'll want to ge this running on CI, but overall looking great! I only skimmed codegen though and would mostly lean on tests to exercise bits and bobs

crates/go/src/wit_future.go
dicej: I'll just link to [my previous rant on this topic](https://bytecodealliance.zulipchat.com/#narrow/channel/217126-wasmtime/topic/Freeing.20resources.20in.20go.20bindings/near/421715248).

crates/go/src/wit_future.go
dicej: Even if the GC does run between these calls, it's not a problem.  Recall that the GC sees `unsafe.Pointer` values as part of its scan.  All we need to do is make sure we pin the pointer before converting it to a `uintptr` (at which point it becomes invisible to the GC).

crates/go/src/wit_option.go
dicej: > Or is this basically "works at the app level nowhere else" for now?

Correct, and I think it's a reasonable place to start, at least.  I do think we'll want to revisit that as we start publishing SDKs and/or an official WASI package containing pre-generated bindings, but I don't feel we need to address that right away.

crates/go/src/wit_runtime.go
dicej: Yeah, I just did that out of paranoia.  I'll replace it with an assertion.

crates/go/src/wit_runtime.go
dicej: I stole that code from @tschneidereit's [PR](https://github.com/bytecodealliance/go-modules/pull/367) :shrug:

crates/go/src/wit_stream.go
dicej: I don't think we can use this exact signature for an arbitrary `T`, since it would require that `dst` be already sized to the maximum desired item count and thus filled with already-initialized values.  That wouldn't be possible if `T` is e.g. an imported resource type with no constructor, and even if it was constructable it wouldn't make sense to force the caller to fill the slice with dummy data just to have it overwritten.

We _could_ however do something like `ReadInto(dst []T) []T` such that the caller passes a slice with zero size and nonzero capacity and returns the same slice with some number of items added.  I'm still getting used to how Go slices work, but I think that's the idiom.

crates/go/src/wit_future.go
alexcrichton: Well, to clarify, I'm not saying this method should be removed. Can you explain why a finalizer specifically should not be present?

My impression is that a huge benefit of GC languages is if you don't think about resources everything works out and there are no leaks. Without a finalizer then programs will leak unless `Drop` is called, which to me I feel like will trip up a lot of folks using this. I understand that conventional usage will probably call `Drop`, but inevitably that will be forgotten and/or some error path will be encountered that doesn't call `Drop`, and to me it seems like it would be unfortunate if that results in a leak.

crates/go/src/wit_future.go
alexcrichton: Oh right, indeed!

crates/go/src/wit_option.go
alexcrichton: Makes sense, and I understand how this is a starting point, but would it be worthwhile to try to chart a course ahead? For example is there a possible viable path to using bindgen in libraries one day? (e.g. like `wit-bindgen` in Rust)

crates/go/src/wit_stream.go
alexcrichton: My recollection of Go is that any type can be allocated with a zero pattern at any time, so it's not possible to have a public type with representation requirements. For example if `T` were an imported resource with no constructor then I believe you could still make a slice of `T` and the expectation is that you should initialize it before use. Given that I think it would be possible to still make this API I'm thinking.

Now whether it's idiomatic, that I don't know.

crates/go/src/wit_runtime.go
tschneidereit: `&^` is `AND NOT`, i.e. this aligns the pointer, and the function returns offset between the aligned and unaligned addresses. As the name "offset" clearly says, obviously. (Sorry for the poor naming and lack of documentation!)

crates/go/src/wit_stream.go
dicej: > My recollection of Go is that any type can be allocated with a zero pattern at any time, so it's not possible to have a public type with representation requirements.

Yeah, if that's true, and the caller can do e.g. `make([]T, 1024)` for arbitrary `T` without futher initialization, then you could be right.  Runs completely counter to my intuition for type safety, but if that's how Go works then we can ~~go~~ roll with it.

crates/go/src/wit_future.go
dicej: My position is that, because the GC is only concerned about memory pressure and is completely oblivious to non-memory resources such as resource/future/stream handles, it should not be responsible for disposing of those handles.  As an application programmer, I'd prefer to leak them entirely rather than have them disposed at some unpredictable future time determined by an opaque algorithm whose job is completely unrelated to such things.

It feels akin to saying "if I forget to explicitly dispose this handle, please dispose of it for me the next time it rains in New York City".  It makes reproducing and debugging resource management issues harder, not easier.  I spent years dealing with this in the Java world before wising up and changing all the finalizers to log big, noisy error messages instead of trying to silently compensate for inadequate resource management.

crates/go/src/wit_option.go
dicej: I think there's definitely a path forward: `wit_runtime.go`, `wit_async.go`, `wit_option.go`, `wit_stream.go`, etc. could all be published as part of a reusable package which such libraries could depend on.  I just don't feel we need that right away.

tests/runtime-async/async/simple-call-import/runner.go
dicej: Yeah, this is awkward, and it's a result of a combination of factors:

- `wit-bindgen-test` expects all the `runner` components to export `wasi:cli/run`, and yet none of the `runner` worlds in the `test.wit` files actually export that interface, so the bindings generator never sees it.
- As of this writing, Go-generated modules are incompatible with `wasi_snapshot_preview1.command.wasm` (there's an unbounded recursion during runtime initialization), so we have to use `wasi_snapshot_preview1.reactor.wasm` instead, and that won't automatically export `wasi:cli/run`.
- For the async tests, we must export `wasi:cli/run@0.3.0-rc-2025-09-16` because `wasi:cli/run@0.2.x` is a sync function, and we can't return e.g. `CALLBACK_CODE_WAIT` if we implement the latter.  The Rust tests currently use `wit_bindgen::block_on`, but that's no longer viable given the new trapping behavior.

I'm open to alternative approaches to addressing these challenges; what I have now works, but it might not be ideal.

tests/runtime-async/async/simple-future/runner.go
dicej: See https://github.com/bytecodealliance/wit-bindgen/pull/1447#discussion_r2590021906

dicej:
I was so focused on the runtime tests that I never ran the codegen ones.  Looks like I have more work to do.  I'll switch this to draft mode while I do that.

crates/go/src/wit_future.go
alexcrichton: Talked with Joel about this and the conclusion we had was "do what `os.File` does" which looks to have close + a finalizer, so current conclusion is to add finalizers to these streams + keep the close method.

crates/go/src/wit_runtime.go
alexcrichton: TIL!

tests/runtime-async/async/simple-call-import/runner.go
alexcrichton: Talked with Joel -- conclusion here is that this is because the WIT export isn't actually part of the WIT used to generate this test case, so it's expected.

crates/go/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "wit-bindgen-go"
+edition = "2024"
+version = { workspace = true }
+repository = { workspace = true }
+license = { workspace = true }
+rust-version = "1.85.0"
```
alexcrichton: Can this (and edition above) also be set to `{ workspace = true }`?

alexcrichton: Also, for publication to crates.io, I think this'll want `description` + `homepage` like other crates

crates/test/src/go.rs
```diff
@@ -0,0 +1,140 @@
...
+        &self,
+        _name: &str,
+        config: &crate::config::WitConfig,
+        _args: &[String],
+    ) -> bool {
+        // TODO: We _do_ support async, but only with a build of Go that has
+        // [this
+        // patch](https://github.com/dicej/go/commit/a1c83220fc9576cdb810e9624366cb998e69b22b).
+        // Once we either publish builds containing that patch or upstream
+        // something equivalent, we can remove the ` || config.async_` here.
```
alexcrichton, line 29: How hard would it be to configure CI to have a Go with this patch? Not worth it?

tests/runtime-async/async/simple-call-import/runner.go
dicej: This has been addressed by https://github.com/bytecodealliance/wit-bindgen/pull/1448

crates/go/Cargo.toml
dicej: I'll add `description` and `homepage`.

> Can this (and edition above) also be set to `{ workspace = true }`?

I'm using `let` chaining, which requires edition 2024.  I can convert that code if needed; are we not ready to start using 2024 in this project, yet?

crates/test/src/go.rs
dicej: I haven't had a chance to look into it.  Hopefully not hard, but I'd prefer to open an issue about this and address it as a follow-up, if that's ok with you.

crates/go/Cargo.toml
alexcrichton: I'd prefer to keep the repo in sync myself. We don't have a msrv policy in this repo, though, so I think it's fine to just bump the workspace to 1.84 and the 2024 edition

alexcrichton (approved):

ydnar:
Thanks to @dgryski for pointing me to this!

@dicej this looks great. We solved the multiple versions issue by optionally generating a versioned Go package, e.g. `wasi/http/v0.2.1`.

Does this require the WASI 0.1 adapter to function in big Go?

dicej:
> @dicej this looks great. We solved the multiple versions issue by optionally generating a versioned Go package, e.g. `wasi/http/v0.2.1`.

Good idea.  Currently we generate package names like `wasi_http_0_2_1_types` because I'm a Go novice, but versioning the Go package makes more sense.

> Does this require the WASI 0.1 adapter to function in big Go?

Yes, currently.  I've opened https://github.com/golang/go/pull/76775 as a baby step towards adding `GOOS=wasip3` upstream.  Meanwhile, you can use any p2 or p3 features you like by generating bindings from the WASI WIT files and using them directly.

BTW, [here's an example](https://github.com/dicej/go-wasi-http-example) that shows off concurrent outbound requests, etc.  Eventually we'll want to do something like you did [here](https://github.com/ydnar/wasi-http-go/) to make that use `net/http` interfaces, etc.

ydnar:
I haven’t dug too deeply into the code—how does it handle the lifting/lowering of pointer types?

Big Go uses 64-bit pointers, which means that `record` types containing pointers (`list`, `string`) aren’t ABI compatible with WASI.

Our plan with the go-modules repo and `wit-bindgen-go` was to rely on big Go merging support for `GOARCH=wasm32`. TinyGo’s pointer size on wasm is 32-bit.

dicej:
> I haven’t dug too deeply into the code—how does it handle the lifting/lowering of pointer types?

It doesn't assume the canonical ABI layout of a given `record` or `variant` type matches the layout of the Go representation.  Instead, the record fields and variant payloads are lifted and lowered one-at-a-time, with pointers converted between `uintptr` and `uint32` as necessary.  We know those conversions will be lossless because there's no way to create a valid pointer with any of the high 32 bits set in a wasm32 module.

ydnar:
Got it. So when a pointer to a list<T> containing pointers is passed, does that require copying on every function call?

We were able to guarantee the Go binary representation matched the canonical ABI, to avoid copying.

dicej:
> Got it. So when a pointer to a list containing pointers is passed, does that require copying on every function call?

Yes, and not only that.  We must also pin those pointers for GC safety when converting them to the canonical ABI, and also unpin them when converting from the ABI.  The pin/unpin requirement will remain even if/when `GOARCH=wasm32` becomes a thing, so lifting and lowering lists containing pointers will remain O(n) operations regardless.

> We were able to guarantee the Go binary representation matched the canonical ABI, to avoid copying.

Indeed; Big Go required a different approach, though.

dgryski:
What needs to be done to this implementation to allow this to work with TinyGo?

dicej:
> What needs to be done to this implementation to allow this to work with TinyGo?

I don't yet have a great mental model of how TinyGo differs from Big Go, so I don't know.  Does TinyGo support e.g. [runtime.Pinner](https://pkg.go.dev/runtime#Pinner)?

Also, IIRC, TinyGo relies on Asyncify to support concurrency.  We'd need to hook into that somehow to bridge goroutines with the component model async ABI and thereby ensure that I/O operations only block the calling goroutine, allowing others to continue running.  For Big Go, I had to patch the `runtime` library to get special access to the scheduler; not sure what the equivalent would be for TinyGo.

One way to start would be to make an example app based on bindings generated by this tool and try building it with TinyGo to see what breaks.

dicej:
BTW, how does `go-modules` handle GC safety with TinyGo?  It looks like TinyGo uses the BDW conservative GC library.  Is there some mechanism to ensure allocations made by the host via `cabi_realloc` are left untouched during GC until the host has finished using them?

dgryski:
We have a partial fix here https://github.com/tinygo-org/tinygo/pull/4897 but will need a better one.  We don't support pinning (yet) but we also don't have a moving collector.  (There three garbage collectors available in TinyGo.).  Supporting "don't free this yet" flags would fall under something similar to the above PR.

Yes, moving to wasip3 would mean moving away from asyncify, which is something we're looking forward to getting rid of.

dicej:
> We don't support pinning (yet) but we also don't have a moving collector.

You might already know this, but `runtime.Pinner` doesn't just prevent an allocation from moving but also ensures that it isn't collected.  So perhaps TinyGo's version of `runtime.Pinner` could just store the pointers it is given and make sure they're treated as roots by the GC.

dgryski:
Yes, that's what I meant by the "don't free this yet" flag.  The PR I linked to attempts to avoid that by storing pointers in a map, but having wit-bindgen-go explicity pin/unpin allocations would solve some of the issues that PR had.

PR #1451, opened 2025-12-07
Small updates to the Go portions of the README 

asteurer:
Making it clear that using `Guest: TinyGo` is no longer the recommended way to generate Go bindings.


alexcrichton:
Thanks!

PR #1457, opened 2025-12-08
run `gofmt` on generated code if available

dicej:


dicej:
> Mind making this a configurable option as well? Effectively aligning with Rust's `--format` option for `wit-bindgen rust`

Sure.  It looks like formatting is disabled by default for Rust.  Any particular reason for that vs. enabling it by default and providing a `--no-format` option instead?

alexcrichton:
I personally prefer "positive" options along the lines of `--format` and `--format=no` vs `--no-format` since I find negatives/double-negatives hard to understand.

Clap supports `--foo[=true|false]` which means this flag can be `--format`, on-by-default, and disabled with `--format=false`, which I think is fine.

dicej:
> I personally prefer "positive" options along the lines of `--format` and `--format=no` vs `--no-format` since I find negatives/double-negatives hard to understand.
> 
> Clap supports `--foo[=true|false]` which means this flag can be `--format`, on-by-default, and disabled with `--format=false`, which I think is fine.

Makes sense, but am I correct that it's off-by-default for Rust?  Should we turn it on by default there to match what I'll be doing for Go?

alexcrichton:
You're right yeah it's off-by-default, but for now I think it's ok for different generators to differ on defaults. I think it's fine to leave it disabled-by-default in Rust since generated code is handled differently than the Go side of things.

PR #1463, opened 2025-12-08
go: adding helper methods for result and option

### LINKED ISSUE #1462: Go: `Result` and `Option` type handling
I'm wondering if the Option and Result handling in the Go portion would be easier to use with some changes to the API. Here's one possibility:

```go

import (
    wit "wit_component/wit_types"
)

optionExample := wit.Some("foo")
resultExample := wit.Ok[int, error](0)

// Current implementation
if optionExample.Tag() == wit.OptionSome {
// ...
}

if resultExample.Tag() == wit.ResultErr {
// ...
}

// Suggested implementation
if optionExample.IsNone() {
// ...
}

if resultExample.IsOk() {
// ...
}
```

asteurer:
Closes #1462

### DIFF
--- a/crates/go/src/wit_result.go
+++ b/crates/go/src/wit_result.go
@@ -28,6 +28,14 @@ func (self Result[T, U]) Err() U {
 	return self.value.(U)
 }
 
+func (self Result[T, U]) IsErr() bool {
+	return self.tag == ResultErr
+}
+
+func (self Result[T, U]) IsOk() bool {
+	return self.tag == ResultOk
+}
+
 func Ok[T any, U any](value T) Result[T, U] {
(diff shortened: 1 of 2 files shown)


asteurer:
@dicej

crates/go/src/wit_result.go
```diff
@@ -28,6 +28,14 @@ func (self Result[T, U]) Err() U {
 	return self.value.(U)
 }
 
+func (self Result[T, u]) IsErr() bool {
```
dicej: Any particular reason the `u` is lowercase here and below?

asteurer: 🤦 Not intentional haha

Issue #1415, opened 2025-11-13
rust: sharing type definitions across multiple interfaces


chenyan2002:
Consider the following WIT file:

```
package test:component;

interface i {
  resource r;
  enum kind { a, b, c }
  f: func(k: kind) -> r;
}
// wrapped-i is exactly the same as i
interface wrapped-i {
  resource r;
  enum kind { a, b, c }
  f: func(k: kind) -> r;
}

world root {
  import i;
  import wrapped-i;
  export i;
  export wrapped-i;
}
```

In wit-bindgen, `kind` will be defined four times: `test::component::i::Kind`, `test::component::wrapped_i::Kind`, `exports::test::component::i::Kind`, and `exports::test::component::wrapped_i::Kind`. It would be good to have an option to say that these four `kind` are the same, and we only define it once, the remaining module will use type alias instead. This avoids value cloning when we want to pass `kind` from one interface to another. This only works for types which contain no resources. With the lazy value proposal, this may change though.

I can see three options:

1) add a `interface_group` flag, so that we can say `import.i`, `import.wrapped-i`, `export.i`, `export.wrapped-i` are in the same group, and the bindgen will find all non-resource types in the interface and make sure they are defined only once.

2) We could also fix this by extending the WIT syntax, something like 
```
world root {
  import i;
  import i as wrapped-i;
  export i;
  export i as wrapped-i
}
```
Then wit-bindgen will know that the non-resource types in these four interfaces are the same just by looking at the WIT file.

3) add a `type_eq` flag, and say `import.i.kind`, `import.wrapped-i.kind`, `export.i.kind` and `export.wrapped-i.kind` are equal. This allows more fine grained control, and can handle cases where the two interfaces are not exactly the same, but we still want to share some data types.

It feels like that the interface level equality is better fixed from the WIT syntax, but the type level equality is a suitable extension in wit-bindgen alone and can also cover the interface equality case without waiting for a syntax change.

alexcrichton:
I agree it'd be nice to solve this! I was talking to folks in Munich recently and this came up as something that would be very nice there too, so you're definitely not alone.

Personally I'd prefer for this to be automatic rather than requiring an opt-in and/or configuration. I think it'd be relatively reasonable to as it's a somewhat-simple hash/equality function of a type. For example records-with-resources are all the same if the resources themselves are the same, so there's no need to handle resources separately per se.

[...]

PR #1468, opened 2025-12-15
Rust: merge structurally equal types in bindgen

chenyan2002:
Fix https://github.com/bytecodealliance/wit-bindgen/issues/1415

Add `merge_structurally_equal_types` flag to merge structurally equal types.

* We consider all resource, future and stream types as non-equal, even for resource types with the same TypeId. The reason is that we always need to update the resource table when moving around resources, so we cannot use type alias for any types that contain resources.
* In preprocess, `collect_equal_types` takes O(n^2) time. If performance is a concern, I can define a hash function to bring down the time to O(n).

### DIFF
--- a/crates/core/src/types.rs
+++ b/crates/core/src/types.rs
@@ -306,10 +306,10 @@ impl Types {
             (TypeDefKind::Tuple(_), _) => false,
             (TypeDefKind::List(la), TypeDefKind::List(lb)) => self.types_equal(resolve, la, lb),
             (TypeDefKind::List(_), _) => false,
-            (TypeDefKind::FixedSizeList(ta, sa), TypeDefKind::FixedSizeList(tb, sb)) => {
+            (TypeDefKind::FixedLengthList(ta, sa), TypeDefKind::FixedLengthList(tb, sb)) => {
                 sa == sb && self.types_equal(resolve, ta, tb)
             }
-            (TypeDefKind::FixedSizeList(..), _) => false,
+            (TypeDefKind::FixedLengthList(..), _) => false,
             (TypeDefKind::Option(oa), TypeDefKind::Option(ob)) => self.types_equal(resolve, oa, ob),
             (TypeDefKind::Option(_), _) => false,
             (TypeDefKind::Result(ra), TypeDefKind::Result(rb)) => {
--- a/crates/rust/src/interface.rs
+++ b/crates/rust/src/interface.rs
@@ -2339,38 +2339,6 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8) 
         }
     }
 
-    pub fn type_alias_to_eqaul_type(
-        &mut self,
-        id: TypeId,
-        eq_ty: TypeId,
-        from_import: Option<&WorldKey>,
-    ) {
-        assert!(self.r#gen.opts.merge_structurally_equal_types());
-        if let Some(name) = from_import {
-            let docs = Docs {
-                contents: Some("wit-bindgen: alias to import equal type".to_string()),
-            };
-            let mut path = self.path_to_root();
-            let import_path = crate::compute_module_path(name, self.resolve, false).join("::");
-            path.push_str(&import_path);
-            path.push_str("::");
-            for (name, mode) in self.modes_of(id) {
-                self.rustdoc(&docs);
-                self.push_str(&format!("pub type {name}"));
-                self.print_generics(mode.lifetime);
-                self.push_str(" = ");
-                self.push_str(&path);
-                self.print_tyid(eq_ty, mode);
-                self.push_str(";\n");
-            }
-        } else {
-            let docs = Docs {
-                contents: Some("wit-bindgen: alias to equal type".to_string()),
-            };
-            self.print_typedef_alias(id, &Type::Id(eq_ty), &docs);
-        }
-    }
-
     fn print_typedef_alias(&mut self, id: TypeId, ty: &Type, docs: &Docs) {
         for (name, mode) in self.modes_of(id) {
             self.rustdoc(docs);
@@ -2607,6 +2575,16 @@ impl<'a> wit_bindgen_core::InterfaceGenerator<'a> for InterfaceGenerator<'a> {
 
     fn type_record(&mut self, id: TypeId, _name: &str, record: &Record, docs: &Docs) {
         self.print_typedef_record(id, record, docs);
+    }
+
+    fn define_type(&mut self, name: &str, id: TypeId) {
+        let equal = self.r#gen.types.get_representative_type(id);
+        if equal == id {
+            wit_bindgen_core::define_type(self, name, id)
+        } else {
+            let docs = &self.resolve.types[id].docs;
+            self.print_typedef_alias(id, &Type::Id(equal), &docs);
+        }
     }
 
     fn type_resource(&mut self, _id: TypeId, name: &str, docs: &Docs) {
--- a/crates/rust/src/lib.rs
+++ b/crates/rust/src/lib.rs
@@ -1025,42 +1025,6 @@ macro_rules! __export_{world_name}_impl {{
             .async_
             .is_async(resolve, interface, func, is_import)
     }
-
-    // Returns the structurally equal type id if exists. If the equal type comes from the
-    // import of the same interface, also returns the interface key, so that we can generate
-    // a type alias to the import type.
-    fn get_equal_type_alias<'a>(
-        &mut self,
-        resolve: &Resolve,
-        iface_key: Option<&'a WorldKey>,
-        ty_id: TypeId,
-    ) -> Option<(TypeId, Option<&'a WorldKey>)> {
-        if !self.opts.merge_structurally_equal_types() {
-            return None;
-        }
-        let ty = &resolve.types[ty_id].kind;
-        if matches!(ty, TypeDefKind::Type(_)) {
-            // preserve all primitive type and type alias definitions
-            return None;
-        }
-        let root = self.types.get_representative_type(ty_id);
-        if root != ty_id {
-            Some((root, None))
-        } else {
-            let TypeOwner::Interface(iface_id) = resolve.types[ty_id].owner else {
-                unreachable!()
-            };
-            // When we allow importing/exporting the same interface multiple times, we need to update this code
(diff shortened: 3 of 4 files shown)


crates/rust/src/lib.rs
```diff
@@ -973,6 +982,42 @@ macro_rules! __export_{world_name}_impl {{
...
+        if root != ty_id {
+            Some((root, None))
+        } else {
+            let TypeOwner::Interface(iface_id) = resolve.types[ty_id].owner else {
+                unreachable!()
+            };
+            // When we allow importing/exporting the same interface multiple times, we need to update this code
+            if !self.types.get(ty_id).has_resource
+                && iface_key.is_some()
+                && let Some(true) = self.interface_last_seen_as_import.get(&iface_id)
```
alexcrichton: It might be my memory failing me but I'm not entirely sure what these clauses are doing. Could you expand the comment to explain more what case this is handling?

crates/rust/src/interface.rs
```diff
@@ -2326,6 +2326,38 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
         }
     }
 
+    pub fn type_alias_to_eqaul_type(
```
alexcrichton: s/eqaul/equal/

crates/core/src/types.rs
```diff
@@ -228,4 +246,132 @@ impl Types {
...
+            (Type::Id(a), Type::Id(b)) => {
+                let a_def = &resolve.types[*a].kind;
+                let b_def = &resolve.types[*b].kind;
+                if self.is_resource_like_type(a_def) || self.is_resource_like_type(b_def) {
+                    return false;
+                }
+                self.equal_types.find(*a) == self.equal_types.find(*b)
+            }
+            (Type::ErrorContext, Type::ErrorContext) => todo!(),
+            _ => a == b,
```
alexcrichton: Personally I prefer to keep things exhaustive where possible to avoid `_ => { /* ... */ }` catch-call cases (same with `_ => false` above)

Could this be changed to soemthing like:

```rust
(Type::Id(a), Type::Id(b)) => { /* logic */ }
(Type::Id(_), _) => false,
(a @ Type::{a, b, c, d}, b) => a == b,
```

That way adding new types is forced to update this to ensure it handles new types appropriately.

crates/core/src/types.rs
```diff
@@ -228,4 +246,132 @@ impl Types {
...
+    fn optional_types_equal(
+        &mut self,
+        resolve: &Resolve,
+        a: &Option<Type>,
+        b: &Option<Type>,
+    ) -> bool {
+        match (a, b) {
+            (Some(a), Some(b)) => self.types_equal(resolve, a, b),
+            (None, None) => true,
+            _ => false,
```
alexcrichton: Similar to above, can the `_` case be filled out to make this `match` exhaustive?

crates/core/src/types.rs
```diff
@@ -228,4 +246,132 @@ impl Types {
...
+        }
+    }
+    fn is_resource_like_type(&self, ty: &TypeDefKind) -> bool {
```
alexcrichton: I think this might work better if this function didn't exist and it were folded directly into `types_equal` or `is_structurally_equal`. In a sense resources shouldn't be handled any differently than anything else, it's just that if they have different origin ids then they're different. I think that should fit into the rest of the logic here pretty well.

crates/core/src/types.rs
```diff
@@ -228,4 +246,132 @@ impl Types {
...
+        match ty {
+            TypeDefKind::Resource | TypeDefKind::Handle(_) => true,
+            TypeDefKind::Future(_) | TypeDefKind::Stream(_) => true,
```
alexcrichton: This part can be safely omitted since futures/streams are safe to deduplicate. It's just resources that can't be deduplicated.

crates/rust/src/lib.rs
```diff
@@ -274,6 +274,15 @@ pub struct Opts {
...
     pub async_: AsyncFilterSet,
+
+    /// Find all structurally equal types and only generate one type definition for
+    /// each equivalence class. Other types in the same class will be type aliases to the
+    /// generated type. This avoids clone when converting between types that are
+    /// structurally equal, which is useful when import and export the same interface.
+    ///
+    /// Types containing resource, future, or stream are never considered equal.
+    #[cfg_attr(feature = "clap", arg(long))]
+    pub merge_structurally_equal_types: bool,
```
alexcrichton: This I think would be reasonable to turn on by default, but are you hesitant to enable it by default?

crates/core/src/types.rs
chenyan2002: I'm not very familiar with futures/streams, with this WIT
```
interface A {
  record R1 { a: stream };
  record R2 { a: stream };
}
interface B {
  record R1 { a: stream };
  record R2 { a: stream };
}
world root {
  import A;
  export B;
}
```
Can `A.R1` and `A.R2` be merged? What about `A.R1` and `B.R1`? Also `self.types.has_resource` includes both resource, future and stream types, which suggests they are all nominal types?

crates/core/src/types.rs
chenyan2002: > if they have different origin ids then they're different.

Not quite. It also depends on how the world import/export these interfaces.
```
interface A {
  resource R;
}
interface B {
  use A.{R};
}
```
If we `import A; import B;`, then `A.R` is equal to `B.R`. If we `import A; export B;`, then `A.R` is not equal to `B.R`. 

In the future, if we allow `import B; import B as B1;` (which is possible in WAT already), then `B.R` is not equal to `B1.R`.

It would be great if we can specify the resource equality precisely in `types_equal`, but since it depends on the world and its transitive dependencies, it feels complicated. This function is trying to be conservative, and say that any types that contain resources are not equal. So that we don't need to handle all these tricky cases.

crates/rust/src/lib.rs
chenyan2002: This branch means that we are processing an interface that has been imported before. With the current WIT parser, it means we are importing and exporting the same interface. For export interface, if we can't find an equal type, we can still alias the type to the imported interface (for non-resource types).

crates/rust/src/lib.rs
chenyan2002: Yeah, sure. Happy to make it on by default. Just being cautious here, not to change the production behavior.

crates/core/src/types.rs
alexcrichton: Yes, all of those types are safe to merge. The `has_resource` bit might be for destructors and/or "needs cleanup", but I sort of forget... From the perspective of the component model and generated bindings those all of the above generate the same type.

crates/core/src/types.rs
alexcrichton: For `import A; export B;`, in that case `R` is actually the same in both cases. The export transitively refers to the import (extra exports aren't injected).

For supporting `import B as B1` that'll require pretty invasive changes, in my opinion, to `wit-parser` to actually support it. Ideally if you did `import A; export A` today that would result in two separate `InterfaceId`s (and `TypeId`s internally) for the two interfaces but it currently doesn't do that.

I agree the way things are handled today is weird since `import A; export A;` uses the same `TypeId`s which isn't great. If that's the rationale for keeping it the way it is, mind leaving a comment which explains that this should in theory be fancier but it's hard today?

crates/rust/src/lib.rs
alexcrichton: Ah yeah that's find with semver and such, and this is something that I've heard others asking for and I don't know why it wouldn't be the default, so I think it's reasonable to have it on-by-default.

crates/rust/src/lib.rs
alexcrichton: Makes sense yeah, mind expanding the comment here with some of that info?

alexcrichton:
As a heads up @chenyan2002 I'm interested in getting this PR merged to assist with fixing https://github.com/bytecodealliance/wit-bindgen/issues/1523. To that end I've done a few things to this PR:

* Merged it with `main` (no conflicts)
* Merged it with #1526 (semantically needed for subsequent commits)
* Addressed some of my review feedback (e.g. removing `_ => false` catch-all claues, moving things around, etc)
* Refactored the internals to simplify the generator after #1526 

I plan on redoing the merge after https://github.com/bytecodealliance/wit-bindgen/pull/1526 lands to have only one merge commit instead of two, and then I plan on merging this. For now I'm going to leave this flag disabled-by-default, but it's set up to make this a more easily toggle-able flag in the future if need be.

alexcrichton:
I'm going to go ahead and flag this for merging, but @chenyan2002 feel free to ping me if anything here looks awry and/or this doesn't end up working for your use case given my changes. Happy to help fix!

crates/core/src/types.rs
```diff
@@ -233,4 +250,185 @@ impl Types {
...
+                    self.is_structurally_equal(resolve, *a, *b)
+                }
+                (Handle::Own(_) | Handle::Borrow(_), _) => false,
+            },
+            (TypeDefKind::Handle(_), _) => false,
+            (TypeDefKind::Unknown, _) => unreachable!(),
+
+            // TODO: for now consider all resources not-equal to each other.
+            // This is because the same type id can be used for both an imported
+            // and exported resource where those should be distinct types.
```
chenyan2002, line 343: With #1526, is this comment out-dated? This branch means resources with different typeId are not equal.

crates/rust/src/lib.rs
```diff
@@ -274,6 +274,18 @@ pub struct Opts {
     #[cfg_attr(feature = "clap", clap(flatten))]
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub async_: AsyncFilterSet,
+
+    /// Find all structurally equal types and only generate one type definition for
+    /// each equivalence class. Other types in the same class will be type aliases to the
+    /// generated type. This avoids clone when converting between types that are
+    /// structurally equal, which is useful when import and export the same interface.
+    ///
+    /// Types containing resource, future, or stream are never considered equal.
```
chenyan2002, line 283:
```suggestion
```

crates/core/src/types.rs
alexcrichton: Oops yes indeed! Got the PR ordering mixed up and forgot to come back to this...

PR #1469, opened 2025-12-16
C++: Support separation of imported interfaces to libraries

cpetig:
This introduces a `--with` parameter for C++, instead of qualified type (like Rust) it works by specifying an include path.


cpetig:
This turned out to be surprisingly simple, I have worked around this limitation for a good year :shrug:

alexcrichton (approved):

Issue #1432, opened 2025-11-24
Conflicting implementations of `StreamPayload`


alexcrichton:
This:

```rust
wit_bindgen::generate!({
    inline: r#"
        package a:b;

        world w {
            import i;
            export i;
        }

        interface i {
            use t.{r};
            f: func(p: stream<r>);
        }

        interface t {
            record r {
                x: u32,
            }
        }
    "#,
});

fn main() {}
```

fails with:

```
$ cargo build --target wasm32-wasip1
   Compiling wat v0.1.0 (/home/alex/code/wat)
error[E0119]: conflicting implementations of trait `StreamPayload` for type `t::R`
  --> src/main.rs:1:1
   |
 1 | / wit_bindgen::generate!({
 2 | |     inline: r#"
 3 | |         package a:b;
...  |
20 | |     "#,
21 | | });
   | |  ^
   | |  |
   | |__first implementation here
   |    conflicting implementation for `t::R`
   |
   = note: this error originates in the macro `wit_bindgen::generate` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0119`.
error: could not compile `wat` (bin "wat") due to 1 previous error
```

alexcrichton:
Fixed in #1482

PR #1482, opened 2025-12-28
wit-bindgen: generate only one StreamPayload or FuturePayload implementation per type alias set

danielvallance:
WIT 'use' statements translate to Rust type aliases in the generated Rust bindings.

Since these aliases may be located at different module paths, creating a StreamPayload or FuturePayload implementation for more than one of these paths will cause the Rust compiler to complain about conflicting trait implementations for the same type.

This commit allows only one StreamPayload or FuturePayload implementation to be generated for a WIT type which will get transpiled into a Rust type alias, and it records mappings of the WIT type to the canonical module path to the type alias, which allows equivalent type aliases to reuse the implementation.

Fixes issue https://github.com/bytecodealliance/wit-bindgen/issues/1432

### DIFF
--- a/crates/rust/src/bindgen.rs
+++ b/crates/rust/src/bindgen.rs
@@ -427,13 +427,19 @@ impl Bindgen for FunctionBindgen<'_, '_> {
             Instruction::FutureLift { payload, .. } => {
                 let async_support = self.r#gen.r#gen.async_support_path();
                 let op = &operands[0];
-                let name = payload
-                    .as_ref()
-                    .map(|ty| {
-                        self.r#gen
-                            .type_name_owned_with_id(ty, Identifier::StreamOrFuturePayload)
-                    })
-                    .unwrap_or_else(|| "()".into());
+                let name = match payload {
+                    Some(Type::Id(type_id)) => {
+                        let dealiased_id = dealias(resolve, *type_id);
+                        self.r#gen.type_name_owned_with_id(
+                            &Type::Id(dealiased_id),
+                            Identifier::StreamOrFuturePayload,
+                        )
+                    }
+                    Some(ty) => self
+                        .r#gen
+                        .type_name_owned_with_id(ty, Identifier::StreamOrFuturePayload),
+                    None => "()".into(),
+                };
                 let ordinal = self
                     .r#gen
                     .r#gen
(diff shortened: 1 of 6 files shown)


alexcrichton (commented):
Thanks for the PR! I'm a bit wary of adding yet-more-layers-of-maps though, especially when there's already a map of `stream_payloads`. Would you be up for doing some extra refactoring to remove the need for these maps? For example naively I'd expect that the `stream_payloads` map could be refactored to be indexed by `TypeId` rather than `String`. Callers would then be sure to call `dealias` to ensure that the only types inserted/looked up would be un-aliased forms to solve the original bug

danielvallance:
hello, thanks for taking a look! yeah I agree that does sound like a cleaner solution, it sounds quite similar to a previous PR I made for this issue (https://github.com/bytecodealliance/wit-bindgen/pull/1478) in which I index the stream_payloads map with an Option < Type >  however I declined this due to CI failures

those CI failures happened because the same wit Type/TypeId was referred to at different paths, and since in this case it was not a TypeDefKind::Type, it resulted in the generation of identical structs at different paths which the compiler considered different types. and since the wit Type was used to index the stream_payloads map, it could only contain a reference to one of the several identical struct types generated from that wit Type, and in some cases it was the wrong one

that sort of led me to think I had to create this new map to handle the cases where a type alias was generated, and in the other cases, use stream_payloads as before

i did also have the thought of generating type aliases instead of regenerating multiple identical struct types at different paths for the same wit Type, however I felt that was quite a large change which could have unforseen consequences. however i may mark this as a draft and investigate this option further for now, do you think that sounds feasible?

alexcrichton:
Ah yeah ok I can see where that would add complications. I believe that's due to the fact that if an interface is both imported and exported that uses the same `TypeId` (regrettably) internally so there's one `TypeId` to refer to two different Rust types which causes problems.

What you describe sounds like it might work yeah, although another possible solution would be to invert things a bit. The trait implementation could be generated when the type is defined, for example. This would mean that a pass would be done over the WIT to determine the set of types that are used in futures/streams, and then when generating types if the type is in that set the trait impl is generated. That I think would work in the fact of import/export and avoid the need to work with aliases too much? (insertion into the set of futures/streams would be where the de-aliasing happens)

danielvallance:
actually yeah that's probably a better way of going about it - thanks for the pointer! i will mark this as draft while i look into doing it this way

PR #1487, opened 2026-01-06
feat: linting

### LINKED ISSUE #1486: Linting
Would y'all be open to me adding clippy to the actions PR workflow?

asteurer:
Closes #1486


crates/test/src/lib.rs
```diff
@@ -555,7 +555,7 @@ impl Runner<'_> {
             })
             .collect::<Vec<_>>();
 
-        println!("");
+        println!();
```
alexcrichton: It looks like there's a lot of changes in this PR (this included) which are caused by perhaps the default set of clippy lints rather than the enabled set of clippy lints. It's not problematic per se but some of them I don't totally agree with. Would you be ok reverting to a clean slate, adding the clippy config, and then only fixing lints that have warnings fire? That would end up backing out some lints but if you feel the changes should be kept it's ok to enable more lints in `Cargo.toml` for that.

asteurer: Oh, duh. Happy to do that!

.github/actions/install-rust/action.yml
alexcrichton, line 1: Oh most of this is specific to wasm-tools/wasmtime/etc, and isn't necessary here. For this clippy job it's fine to do the same `rustup update ...` that the other jobs are doing. The main thing from this file is to set `RUSTFLAGS="-D warnings"` but that can be done manually on the one clippy job.

src/bin/wit-bindgen.rs
```diff
@@ -160,10 +160,11 @@ fn main() -> Result<()> {
             Some(path) => path.join(name),
             None => name.into(),
         };
-        eprintln!("Generating {:?}", dst);
+        eprintln!("Generating {}", dst.display());
 
         if opt.check {
-            let prev = std::fs::read(&dst).with_context(|| format!("failed to read {:?}", dst))?;
+            let prev =
+                std::fs::read(&dst).with_context(|| format!("failed to read {}", dst.display()))?;
```
alexcrichton: How come these changed? Rendering `:?` vs `path.display()` is pretty different and while it's generally fine either way I'm not sure why the lints are firing for this crate as it shouldn't be enabled

crates/test-helpers/src/lib.rs
```diff
@@ -26,7 +26,7 @@ pub fn test_directory(suite_name: &str, gen_name: &str, wit_name: &str) -> PathB
 
     drop(fs::remove_dir_all(&me));
     fs::create_dir_all(&me).unwrap();
-    return me;
+    me
```
alexcrichton: I think that this lint (and many other perhaps?) are due to a missing `[lints] workspace = true` for this crate which means that more lints are enabled. Could you go through and ensure that it's annotated for all crates in the workspace to ensure that the same set of clippy lints are used?

asteurer:
Thanks for your patience while I learned how to use clippy :sweat_smile:

PR #1489, opened 2026-01-06
fix(csharp): Add runtime initialization guard for reactor components

dirkwa:
## Summary

This PR adds initialization guard infrastructure to the C# wit-bindgen code generator for reactor (library) components. It ensures `__wasm_call_ctors` is called before any exported function executes.

## Problem

When building .NET WASM components as libraries (reactors without a `_start` entry point), calling exports directly fails with:

**wasmtime:**
wasm trap: uninitialized element wasm backtrace: 0: InitializeGlobalTablesForModule

**jco/Node.js:**
RuntimeError: null function or function signature mismatch at InitializeGlobalTablesForModule

## Root Cause Analysis

The issue involves a circular dependency in the .NET NativeAOT-LLVM runtime:

1. Reactor components have no `_start` to trigger runtime initialization
2. When exports are called, `RhpReversePInvoke` attempts to initialize the runtime
3. But `RhpReversePInvoke` itself requires initialized function tables
4. This causes a crash during initialization, not before it

## This PR's Contribution

This PR adds the **correct initialization pattern** at the wit-bindgen layer:

```csharp
internal static class WasmInteropInitializer
{
    private static volatile bool _initialized;
    private static readonly object _lock = new object();

    [DllImport("*", EntryPoint = "__wasm_call_ctors")]
    private static extern void WasmCallCtors();

    internal static void EnsureInitialized()
    {
        if (_initialized) return;
        lock (_lock)
        {
            if (_initialized) return;
            WasmCallCtors();
            _initialized = true;
        }
    }
}
```

Each generated export function calls EnsureInitialized() first:

```csharp
[UnmanagedCallersOnly(EntryPoint = "test:minimal/plugin@1.0.0#get-id")]
public static unsafe nint wasmExportGetId() {
    MinimalWorld.WasmInteropInitializer.EnsureInitialized();
    // ... rest of export code
}
```

## Limitations
This fix alone does not fully resolve the issue. The complete solution requires changes in:
- NativeAOT-LLVM runtime: To properly handle reactor initialization where RhpReversePInvoke is called before tables are populated
- Or componentize-dotnet/wasm-tools adapter: To ensure _initialize is called before any exports

This PR establishes the correct pattern so that when the runtime-level fix is implemented, the initialization will be properly triggered.

## Testing
- Verified generated code includes WasmInteropInitializer class
- Verified EnsureInitialized() is called in all export functions
- Confirmed the crash now occurs during initialization (progress from "no init attempted")

## Related Issues
- https://github.com/bytecodealliance/componentize-dotnet/issues/103
- https://github.com/bytecodealliance/jco/issues/1173

Both issues report the same root cause - .NET reactor components crash when exports are called directly because runtime initialization hasn't occurred properly.

### DIFF
--- a/crates/csharp/src/interface.rs
+++ b/crates/csharp/src/interface.rs
@@ -537,12 +537,14 @@ impl InterfaceGenerator<'_> {
         };
 
         let access = self.csharp_gen.access_modifier();
+        let qualifier = self.csharp_gen.qualifier();
 
         uwrite!(
             self.csharp_interop_src,
             r#"
             [global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "{export_name}")]
             {access} static unsafe {wasm_result_type} {interop_name}({wasm_params}) {{
+                {qualifier}WasmInteropInitializer.EnsureInitialized();
                 {vars}
                 {src}
             }}
--- a/crates/csharp/src/world_generator.rs
+++ b/crates/csharp/src/world_generator.rs
@@ -41,6 +41,7 @@ pub struct CSharp {
     pub(crate) all_resources: HashMap<TypeId, ResourceInfo>,
     pub(crate) world_resources: HashMap<TypeId, ResourceInfo>,
     pub(crate) import_funcs_called: bool,
+    pub(crate) has_exports: bool,
 }
 
 impl CSharp {
@@ -182,6 +183,7 @@ impl WorldGenerator for CSharp {
         id: InterfaceId,
         _files: &mut Files,
     ) -> anyhow::Result<()> {
+        self.has_exports = true;
         let name = interface_name(self, resolve, key, Direction::Export);
         self.interface_names.insert(id, name.clone());
         let mut r#gen = self.interface(resolve, &name, Direction::Export);
@@ -226,6 +228,7 @@ impl WorldGenerator for CSharp {
         funcs: &[(&str, &Function)],
         _files: &mut Files,
     ) -> anyhow::Result<()> {
+        self.has_exports = true;
         let name = &format!("{}-world", resolve.worlds[world].name).to_upper_camel_case();
         let name = &format!("{name}.I{name}");
         let mut r#gen = self.interface(resolve, name, Direction::Export);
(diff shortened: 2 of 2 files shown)


jsturtevant:
I will need to look a bit closer later, but this looks to be similar to https://github.com/bytecodealliance/wit-bindgen/issues/777.  fyi @yowl

jsturtevant:
@dirkwa looking at ci failures it seems some of the examples need to bring in the proper imports for this new functionality or it needs to be added to the import list in the binding generator

yowl:
Is there a repo or a stack trace?

dirkwa:
> Is there a repo or a stack trace?  

See here:

https://github.com/bytecodealliance/jco/issues/1173#issuecomment-3716703996

yowl:
So would this be fixed by moving to wasi SDK 29 ?

yowl:
Note that NAOT-LLVM upgraded to wasi sdk 29 today, so the nugets for that should be out in a few hours.

dirkwa:
> So would this be fixed by moving to wasi SDK 29 ?

I hope so, if the commit will be accepted by the maintainers. 

## Nevertheless this PR has a reason. 

If you look at the full https://github.com/bytecodealliance/jco/issues/1173 there is my test-repo to locally verify it. With the workaround in the comment I was able to sucessfully build it.

jsturtevant:
> > So would this be fixed by moving to wasi SDK 29 ?
> 
> I hope so, if the commit will be accepted by the maintainers.
> 
> ## Nevertheless this PR has a reason.
> If you look at the full [bytecodealliance/jco#1173](https://github.com/bytecodealliance/jco/issues/1173) there is my test-repo to locally verify it. With the workaround in the comment I was able to sucessfully build it.

Could we add an test example based on that test repo that demonstrates this issue and fixes?

It says in the attached issue:

> Keep it open - The EnsureInitialized() call added in https://github.com/bytecodealliance/wit-bindgen/pull/1489 is still valuable as a defense-in-depth measure. Once _initialize runs via the start-shim, the check becomes a fast no-op. It provides safety in case the component is used in an environment that doesn't honor the component model's initialization semantics.

What is an example of an environment that can run wasm components but would not have the component models initialization semantics?  Wouldn't that would be a bug in the runtime?

PR #1490, opened 2026-01-07
feat(go): preventing concurrent reads/writes on streams and futures

### LINKED ISSUE #1460: go: prevent concurrent reads or writes on the same stream or future
There's currently nothing in the code generated for Go to prevent the programmer from issuing multiple concurrent reads or writes on the same stream or future, which is currently disallowed by the component model spec and will lead to a trap.  We should handle this more gracefully by returning an idiomatic error from `{Stream,Future}Writer.Write` and `{Stream,Future}Reader.Read` (or else a panic if we feel this is a programmer error rather than a runtime error).

asteurer:
# Overview
This change is intended to prevent concurrent reads and writes on streams and futures

Closes #1460

## Observations
Given the following guest application code:
```go
func Handle(request *Request) Result[*Response, ErrorCode] {
	tx, rx := MakeStreamU8()

	go func() {
		defer tx.Drop()
		for i := 0; i < 100; i++ {
			tx.Write([]byte(fmt.Sprintf("Hello %d", i)))
			time.Sleep(1 * time.Millisecond)
		}
	}()

	var wg sync.WaitGroup
	for i := 0; i < 3; i++ {
		wg.Add(1)
		go func(id int) {
			defer wg.Done()
			buffer := make([]uint8, 7)
			for j := 0; j < 10; j++ {
				// ERROR: READ CONCURRENTLY
				count := rx.Read(buffer)
				if count == 0 {
					break
				}

				fmt.Printf("output: %s", string(buffer))
			}
		}(i)
	}

	wg.Wait()

	return Err[*wasi_http_handler.Response](MakeErrorCodeDestinationIpUnroutable())

}
```

I would expect there to be a panic and have there be a `"nil handle"` message; however, this is the output I get:
```
worker error: error while executing at wasm backtrace:
    0: 0x2916ef - <unknown>!<wasm function 16>
    1: 0x182eb0 - <unknown>!wit_component_wasi_http_types.wasm_stream_read_u8
    2: 0x184933 - <unknown>!github.com_bytecodealliance_wit_bindgen_wit_types.__StreamReader_go.shape.uint8__.Read
    3: 0x1890b5 - <unknown>!wit_component_export_wasi_http_handler.Handle.func2
    4: 0x188d81 - <unknown>!wit_component_export_wasi_http_handler.Handle.gowrap1
    5: 0x12204b - <unknown>!wasm_pc_f_loop_export
    6: 0x18c0d1 - <unknown>!_async_lift_wasi_http_handler_0.3.0_rc_2025_09_16_handle
error: channel closed
```

This is concerning because the Goroutine appears to be moving past the panic in the `wit_runtime.Handle.Take()` method:
```go
func (self *Handle) Take() int32 {
	if self.value == 0 {
		panic("nil handle")
	}
	value := self.value
	self.value = 0
	return value
}
```

@dicej any thoughts and/or suggestions?

### DIFF
--- a/crates/go/src/package/wit_types/wit_future.go
+++ b/crates/go/src/package/wit_types/wit_future.go
@@ -26,6 +26,11 @@ type FutureReader[T any] struct {
 	handle *wit_runtime.Handle
 }
 
+// Blocks until the future completes and returns its value.
+//
+// # Panic
+//
+// Read will panic if multiple concurrent or sequential reads are attempted on the same future.
 func (self *FutureReader[T]) Read() T {
 	handle := self.handle.Take()
 	defer self.vtable.DropReadable(handle)
@@ -53,6 +58,7 @@ func (self *FutureReader[T]) Read() T {
 	}
 }
 
+// Notify the host that the FutureReader is no longer being used.
 func (self *FutureReader[T]) Drop() {
 	handle := self.handle.TakeOrNil()
 	if handle != 0 {
@@ -85,6 +91,11 @@ type FutureWriter[T any] struct {
 	handle *wit_runtime.Handle
 }
 
+// Writes data to a future.
+//
+// # Panic
+//
+// Write will panic if multiple concurrent or sequential writes are attempted on the same future.
 func (self *FutureWriter[T]) Write(item T) bool {
 	handle := self.handle.Take()
 	defer self.vtable.DropWritable(handle)
@@ -119,6 +130,7 @@ func (self *FutureWriter[T]) Write(item T) bool {
 	}
 }
 
+// Notify the host that the FutureWriter is no longer being used.
 func (self *FutureWriter[T]) Drop() {
 	handle := self.handle.TakeOrNil()
 	if handle != 0 {
--- a/crates/go/src/package/wit_types/wit_stream.go
+++ b/crates/go/src/package/wit_types/wit_stream.go
@@ -31,8 +31,22 @@ func (self *StreamReader[T]) WriterDropped() bool {
 	return self.writerDropped
 }
 
+// Reads data from a stream into a destination slice.
+//
+// Blocks until the read completes or the destination slice is full.
+//
+// # Panic
+//
+// Read will panic if:
+//   - dst is empty (length 0)
+//   - multiple concurrent reads are attempted on the same stream
 func (self *StreamReader[T]) Read(dst []T) uint32 {
-	handle := self.handle.Use()
+	if len(dst) == 0 {
+		panic("StreamReader.Read: destination slice cannot be empty")
+	}
+
+	handle := self.handle.Take()
+	defer self.handle.Set(handle)
 
 	if self.writerDropped {
 		return 0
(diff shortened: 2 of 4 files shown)


dicej:
> ```go
> time.Sleep(1 * time.Millisecond)
> ```

`time.Sleep` will block the entire runtime since there's no `GOOS=wasip3` yet.  You'll need to use `wasi:clocks@0.3.0-rc- 2025-09-26/monotonic-clock#wait-for` instead (i.e. generating bindings from the WASIp3 WIT files and using those directly).

crates/go/src/package/wit_types/wit_future.go
```diff
@@ -28,6 +28,7 @@ type FutureReader[T any] struct {
 
 func (self *FutureReader[T]) Read() T {
 	handle := self.handle.Take()
+	defer self.handle.Set(handle)
```
dicej: `FutureReader.Read` should permanently consume the handle, so we shouldn't be calling `Handle.Set` here.

crates/go/src/package/wit_types/wit_future.go
```diff
@@ -87,6 +88,7 @@ type FutureWriter[T any] struct {
 
 func (self *FutureWriter[T]) Write(item T) bool {
 	handle := self.handle.Take()
+	defer self.handle.Set(handle)
```
dicej: As with `FutureReader.Read`, we shouldn't be calling `Handle.Set` here.

asteurer:
@dicej Looks like the CI/Check GH Actions workflow failed. I assume the fix for that is just re-running it? If so, would you mind doing that for me?

dicej:
> @dicej Looks like the CI/Check GH Actions workflow failed. I assume the fix for that is just re-running it? If so, would you mind doing that for me?

👍 

BTW, would you be able to add test coverage for this to the tests/runtime-async suite, perhaps by expanding `simple-stream-payload`?

asteurer:
@dicej When you have a moment, this is ready to be reviewed

crates/go/src/package/wit_types/wit_future.go
```diff
@@ -26,6 +26,11 @@ type FutureReader[T any] struct {
 	handle *wit_runtime.Handle
 }
 
+// Blocks until the future completes and returns its value.
+//
+// # Panic
+//
+// Read will panic if multiple concurrent reads are attempted on the same future.
```
dicej:
```suggestion
// Read will panic if multiple concurrent or sequential reads are attempted on the same future.
```

crates/go/src/package/wit_types/wit_future.go
```diff
@@ -85,6 +91,11 @@ type FutureWriter[T any] struct {
 	handle *wit_runtime.Handle
 }
 
+// Writes data to a future.
+//
+// # Panic
+//
+// Write will panic if multiple concurrent writes are attempted on the same future.
```
dicej:
```suggestion
// Write will panic if multiple concurrent or sequential writes are attempted on the same future.
```

crates/go/src/package/wit_types/wit_stream.go
```diff
@@ -31,8 +31,22 @@ func (self *StreamReader[T]) WriterDropped() bool {
 	return self.writerDropped
 }
 
+// Reads data from a stream into a destination slice.
+//
+// Blocks until the stream completes or the destination slice is full.
```
dicej:
```suggestion
// Blocks until the read completes or the destination slice is full.
```

PR #1491, opened 2026-01-08
feat(c): Add co-op threading builtins to C generator

TartanLlama:
Adds the co-op threading builtins for the component model to the C generator. This is needed for work on [wasip3](https://github.com/WebAssembly/wasi-libc/issues/691) in `wasi-libc`.

Some design decisions we may want to go a different way on:

- Introduces two breaking changes:
  - `context_get/set` now take an additional parameter indicating which context slot to operate on. We could maintain the old versions, but I think this will be an easy fix for anyone who is using them.
  - `<world>_yield` has been renamed `<world>_thread_yield` for parity with the other builtins. I think this is a similarly easy fix for users.
- Does not put the new builtins behind a flag: if any async functions are required, the new builtins are also included. The knock-on effect is that users of async with wasmtime would need to additionally specify `-W component-model-threading -W component-model-async-stackful`, or they would see instantiation errors indicating to add those flags. I'm happy to split them out if we think that's the best approach.
- Added a new `--generate-async-helpers` flag to force generation of the async helpers in situations where you don't have any futures or functions marked `async`, but you still want the threading builtins. 

Rust support can be done in a follow-up PR; I'm not familiar with the design of the Rust guest generator, so focused on this part that is needed for `wasi-libc`.

### DIFF
--- a/tests/runtime-async/async/threading-builtins/test.c
+++ b/tests/runtime-async/async/threading-builtins/test.c
@@ -1,5 +1,6 @@
 //@ args = '--rename a:b/i=test --generate-threading-helpers'
 //@ [lang]
+//@ cflags = "-O2"
 //@ ldflags = "-Wl,--export-table"
 
 #include <assert.h>


crates/test/src/c.rs
```diff
@@ -139,6 +139,7 @@ fn compile(runner: &Runner<'_>, compile: &Compile<'_>, compiler: PathBuf) -> Res
         .arg("-Wc++-compat")
         .arg("-Wno-unused-parameter")
         .arg("-g")
+        .arg("-Wl,--export-table")
```
alexcrichton: Could this be moved to the test in question? You can put `//@ cflags = '-O'` at the top of the file for example

crates/c/src/lib.rs
```diff
@@ -847,32 +860,104 @@ void {snake}_backpressure_dec(void) {{
...
-void* {snake}_context_get() {{
-    return __context_get();
+__attribute__((__import_module__("$root"), __import_name__("[context-get-1]")))
+extern void* __context_get_1(void);
+
+void* {snake}_context_get(size_t index) {{
+    if (index == 0) return __context_get_0();
+    if (index == 1) return __context_get_1();
+    return NULL;
 }}
```
alexcrichton: Personally I'd say that we should probably drop the `context_get` helper and instead only support `context_get_{0,1}()` since callers effectively need to choose which themselves

crates/c/src/lib.rs
```diff
@@ -847,32 +860,104 @@ void {snake}_backpressure_dec(void) {{
...
 
 __attribute__((__import_module__("$root"), __import_name__("[cancellable][thread-yield]")))
 extern uint32_t __thread_yield_cancellable(void);
 
-uint32_t {snake}_yield_cancellable(void) {{
+uint32_t {snake}_thread_yield_cancellable(void) {{
     return __thread_yield_cancellable();
 }}
+
+__attribute__((__import_module__("$root"), __import_name__("[thread-index]")))
```
alexcrichton: Could this, and other thread related intrinsics, for now go behind an extra flag in the C generator? Similar to `generate_async_helpers` except focused just on these options. I'm mostly worried about generating code that won't validate in C when compiled unless it's compiled with `-ffunction-sections` which may not be all users.

tests/runtime-async/async/threading-builtins/test.c
alexcrichton, line 1: This test technically only "works by accident" today insofar as the global stack pointer should be getting extra clobbered here but happens to not because of the test not actually using the stack that much? Or, put another way, how is this test passing? I figured that in debug mode the stack would be used at least a little bit and would expect it to be pretty clobbered by this test.

crates/test/src/c.rs
TartanLlama: I'd have to add an `ldflags` option, but yeah!

crates/c/src/lib.rs
TartanLlama: Yeah, that makes sense to me

crates/c/src/lib.rs
TartanLlama: Yep, I've added a `--generate-threading-helpers` flag

tests/runtime-async/async/threading-builtins/test.c
TartanLlama: Indeed, it's certainly forseeable that code could be generated that would cause this test to fail on account of expecting no change to `__stack_pointer` and having it modified by the other thread. I'm not entirely sure of the best way to both test that these intrinsics all link correctly and `wasmtime` is happy with their signatures, without having to rely on "accidental" codegen. Maybe if we just use globals for those thread IDs that'll be enough for it to be unlikely to break?

alexcrichton: Even with globals I'd be afraid of local values being present on the stack for debuginfo or stuff like that. Maybe compile with optimizations for now and assume that's good enough?

PR #1492, opened 2026-01-09
Update wasm-tools dependencies

alexcrichton:
Keeping things up-to-date


pchickey:
Your editor has grown the same bad habit mine has, formtting code for the wrong rust edition...

PR #1498, opened 2026-01-12
feat(go): enable user to configure the package name for generated types

### LINKED ISSUE #1494: go: enable user to configure the package name for generated types
It would be really nice to be able to run a command like `wit-bindgen go --pkg github.com/foo/bar/v1` and have the generated types prepend the custom package to the `wit_component` path: 
```go
import "github.com/foo/bar/v1/wit_component/wasi_http_0_2_0_types"
```

As part of this, we will probably want to alter the `wit_bindings.go` file:
```go
// Instead of "package main"
package wit_bindgen
```

This would be helpful for cases where an SDK wants to make use of the bindings, but doesn't want to expose the bindings to a user (see [this version of the spin-go-sdk](https://github.com/asteurer/spin-go-sdk/blob/9cc21b464db81ce2f6e006423bb846e0d8936e65/v3/http/http.go) as an example).

If there's a better way to do this, I'm also open to other suggestions.

@dicej if this is something you would like to move forward with, will you assign this to me, please?

asteurer:
Closes #1494 

@dicej

### DIFF
--- a/crates/go/src/lib.rs
+++ b/crates/go/src/lib.rs
@@ -858,6 +858,15 @@ impl WorldGenerator for Go {
             "main"
         };
 
+        // If the package name isn't "main", a main function isn't required.
+        let main_func = if self.opts.mod_name.is_some() {
+            ""
+        } else {
+            r#"// Unused, but present to make the compiler happy
+func main() {}
+"#
+        };
+
         files.push(
             "wit_bindings.go",
             &maybe_gofmt(
@@ -877,8 +886,7 @@ var {SYNC_EXPORT_PINNER} = runtime.Pinner{{}}
 
 {src}
 
-// Unused, but present to make the compiler happy
-func main() {{}}
+{main_func}
 "#
                 )
                 .as_bytes(),


crates/go/src/lib.rs
```diff
@@ -843,13 +849,22 @@ impl WorldGenerator for Go {
...
 
+        // When a custom module name is provided, the generated bindings use package name
+        // "wit_component" so they can be imported into another package. Without a custom
+        // module name, the bindings use package "main" for standalone execution.
+        let package_name = if self.opts.mod_name.is_some() {
+            "wit_component"
+        } else {
+            "main"
+        };
+
```
asteurer: I'm concerned about this portion. If the user decides to create a module named `github.com/foo/bar/baz` and output the bindings in directory `baz`, there will be a mismatch between the package name `wit_component` and the directory name `baz`. Any thoughts and/or suggestions?

crates/go/src/lib.rs
dicej: I imagine we could remove this part if `package_name == "wit_component"` since the compiler only treats `main.main` as special AFAIK.

asteurer:
@dicej Just wanting to make sure that you saw [my earlier question](https://github.com/bytecodealliance/wit-bindgen/pull/1498/changes#r2684190467)?

dicej:
> @dicej Just wanting to make sure that you saw [my earlier question](https://github.com/bytecodealliance/wit-bindgen/pull/1498/changes#r2684190467)?

Sorry, I'm not seeing a question in that link -- it just takes me to a diff.

PR #1507, opened 2026-01-21
C# pending import

yowl:
This PR add support for C# for the async `pending-import` runtime test.  It moves the C# codegen away from minimalist `C` style to a more Go and Rust like approach.  This PR gets this next test passing on the understanding there is still lots to do for C# async codegen.  Apologies for the largeness of the PR, I hope the restructure is mostly complete here, and the next PR will be smaller.

### DIFF
--- a/crates/test/src/csharp.rs
+++ b/crates/test/src/csharp.rs
@@ -42,7 +42,7 @@ impl LanguageMethods for Csharp {
     ) -> bool {
         // TODO: remove these exclusions as support is created
         matches!(
-            name, 
+            name,
             "error-context.wit"
                 | "resource-fallible-constructor.wit"
                 | "async-resource-func.wit"


crates/csharp/src/AsyncSupport.cs
```diff
@@ -33,51 +47,140 @@ public static class AsyncSupport
...
+            Code = code;
+            WaitableStatus = new WaitableStatus(code & 0xf);
+        }
+
+        public readonly int Raw;
+        public readonly int Waitable;
+        public readonly uint Code;
+
+        public readonly EventCode EventCode => (EventCode)Raw;
+        public readonly WaitableStatus WaitableStatus;
```
dicej: This field seems redundant.  Could it use a closure like `EventCode` and `WaitableCount`?

crates/csharp/src/AsyncSupport.cs
```diff
@@ -33,51 +47,140 @@ public static class AsyncSupport
...
+    {
+        // TODO: Looks complicated....
+        if(PendingCallbacks.TryRemove((IntPtr)(contextPtr), out var tcs))
+        {
+            Marshal.FreeHGlobal((IntPtr)contextPtr);
+            taskReturn();
+
+            tcs.SetResult();
+        }
+        return CallbackCode.Exit;
```
dicej: FYI, we should only return `CallbackCode.Exit` here if the current task has nothing left to do.  This is fine to get a simple test passing, but we'll soon need to change this for more complicated tests.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -109,12 +225,114 @@ public static (FutureReader<T>, FutureWriter<T>) RawFutureNew<T>(FutureVTable vt
...
+        status = status & 0xF;
+
+        if(subTaskStatus.IsSubtaskStarting || subTaskStatus.IsSubtaskStarted)
+        {
+            if(state.WaitableSet == null) {
+                state.WaitableSet = AsyncSupport.WaitableSetNew();
+            }
+
+            // TODO join and complete the task somwhere.
+            TaskCompletionSource tcs = new TaskCompletionSource();
```
dicej, line 222: I'd expect we'd insert into `PendingCallbacks` here (and ditto for the generic version below).

crates/csharp/src/AsyncSupport.cs
```diff
@@ -128,36 +346,40 @@ public int TakeHandle()
     }
 
     // TODO: Generate per type for this instrinsic.
-    public Task Read()
+    public unsafe Task Read()
```
dicej: What are the implications of marking this `unsafe`?  I.e. what must the caller ensure to use it safely?

crates/csharp/src/function.rs
```diff
@@ -1175,7 +1147,24 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
                     
+                    AsyncSupport.ContextTask* contextTaskPtr = (AsyncSupport.ContextTask*)System.Runtime.InteropServices.Marshal.AllocHGlobal(sizeof(AsyncSupport.ContextTask));
+                    AsyncSupport.ContextSet(contextTaskPtr);
+
+                    // This TaskCompletionSource does nothing, but the callback code expects one.
+                    var tcs = new System.Threading.Tasks.TaskCompletionSource();
+                    AsyncSupport.PendingCallbacks.TryAdd((IntPtr)contextTaskPtr, tcs);
+
                     // TODO: Defer dropping borrowed resources until a result is returned.
                     return (uint)CallbackCode.Yield;
```
dicej: I'd expect we'd return `((uint)CallbackCode.Wait) | (contextTaskPtr.Set)` here or similar.  Returning `Yield` is basically a busy wait.

yowl:
I'm fine with whatever you think is best.  Do you want me to address the comments and pull this PR to get those 2 extra tests passing?

dicej:
> I'm fine with whatever you think is best. Do you want me to address the comments and pull this PR to get those 2 extra tests passing?

Yes, if it's not too much trouble.  That would give me more confidence we're heading in the right direction.

crates/csharp/src/AsyncSupport.cs
yowl: I will remove the unsafe here, it is no longer required.

yowl:
Closing and will open a new improved PR.

PR #1508, opened 2026-01-21
feat(go): migrate module to go.bytecodealliance.org/wit-bindgen

### LINKED ISSUE #1506: go: create better endpoint for packages directory
The [go-modules](https://github.com/bytecodealliance/go-modules) project uses an endpoint for their wit types that looks like this: `go.bytecodealliance.org/cm`. I'm wondering if we can create a similar endpoint for wit-bindgen's Go package and have it be something like `go.bytecodealliance.org/wit-bindgen`. This would mean we wouldn't need to add a replace directive in the mod file:

```diff
require (
-	github.com/bytecodealliance/wit-bindgen v0.0.0-00010101000000-000000000000
+       go.bytecodealliance.org/wit-bindgen v0.51.0
	github.com/julienschmidt/httprouter v1.3.0
)

- replace github.com/bytecodealliance/wit-bindgen => github.com/bytecodealliance/wit-bindgen/crates/go/src/package v0.51.0 
```

Also, I don't have super strong opinions on what the actual endpoint will be named. If there are other endpoints that folks like better than `/wit-bindgen`, please feel free to share!

asteurer:
Closes #1506

@dicej @tschneidereit


crates/go/src/package/wit_async/wit_async.go
```diff
@@ -5,7 +5,7 @@ import (
 	"runtime"
 	"unsafe"
 
-	"github.com/bytecodealliance/wit-bindgen/wit_runtime"
+	wit_runtime "go.bytecodealliance.org/wit-bindgen/wit_runtime"
```
dicej: Why is it necessary to provide an explicit `wit_runtime` name here.  Isn't that the default?

asteurer: Ope, not sure why I did that. Thanks for catching!

asteurer:
@dicej @tschneidereit I had an assumption that we would be able to redirect requests from `go.bytecodealliance.org` to `github.com/bytecodealliance/wit-bindgen/crates/go/src/package`. Is this actually possible, or will we need to discuss other paths forward?

crates/go/src/go.bytecodealliance.org/wit-bindgen.html
```diff
@@ -0,0 +1,6 @@
+<html>
+	<head>
+		<title>go.bytecodealliance.org/wit-bindgen</title>
+		<meta name="go-import" content="go.bytecodealliance.org git https://github.com/bytecodealliance/wit-bindgen/crates/go/src/package" />
```
asteurer, line 4: Will the git retrieval work from a nested directory like this?

tschneidereit:
@asteurer I unfortunately don't know how this works, apologies. One important thing I want to emphasize again is that it's crucial to ensure that existing code using `go-modules` continues working, so whatever the setup looks like in the end, it needs to ensure that current links continue to work as well.

asteurer:
Good to know! I'll convert this to a draft PR until I've found a way to properly test everything.

PR #1519, opened 2026-01-29
Component model threading intrinsic updates 

TartanLlama:
A draft PR for upcoming proposed changes to the component model threading intrinsics to support the `wasi-libc` `pthreads` implementation.


alexcrichton:
After the update to wasm-tools 245 CI [is failing](https://github.com/bytecodealliance/wit-bindgen/actions/runs/21878609843/job/63154448336) predictably due to the `threading-builtins` test using now-outdated names of intrinsics to import. It's still failing with this PR, but that's because Wasmtime hasn't updated to implement the new intrinsics. That's all as-expected to me, and the async tests are allowed to fail in CI, so I'm going to go ahead and merge this.

Thanks again @TartanLlama!

PR #1526, opened 2026-02-10
 Generate nominal IDs for all bindings generation 

alexcrichton:
This commit leverages https://github.com/bytecodealliance/wasm-tools/pull/2447 to make many
existing and future problems much easier in `wit-bindgen`. Namely a
`TypeId` now uniquely identifies a type to be generated rather than
simultaneously representing both an import and an export in some
situations. This isn't immediately leveraged in bindings generators just
yet but it's intended to open up the doors to benefitting from this in
the future.

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -24,7 +24,9 @@ pub enum Direction {
 
 pub trait WorldGenerator {
     fn generate(&mut self, resolve: &mut Resolve, id: WorldId, files: &mut Files) -> Result<()> {
-        resolve.generate_nominal_type_ids(id);
+        if self.uses_nominal_type_ids() {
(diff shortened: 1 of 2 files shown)


pchickey (approved):

crates/c/src/lib.rs
```diff
@@ -186,13 +186,18 @@ impl WorldGenerator for C {
         let world = &resolve.worlds[world];
         for (key, _item) in world.imports.iter().chain(world.exports.iter()) {
             let name = resolve.name_world_key(key);
-            interfaces.insert(name, key.clone());
+            interfaces
+                .entry(name)
+                .or_insert(Vec::new())
```
ricochet, line 191:
```suggestion
                .or_insert_with(Vec::new)
```

PR #1524, opened 2026-02-10
fix(rust): dealias stream/future types

ricochet:
This change dealiases inside print_ty during recursion, so it catches
aliases at every nesting depth.

PR #1482 only dealiases the top-level payload type.
That works when the payload itself is an alias
(e.g., future<error-code>), but not when the payload is an anonymous
type that contains aliases (e.g., future<result<_, error-code>>).

Fixes #1523
Resolves #1432

Issue #1523, opened 2026-02-09
Rust: P3 type aliases for futures with conflicting trait impls


ricochet:
When multiple interfaces use the same type from a shared interface (e.g., `types.{error-code}`) and each return a future parameterized with that type (`future<result<_, error-code>>`), wit-bindgen generates a separate `impl FuturePayload` block for each interface's type alias. Since all aliases resolve to the same concrete Rust type, this produces conflicting trait implementations.

Surfaced via the most recent rc (v0.3.0-rc-2026-02-09) where `wasi:cli` for stdin, stdout, stderr.

`tests/codegen/future-shared-type-duplicate-impls.wit`:
```wit
//@ async = true
package test:shared-future-type;

interface types {
  enum error-code {
    io,
    closed,
  }
}

interface a {
  use types.{error-code};
  f: func() -> future<result<_, error-code>>;
}

interface b {
  use types.{error-code};
  f: func() -> future<result<_, error-code>>;
}

interface c {
  use types.{error-code};
  f: func() -> future<result<_, error-code>>;
}

world w {
  import a;
  import b;
  import c;
}
```

```bash
cargo run -- test --languages rust --artifacts ./target/artifacts ./tests/codegen/future-shared-type-duplicate-impls.wit

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.28s
     Running `target/debug/wit-bindgen test --languages rust --artifacts ./target/artifacts ./tests/codegen/future-shared-type-duplicate-impls.wit`
Building `wit-bindgen` from crates.io...
=== Running codegen tests ===

running 5 tests
test rust future-shared-type-duplicate-impls.wit-borrowed-duplicate "./tests/codegen/future-shared-type-duplicate-impls.wit" ... FAILED
test rust future-shared-type-duplicate-impls.wit-borrowed "./tests/codegen/future-shared-type-duplicate-impls.wit"           ... FAILED
test rust future-shared-type-duplicate-impls.wit-no-std "./tests/codegen/future-shared-type-duplicate-impls.wit"             ... ok
test rust future-shared-type-duplicate-impls.wit "./tests/codegen/future-shared-type-duplicate-impls.wit"                    ... FAILED
test rust future-shared-type-duplicate-impls.wit-async "./tests/codegen/future-shared-type-duplicate-impls.wit"              ... FAILED

failures:
[...]

Issue #1433, opened 2025-11-24
rust: Stream/future bindings fail with root-level export function


alexcrichton:
This input:

```rust
wit_bindgen::generate!({
    inline: r#"
        package a:b;

        world w {
            use t.{r};
            export f: func(p: stream<r>);
        }

        interface t {
            record r {
                x: u32,
            }
        }
    "#,
});

fn main() {}
```

fails with:

```
$ cargo build --target wasm32-wasip1
   Compiling wat v0.1.0 (/home/alex/code/wat)
error[E0412]: cannot find type `R` in this scope
  --> src/main.rs:1:1
   |
 1 | / wit_bindgen::generate!({
 2 | |     inline: r#"
 3 | |         package a:b;
...  |
15 | |     "#,
16 | | });
   | |__^ not found in this scope
   |
   = help: consider importing one of these items:
           crate::R
           crate::a::b::t::R
   = note: this error originates in the macro `wit_bindgen::generate` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0412`.
error: could not compile `wat` (bin "wat") due to 1 previous error
```

alexcrichton:
Fixed in #1482

PR #1528, opened 2026-02-10
 rust: Reimplement how stream/future payloads work 

alexcrichton:
Previously stream/future payload were generated by collecting the set of
types used in `future` and `stream` types in a WIT, rendering them to a
string, deduplicating based on this string representation, and then
generating various impls-with-vtables. This stringification strategy
unfortunately falls down in a few situations such as:

* Type aliases in WIT render as two names in Rust, but they're using the
  same Rust type.
* Types with the same definition, but in multiple modules, will have two
  different paths in Rust but alias the same type.
* Primitives may be used directly in streams/futures but then
  additionally used as a WIT type alias.

In all of these situations it's effectively exposing how Rust requires
at most one-impl-per-type-definition but the stringification/deduping
was just a proxy for implementing this restriction and not a precise
calculation. Using the work from https://github.com/bytecodealliance/wasm-tools/pull/2447 as
well as https://github.com/bytecodealliance/wit-bindgen/pull/1468 it's possible to do all of this without stringifying.
Specifically https://github.com/bytecodealliance/wit-bindgen/pull/1468, transitively enabled by
https://github.com/bytecodealliance/wasm-tools/pull/2447, enables building a set of equal types
that the Rust generator knows will all alias the same type definition.
Using this it's possible to translate a payload to its "canonical
payload" representation ID-wise and perform hashing/deduplication based
on that. This in turn solves all of the issues above as well as previous
issues such as https://github.com/bytecodealliance/wit-bindgen/issues/1432 and https://github.com/bytecodealliance/wit-bindgen/issues/1433 without requiring the workaround in https://github.com/bytecodealliance/wit-bindgen/pull/1482.

The end result is that all of these various bugs should be fixed and the
Rust generator should be much more reliable about when exactly a trait
impl is emitted vs not.

Closes https://github.com/bytecodealliance/wit-bindgen/issues/1523
Closes https://github.com/bytecodealliance/wit-bindgen/pull/1524

### DIFF
--- a/crates/test/src/rust.rs
+++ b/crates/test/src/rust.rs
@@ -63,8 +63,9 @@ impl LanguageMethods for Rust {
         // no_std doesn't currently work with async
         if config.async_
             && args.iter().any(|s| s == "--std-feature")
-            // Except this one actually _does_ work:
+            // Except these actually do work:
             && name != "async-trait-function.wit-no-std"
+            && name != "async-resource-func.wit-no-std"
         {
             return true;
         }


alexcrichton:
This is built on https://github.com/bytecodealliance/wit-bindgen/pull/1526 and https://github.com/bytecodealliance/wit-bindgen/pull/1468 currently, on the last commit is relevant.

crates/rust/src/lib.rs
```diff
@@ -1101,6 +1128,39 @@ impl WorldGenerator for RustWasm {
...
+        self.types.collect_equal_types(resolve, &|a| {
+            // If `--merge-structurally-equal-types` is enabled then any type
+            // anywhere can be generated as a type alias to anything else.
+            if self.opts.merge_structurally_equal_types() {
+                return true;
+            }
+
+            match resolve.types[a].kind {
+                // These types are all defined with `type Foo = ...` in Rust
+                // since Rust either has native representstions or they live in
```
ricochet: nit, doesn't matter if fixed

```suggestion
                // since Rust either has native representations or they live in
```

ricochet:
C++ codegen doesn't have async stream/futures support yet. 

You could extend the list in `crates/test/src/cpp.rs` `should_fail_verify` list: `issue-1432.wit`, `issue-1433.wit`, `future-same-type-different-names.wit`

Looking at this with fresh eyes, I wish I had done this previously:

```rust
    fn should_fail_verify(
        &self,
        name: &str,
        config: &crate::config::WitConfig,
        _args: &[String],
    ) -> bool {
        // C++ doesn't support async yet
        config.async_
            || config.error_context
            || name == "async-resource-func.wit" // <-- missing async = true
    }
```

PR #1529, opened 2026-02-11
C#: Add partial support for Futures and Streams

yowl:
This PR adds partial support for streams and futures to the C# codegen.   Lifting/lowering is not implemented, and only byte type streams are supported.  I've not added the context here, not sure under what circumstance that would be required.  

In general there is a ReaderBase and a WriterBase as the base class for the Future/Future<T>/Stream/Stream<T> types,

### DIFF
--- a/crates/csharp/src/AsyncSupport.cs
+++ b/crates/csharp/src/AsyncSupport.cs
@@ -340,8 +340,8 @@ public static class FutureHelpers
         var readerHandle = (int)(packed & 0xFFFFFFFF);
         var writerHandle = (int)(packed >> 32);
 
-        Console.WriteLine($"Createing reader<T> with handle {readerHandle}");
-        Console.WriteLine($"Createing writer<T> with handle {writerHandle}");
+        Console.WriteLine($"Creating reader<T> with handle {readerHandle}");
+        Console.WriteLine($"Creating writer<T> with handle {writerHandle}");
         return (new StreamReader<T>(readerHandle, vtable), new StreamWriter<T>(writerHandle, vtable));
     }
 }


yowl:
Looks like some CI errors that are unrelated.

dicej:
Thanks, @yowl! I'm planning to review this tomorrow.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -31,68 +44,268 @@ public void Dispose()
...
 public static class AsyncSupport
 {
+    private static ConcurrentDictionary<int, ConcurrentDictionary<int, WaitableInfoState>> pendingTasks = new ();
+    internal static class PollWasmInterop
+    {
+        [DllImport("wasi:io/poll@0.2.0", EntryPoint = "poll"), WasmImportLinkage]
+        internal static extern void wasmImportPoll(nint p0, int p1, nint p2);
+    }
+
+    // TODO: How do we allow multiple waitable sets?
```
dicej, line 54: You'll want to store and retrieve the waitable set (and any other task-specific state) using the `context.set` and `context.get` intrinsics, respectively.  See e.g. [here](https://github.com/bytecodealliance/wit-bindgen/blob/d61d3708fe4a90ab811b94e1fd608f2f05fa2513/crates/go/src/package/wit_async/wit_async.go#L130) and [here](https://github.com/bytecodealliance/wit-bindgen/blob/d61d3708fe4a90ab811b94e1fd608f2f05fa2513/crates/go/src/package/wit_async/wit_async.go#L62) for how we do in in Go.

dicej: Oh, but it looks like you've already addressed this by adding the `ContextTask` struct below.  You should probably just remove this field entirely in favor of `ContextTask`.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -101,22 +314,51 @@ public static (FutureReader, FutureWriter) RawFutureNew(FutureVTable vtable)
...
+        return (new StreamReader(readerHandle, vtable), new StreamWriter(writerHandle, vtable));
+    }
+
+    internal static (StreamReader<T>, StreamWriter<T>) RawStreamNew<T>(StreamVTable vtable)
+    {
+        var packed = vtable.New();
+        var readerHandle = (int)(packed & 0xFFFFFFFF);
+        var writerHandle = (int)(packed >> 32);
+
+        Console.WriteLine($"Createing reader<T> with handle {readerHandle}");
```
dicej:
```suggestion
        Console.WriteLine($"Creating reader<T> with handle {readerHandle}");
```

crates/csharp/src/AsyncSupport.cs
```diff
@@ -101,22 +314,51 @@ public static (FutureReader, FutureWriter) RawFutureNew(FutureVTable vtable)
...
+        Console.WriteLine($"Createing writer<T> with handle {writerHandle}");
```
dicej:
```suggestion
        Console.WriteLine($"Creating writer<T> with handle {writerHandle}");
```

dicej (approved):
Thanks for continuing to iterate on this, @yowl.  It's great to see some more sophisticated tests passing.

The main thing I'd like to see addressed soon (but not necessarily a blocker for this PR) is to get rid of the `WaitableSet` global variable in favor of `ContextTask`.  That will help ensure multiple concurrent tasks don't interfere with each other.

Besides that, there are obviously still a lot of TODO items, but I'm sure those will all get addressed as we get more tests written and passing.

crates/csharp/src/AsyncSupport.cs
yowl: Unfortunately, dipping into this work not continuously means I sometimes forget where I was.  I've added another task to remove the global and use the context, where I can sort out `WaitableSet` and `ContextTask`

dicej:
@yowl do you have anything else planned for this PR, or is it ready to merge?

yowl:
I think it is OK to merge now.  Not sure about the protocol in this repo, do I do that ?

dicej:
> I think it is OK to merge now. Not sure about the protocol in this repo, do I do that ?

Usually a maintainer does it; I just wanted to make sure not to merge it if you were still iterating on it.

Issue #1515, opened 2026-01-28
csharp bindgen issues


kavoc-brown:
## Summary
This issue buckets several problems in `wit-bindgen csharp` to avoid opening many separate issues at once.

Each sub-item is independently reproducible and makes the generated C# fail to build.

If you prefer, I can split this bucket into separate issues.

## Repro template
```bash
WIT_BINDGEN=wit-bindgen
OUT=/tmp/wit-csharp
rm -rf "$OUT" && mkdir -p "$OUT"

"$WIT_BINDGEN" csharp --runtime native-aot --out-dir "$OUT" /tmp/CASE.wit

cat > "$OUT/Repro.csproj" <<'XML'
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
  </PropertyGroup>
</Project>
XML

dotnet build -c Release "$OUT/Repro.csproj"
```

#### Reproduction
```
package local:demo;

world w {
  export i;
}

interface i {
  /// */
  record r {
    x: u32,
  }
}
```

#### Reproduction
```
package local:demo;

world w {
  export i;
}

interface i {
  resource wasm-interop;

  foo: func(x: own<wasm-interop>);
}
```

#### Reproduction
```
package local:demo;

world w {
  import i;
}

interface i {
  resource t-handle;
}
```

#### Reproduction
```
package local:demo;

world w {
  export i;
}

interface i {
  resource r;
}
```

#### Reproduction
```
package local:demo;

world w {
  import i;
}

interface i {
  foo: func(cleanups: string);
}
```

#### Reproduction
```
package local:demo;

world w {
  export i;
}

interface i {
  foo: func(DELEGATE: u32);
}
```

#### Reproduction
```
package local:demo;

world w {
  export i;
}

interface i {
  variant tags {
    a,
  }

  foo: func(x: tags);
}
```

PR #1533, opened 2026-02-12
C#: Dont close the block comment if the user comment contains */

yowl:
This PR addresses issue 1 in #1515 as per the title.


dicej:
Would you mind adding a codegen test to cover this?  Looks like #1515 has a WIT file ready-made for that purpose.

yowl:
Go failure doesn't look related, will retry

Issue #1514, opened 2026-01-28
c++ bindgen can have multiple name collisions


kavoc-brown:
## Summary
Another bucket of issues in `wit-bindgen cpp`.

Each one still makes the generated C++ fail to compile, possibly due to name collisions.

I can split this bucket into separate issues if needed.

## Environment
- wit-bindgen-cli: 0.47.0
- OS: Ubuntu 22.04.5 LTS (x86_64)

Same compile command: `clang++ -c "$OUT/w.cpp" -I"$OUT" -o "$OUT/w.o"`.

#### Reproduction
```
package demo:poc;

world w {
  export a: interface {
    f: func() -> u32;
  }

  export b: interface {
    f: func() -> u64;
  }
}
```

#### Reproduction
```
package namespace:demo;

interface i {
  f: func();
}

world w {
  export i;
}
```

#### Reproduction
```
package demo:poc;

world w {
  import foo: func(ptr0: string) -> u32;
}
```

#### Reproduction
```
package demo:poc;

world w {
  import foo: func(class: u32) -> u32;
}
```

#### Reproduction
```
package demo:poc;

interface i {
  record t {
    ret: u32,
  }

  foo: func(x: t);
}

world w {
  import i;
}
```

#### Reproduction
```
package demo:poc;

interface i {
  variant v {
    v,
  }
  f: func(x: v);
}

world w {
  export i;
}
```

cpetig:
Likely fixed in 0.48.1+

avrabe:
Tested with wit-bindgen 0.51.0. Three sub-issues still reproduce:
- Sub-issue 1 (function overload): same function name in different interfaces
- Sub-issue 3 (ptr0 collision): parameter name conflicts with internal variable
- Sub-issue 6 (variant naming): variant case same name as variant type

Minimal reproducers: https://gist.github.com/avrabe/057880a37145ae339d4375e511873f74

PR #1535, opened 2026-02-12
C#: Compilation failure if variant has same name as one of the cases.

yowl:
This PR suffixes the variant case name with an underscore if it has the same name as the variant.  Fixes issue 6 in #1515


yowl:
#1514  has the test case for this, so added the c++ exclusion as that is known to fail.

PR #1536, opened 2026-02-12
fix: merge TypeInfo for structurally equal types

sumleo:
## Summary

- When `collect_equal_types` unions structurally identical types from different interfaces, the `TypeInfo` (borrowed/owned/error flags) was not being merged
- This caused incorrect code generation when a type appeared as an import parameter in one interface and an export return in another — only one side's flags would be visible, leading to missing borrowed or owned type variants
- Added a post-pass in `collect_equal_types` that merges `TypeInfo` across all members of each equivalence class using the existing `BitOrAssign` implementation

## Test plan

- [ ] Added codegen test in `crates/rust/tests/codegen.rs` with two interfaces containing structurally equal `payload` records (one imported, one exported) using `merge_structurally_equal_types` and `Borrowing` ownership
- [ ] Verified the test compiles and generates code without panic

### DIFF
--- a/crates/core/src/types.rs
+++ b/crates/core/src/types.rs
@@ -128,6 +128,21 @@ impl Types {
                 }
             }
         }
+
+        // Merge TypeInfo for structurally equal types so that all members
+        // of an equivalence class carry the union of all usage flags
+        // (borrowed, owned, etc.).
+        let mut merged: HashMap<TypeId, TypeInfo> = HashMap::new();
+        for (&id, &info) in &self.type_info {
+            let rep = self.equal_types.find(id);
+            *merged.entry(rep).or_default() |= info;
+        }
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Could the test be modeled as a new file in `tests/codegen/*.wit`? It could then be modeled [here](https://github.com/bytecodealliance/wit-bindgen/blob/51080a08cafd3e056abf17f75bf9b7a01a98a8a7/crates/test/src/rust.rs#L82-L92) with a new variant that passes `--merge-structurally-equal-types`

sumleo:
Done. Moved the test to `tests/codegen/merge-structurally-equal-types.wit` and added a `merge-equal` variant in the Rust codegen test infrastructure that passes `--merge-structurally-equal-types`. Also rebased on latest main.

PR #1537, opened 2026-02-12
fix: support named fixed-length list types in code generation

sumleo:
## Summary

- The `define_type` function in the core `InterfaceGenerator` trait had a `todo!()` for `FixedLengthList`, causing a panic when WIT files contained named fixed-length list types like `type my-array = list<u32, 4>`
- Added `type_fixed_length_list` method to the `InterfaceGenerator` trait with a default implementation
- Implemented it in the Rust backend following the `type_list` pattern, generating type aliases like `pub type MyArray = [u32; 4];`
- Updated `define_type` dispatch to call the new method

## Test plan

- [ ] Added codegen test WIT file `tests/codegen/named-fixed-length-list.wit` with named fixed-length list types in both import and export positions
- [ ] Added inline codegen test module in `crates/rust/tests/codegen.rs` verifying Rust codegen compiles

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -165,6 +165,7 @@ pub trait InterfaceGenerator<'a> {
     fn type_enum(&mut self, id: TypeId, name: &str, enum_: &Enum, docs: &Docs);
     fn type_alias(&mut self, id: TypeId, name: &str, ty: &Type, docs: &Docs);
     fn type_list(&mut self, id: TypeId, name: &str, ty: &Type, docs: &Docs);
+    fn type_fixed_length_list(&mut self, id: TypeId, name: &str, ty: &Type, size: u32, docs: &Docs);
     fn type_builtin(&mut self, id: TypeId, name: &str, ty: &Type, docs: &Docs);
     fn type_future(&mut self, id: TypeId, name: &str, ty: &Option<Type>, docs: &Docs);
     fn type_stream(&mut self, id: TypeId, name: &str, ty: &Option<Type>, docs: &Docs);
@@ -199,7 +200,9 @@ where
         TypeDefKind::Future(t) => generator.type_future(id, name, t, &ty.docs),
         TypeDefKind::Stream(t) => generator.type_stream(id, name, t, &ty.docs),
         TypeDefKind::Handle(_) => panic!("handle types do not require definition"),
-        TypeDefKind::FixedLengthList(..) => todo!(),
+        TypeDefKind::FixedLengthList(t, size) => {
+            generator.type_fixed_length_list(id, name, t, *size, &ty.docs)
+        }
         TypeDefKind::Map(..) => todo!(),
         TypeDefKind::Unknown => unreachable!(),
     }
--- a/crates/test/src/rust.rs
+++ b/crates/test/src/rust.rs
@@ -82,6 +82,11 @@ impl LanguageMethods for Rust {
             return true;
         }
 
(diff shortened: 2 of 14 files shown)


crates/core/src/lib.rs
```diff
@@ -165,6 +165,17 @@ pub trait InterfaceGenerator<'a> {
...
+        &mut self,
+        id: TypeId,
+        name: &str,
+        ty: &Type,
+        size: u32,
+        docs: &Docs,
+    ) {
+        let _ = (id, name, ty, size, docs);
+        todo!("named fixed-length list types are not yet supported in this backend")
+    }
```
alexcrichton: Could this default implementation be omitted and the `todo!` placed in bindings generators downstream that don't support this?

crates/rust/tests/codegen.rs
```diff
@@ -142,6 +142,27 @@ mod newtyped_list {
...
+        }
+
+        world test {
+            import types;
+            export types;
+        }
+        "#,
+        generate_all,
+    });
+}
```
alexcrichton: Could this be added to `tests/codegen/*.wit`?

crates/core/src/lib.rs
sumleo: Done. Removed the default implementation and added explicit `type_fixed_length_list` to all backends:

- **Rust**: generates `pub type Name = [T; N];` (already existed)
- **Go**: generates `type Name = [N]Type`
- **Markdown**: delegates to `type_alias`
- **Moonbit**: no-op (maps to `FixedArray[T]` natively)
- **C, C++, C#**: explicit `todo!()` with expected test failures

crates/rust/tests/codegen.rs
sumleo: Done. Moved the test to `tests/codegen/named-fixed-length-list.wit` and removed the inline version from `crates/rust/tests/codegen.rs`. Added expected failures in the test infrastructure for C, C++, and C# backends that don't yet support this.

alexcrichton (approved):
Thanks! Mind rebasing this as well to get this added to the merge queue?

crates/test/src/rust.rs
```diff
@@ -76,6 +76,11 @@ impl LanguageMethods for Rust {
             return true;
         }
 
+        // Named fixed-length lists don't work with async yet.
+        if name == "named-fixed-length-list.wit-async" {
```
cpetig, line 86: I would be curious to see the error (message) as these features should be orthogonal. Should be easy to fix (as should be a C++ implementation using `std::array<>`, C might be more tricky as it could require adding a struct).

sumleo: The error is a `todo\!()` panic in `deallocate` for `FixedLengthList` at `crates/core/src/abi.rs:2407` - the async code path calls `deallocate_lists_in_types` which hits the unimplemented deallocation. This is addressed by #1538, so once that lands this expected failure can be removed.

```
thread 'main' panicked at crates/core/src/abi.rs:2407:53:
not yet implemented
   3: wit_bindgen_core::abi::Generator<B>::deallocate
   6: wit_bindgen_rust::interface::InterfaceGenerator::deallocate_lists
   7: wit_bindgen_rust::interface::InterfaceGenerator::generate_guest_import_body_async
```

Re C++: agreed, `std::array<>` would work directly. I marked it as expected failure for now since this PR focuses on the named type alias dispatch, but a follow-up to implement it in C/C++/C# would be straightforward.

Issue #1501, opened 2026-01-13
C++, Rust limitation: Fixed Size List code can't handle non-POD elements


cpetig:
Non-POD/Copy elements require a transmute to/from `[MaybeUninit<T>, N]` and plenty of unsafe code to fill or extract the array element-wise.

In C++ you can use a union and placement new for creating the array.

flelchuk:
Union with raw storage and placement-new is likely UB. But there are workarounds in C++.
Here's one - not too ugly - [example](https://godbolt.org/z/q1M71Gh6z) on Compiler Explorer.

PR #1538, opened 2026-02-12
fix: implement deallocation for fixed-length lists with heap elements

sumleo:
## Summary

- The `deallocate` function in core ABI had a `todo!()` for `FixedLengthList`, panicking when flat deallocation was needed (e.g., `list<own<resource>, 3>`)
- The `deallocate_indirect` function was a silent no-op (`=> {}`), leaking memory for fixed-length lists containing heap-allocated elements like strings or nested lists
- For flat deallocation, reused `flat_for_each_record_type` with repeated element types to iterate and deallocate each element
- For indirect deallocation, reused `deallocate_indirect_fields` with a vec of repeated element types to correctly compute memory offsets

## Test plan

- [ ] Added codegen test verifying `cabi_post_` deallocation functions are generated for `list<string, 3>` exports
- [ ] `needs_deallocate` already correctly returns `true` for fixed-length lists with complex elements — the fix ensures the deallocation instructions are actually emitted

### DIFF
--- a/tests/runtime/fixed-length-lists/runner.rs
+++ b/tests/runtime/fixed-length-lists/runner.rs
@@ -4,30 +4,66 @@ include!(env!("BINDINGS"));
 
 use test::fixed_length_lists::to_test::*;
 
+mod alloc;
+
+struct Guard {
+    me_before: usize,
+    remote_before: u32,
+}
+
+impl Guard {
+    fn new() -> Guard {
+        Guard {
+            me_before: alloc::get(),
+            remote_before: allocated_bytes(),
+        }
+    }
+}
+
+impl Drop for Guard {
+    fn drop(&mut self) {
+        assert_eq!(self.me_before, alloc::get());
+        assert_eq!(self.remote_before, allocated_bytes());
+    }
+}
+
 struct Component;
 
 export!(Component);
 
 impl Guest for Component {
     fn run() {
-        list_param([1, 2, 3, 4]);
-        list_param2([[1, 2], [3, 4]]);
-        list_param3([
-            -1, 2, -3, 4, -5, 6, -7, 8, -9, 10, -11, 12, -13, 14, -15, 16, -17, 18, -19, 20,
-        ]);
         {
+            let _guard = Guard::new();
+            list_param([1, 2, 3, 4]);
+        }
+        {
+            let _guard = Guard::new();
+            list_param2([[1, 2], [3, 4]]);
+        }
+        {
+            let _guard = Guard::new();
+            list_param3([
+                -1, 2, -3, 4, -5, 6, -7, 8, -9, 10, -11, 12, -13, 14, -15, 16, -17, 18, -19, 20,
+            ]);
+        }
+        {
+            let _guard = Guard::new();
             let result = list_result();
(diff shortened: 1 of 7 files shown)


sumleo:
Done. Extended the existing `tests/runtime/fixed-length-lists/` test with:

- `string-list-param`, `string-list-result`, and `string-list-roundtrip` functions using `list<string, N>` in the WIT interface
- Custom allocator (`alloc.rs`) in both test and runner components
- `allocated-bytes` function to track heap allocation
- Guard pattern in the runner to verify no memory leaks after each call

This exercises both the `deallocate` (flat) and `deallocate_indirect` (pointer-based) code paths for fixed-length lists with heap-allocated elements.

crates/rust/tests/codegen.rs
```diff
@@ -142,6 +142,44 @@ mod newtyped_list {
...
+    let code = String::from_utf8_lossy(contents);
+
+    // Verify post-return deallocation functions are generated for
+    // fixed-length lists containing heap-allocated elements (strings)
+    assert!(
+        code.contains("cabi_post_"),
+        "expected cabi_post_ deallocation function for fixed-length list \
+         with string elements, but none was generated"
+    );
+}
```
alexcrichton: Mind dropping this test as well? This should be exercised by `tests/runtime/fixed-length-lists` and avoids trying to peek into the generated code like this

cpetig:
I think this ~~(at least partially as it doesn't fix c++ code)~~ fixes https://github.com/bytecodealliance/wit-bindgen/issues/1501 .

Update: While the test for the new functionality is (incorrectly) tied to Rust, this could also fix the C++ code because of the change in `abi.rs`. I am positive but I wouldn't promise a full fix before I tested more.

tests/runtime/fixed-length-lists/runner.rs
```diff
@@ -4,30 +4,66 @@ include!(env!("BINDINGS"));
...
             let result = list_minmax16([0, 1024, 32768, 65535], [1, 2048, -32767, -2]);
             assert_eq!(result, ([0, 1024, 32768, 65535], [1, 2048, -32767, -2]));
         }
         {
+            let _guard = Guard::new();
             let result = list_minmax_float([2.0, -42.0], [0.25, -0.125]);
             assert_eq!(result, ([2.0, -42.0], [0.25, -0.125]));
         }
         {
+            let _guard = Guard::new();
```
cpetig, line 66: All these allocation guards test for a no-op (as there must be no allocations for any of the tested types), but they don't hurt either. 

The `list<string,3>` test would become interesting with a guard though.

crates/rust/tests/codegen.rs
sumleo: Done. Removed the codegen test — the runtime tests now cover this with actual allocation tracking.

tests/runtime/fixed-length-lists/runner.rs
sumleo: Good point. Added `list<string, 3>` test functions (`string-list-param`, `string-list-result`, `string-list-roundtrip`) with Guards that actually exercise deallocation of heap-allocated elements.

alexcrichton (approved):

PR #1539, opened 2026-02-12
fix(go): prevent GC finalizer on borrowed resource handles

sumleo:
## Summary

- `FromBorrowHandle` for imported resources was incorrectly delegating to `FromOwnHandle`, which registers a GC finalizer via `runtime.AddCleanup`
- This violates the Component Model specification: borrowed handles are not owned by the callee, so the callee must not drop them
- When Go's garbage collector ran, it would invoke the finalizer and call `resourceDrop` on the borrowed handle, corrupting the host's resource table
- Fixed `FromBorrowHandle` to create a handle object directly without registering a GC finalizer, preserving correct ownership semantics

## Test plan

- [ ] Added integration test `tests/go_borrow_handle.rs` that generates Go code for an imported resource with `borrow<blob>` parameter
- [ ] Verified the generated `FromBorrowHandle` function does NOT contain `FromOwnHandle` delegation or `AddCleanup` calls
- [ ] The existing `FromBorrowHandle` for exported resources (which correctly uses pointer-based rep without finalizer) is unchanged

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,9 +119,3 @@ csharp = ['dep:wit-bindgen-csharp']
 csharp-mono = ['csharp']
 moonbit = ['dep:wit-bindgen-moonbit']
 async = []
-
-[dev-dependencies]
-wit-bindgen-core = { workspace = true }
-wit-bindgen-rust = { workspace = true }
-wit-bindgen-go = { workspace = true }
-wit-parser = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1502,7 +1502,6 @@ dependencies = [
  "wit-bindgen-rust",
  "wit-bindgen-test",
  "wit-component",
- "wit-parser",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


alexcrichton:
Thanks! Would it be possible to move the test to something in `tests/runtime/*` to actually exercise the generated/compiled code?

sumleo:
Done. Replaced the codegen test with a runtime test by adding Go test files (`test.go` and `runner.go`) to the existing `tests/runtime/resource-borrow/` directory. The test creates a `Thing` resource with a constructor that adds 1 to the value, then calls `foo` which adds 2 to it via a borrowed handle, verifying the result.

Cargo.toml
```diff
@@ -119,3 +119,9 @@ csharp = ['dep:wit-bindgen-csharp']
 csharp-mono = ['csharp']
 moonbit = ['dep:wit-bindgen-moonbit']
 async = []
+
+[dev-dependencies]
+wit-bindgen-core = { workspace = true }
+wit-bindgen-rust = { workspace = true }
+wit-bindgen-go = { workspace = true }
+wit-parser = { workspace = true }
```
alexcrichton: I think this might be an unrelated stray change?

sumleo: Good catch, removed. It was a leftover from the previous codegen test that needed those dev-dependencies.

PR #1546, opened 2026-02-15
[C#] Remove 'readonly' from struct and field definitions

andreakarasho:
The host pass a vector of `Velocity` to the guest.
You want to change only the `X`.
Today you have to re-assign each index like:

```csharp
readonly struct Velocity { public readonly float X, Y; }
...
velocities[i] = new Velocity(velocities[i].X + 1, velocities[i].Y);
```

Removing the `readonly` you can simply do:

```csharp
struct Velocity { public float X, Y; }
...
velocities[i].X += 1;
```


yowl (approved):
This looks good.  Thanks.

Issue #1543, opened 2026-02-13
rust: bug with merge structurally equal types


chenyan2002:
When merge structurally equal types is enabled, I get panic at [path_to_interface](https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/rust/src/interface.rs#L2415) function. Basically, the owner of `TypeId` from `get_representative_type` is not in `self.r#gen.interface_names`.

I can work around some of these cases by calling `compute_module_path` when the lookup fails. This happens when WIT import and export the same interface, but it doesn't always panic.

In some other cases, the `interfaceId` points to an unused interface, which is not present in `bindings.rs`. The case I see so far is for primitive types,

```
package a;
type Duration = u32;

package b;
type Counter = u32;
```

With equal types, we get `type Counter = Duration`, but since `Duration` is not used, it's not in the bindgen code. This particular case can be fixed by disallowing aliasing for primitive types. But I don't know if there are cases other than primitives.

I don't have a minimal repo yet, but think this may be related to the nominal id patch. I tried to bump wasm-tools to 0.245.1, but still the same problem. When can an interface live outside of `self.r#gen.interface_names` or we need to maintain dependency differently when merging equal types?

alexcrichton:
Hm one way this might happen is because a `Resolve` can contain more interfaces than are present in a `world`. It might be that `collect_equal_types` needs to just work on types reachable from a `world` and nothing else?

chenyan2002:
Another problem is that we insert `self.interface_names` in `import_interface`. With equal_types, we may get a type alias that is not inserted yet. We may need to lift the interface_names insertion in the preprocess phase.

alexcrichton:
One trick taken from https://github.com/bytecodealliance/wit-bindgen/pull/1547 which might work here is to use equal types as a hash key but still work on the provided id-at-hand to actually generate bindings. That would mean that we still only operate on provided ids, and where exactly the equal ids are located ends up being irrelevant

chenyan2002:
I see. I have an alternative fix to move `name_interface` to the preprocess phase. We could use the same trick as #1547 as well.

Hmm, this might also fix the non-reachable types from the current world. I will give it another try.

chenyan2002:
> still work on the provided id-at-hand to actually generate bindings

Thinking a bit more, we don't have an id-at-hand in `define_type`, so we probably cannot apply the same trick from #1547?

[...]

PR #1549, opened 2026-02-19
fix(rust): Restrict equal types to the current world

chenyan2002:
Fix [#1543](https://github.com/bytecodealliance/wit-bindgen/issues/1543)

* Move `name_interface` to preprocess phase to avoid crashing when referring to not-yet-generated types from bindgen
* Restrict equal types to the current world, so that bindgen doesn't refer to types not reachable from the current world
* Reenable test `wasi-filesystem-merge-equal`, which is fixed by this PR.

### DIFF
--- a/crates/core/src/types.rs
+++ b/crates/core/src/types.rs
@@ -113,24 +113,19 @@ impl Types {
     ) {
         let mut live_types = wit_parser::LiveTypes::default();
         live_types.add_world(resolve, world_id);
-        let types: Vec<_> = resolve
-            .types
-            .iter()
-            .filter(|(t, _)| live_types.contains(*t))
-            .collect();
-        for (i, (ty, _)) in types.iter().enumerate() {
-            if !may_alias_another_type(*ty) {
+        for (i, ty) in live_types.iter().enumerate() {
+            if !may_alias_another_type(ty) {
                 continue;
             }
             // TODO: we could define a hash function for TypeDefKind to prevent the inner loop.
-            for (earlier, _) in types.iter().take(i) {
-                if self.equal_types.find(*ty) == self.equal_types.find(*earlier) {
+            for earlier in live_types.iter().take(i) {
+                if self.equal_types.find(ty) == self.equal_types.find(earlier) {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oh sorry I didn't realize this was ready for some review. I figured draft + failing tests meant there was still some work to do locally.

These changes in general look good to me, I'm not entirely sure what would cause the test failures though. To me though it seems like it's in the right direction at least for resolving #1543 (I'd also request to be sure to include tests as well here though)

crates/core/src/types.rs
```diff
@@ -109,21 +108,43 @@ impl Types {
...
+        let types: Vec<_> = resolve
+            .types
+            .iter()
+            .filter(|(t, _)| match resolve.types[*t].owner {
+                TypeOwner::Interface(id) => interfaces.contains(&id),
+                TypeOwner::World(id) => world_id == id,
+                // primitive types are always reachable
+                TypeOwner::None => true,
+            })
+            .collect();
```
alexcrichton: For this, looking at this again, I might recommend using [`wit_parser::LiveTypes`](https://docs.rs/wit-parser/latest/wit_parser/struct.LiveTypes.html) as you should be able to call `add_world` and that'll automatically spider/probe all types reachable from a world in a topological fashion, no need to iterate manually.

crates/core/src/types.rs
```diff
@@ -109,21 +108,29 @@ impl Types {
...
         may_alias_another_type: &dyn Fn(TypeId) -> bool,
     ) {
-        for (i, (ty, _)) in resolve.types.iter().enumerate() {
-            if !may_alias_another_type(ty) {
+        let mut live_types = wit_parser::LiveTypes::default();
+        live_types.add_world(resolve, world_id);
+        let types: Vec<_> = resolve
+            .types
+            .iter()
+            .filter(|(t, _)| live_types.contains(*t))
```
alexcrichton: Oh for this it should be sufficient to switch the iterator below to iterating over the `live_types` set, no need to iterate over the resolve's types, filter to what `live_types` has, and then iterate over that set

chenyan2002:
Thanks! This is indeed much better!

PR #1551, opened 2026-03-01
Add support for `implements` clause

ricochet:
The `implements` clause lets a world import or export the same interface multiple times under different plain names:

    world runner {
      import primary: store;
      import backup:  store;
    }

No bindings-generator changes needed. wit-parser surfaces each label as a plain-name interface import/export, `generate_nominal_type_ids` gives each its own type identity, and the existing plain-name interface code path emits label-named modules and `wasm_import_module`s.

The test covers Rust, Go, and WAT runners with Rust and Go tests, composed via `compose.wac` so the two labels resolve to two independent `store` instances.

Temporary, until releases catch up:
- `wac` is pointed at the `implements` branch; composition support for the clause is not yet in a released wac.
- CI's wasmtime is pinned to the `dev` build; the implements has landed on wasmtime `main` but is not yet released.

### DIFF
--- a/.github/actions/install-wasi-sdk/action.yml
+++ b/.github/actions/install-wasi-sdk/action.yml
@@ -38,4 +38,8 @@ runs:
     - name: Setup `wasmtime`
       uses: bytecodealliance/actions/wasmtime/setup@v1
       with:
-        version: "44.0.0"
+        # Pinned to the rolling `dev` build (main, which will become wasmtime
+        # 47) because the component model `implements` clause used by the
+        # `implements` runtime test has landed on `main` but is not in a
+        # released wasmtime yet.
+        version: "dev"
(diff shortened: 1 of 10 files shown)


alexcrichton:
The guest side of things I think is going to be tricky in its own right compared to the host side of things. Notably if an interface is imported twice this'll need to generate two different types for all types contained within the interface. Without that it's not a reflection of the original WIT itself, so this means that (at least from my understanding at a skim) the duplicate-related handling will need to be removed.

To handle this though what I'd recommend is updating the [`Resolve::generate_nominal_type_ids`](https://docs.rs/wit-parser/latest/wit_parser/struct.Resolve.html#method.generate_nominal_type_ids) method. That would transform imports of the same interface multiple times into imports of distinct interfaces which will make it much easier to support this in bindings generators (different `InterfaceId`s and `TypeId`s for different interfaces/types). The logic of `generate_nominal_type_ids` will be a bit tricky, but I think that's the best foundational primitive to use here.

PR #1552, opened 2026-03-01
[C#] Remove WaitableSet and its global.  Tidy <T> in codegen

### LINKED ISSUE #1550: [C#] Remove WaitableSet and use context(set/get) instead.
See https://github.com/bytecodealliance/wit-bindgen/pull/1529#pullrequestreview-3799794533 for original comment.

yowl:
This PR removes the disposable WaitableSet class and the global, replacing it with `ContextSet`/`ContextGet`

Closes #1550

### DIFF
--- a/crates/csharp/src/AsyncSupport.cs
+++ b/crates/csharp/src/AsyncSupport.cs
@@ -54,7 +54,7 @@ public static class AsyncSupport
         internal static unsafe extern uint WaitableSetPoll(int waitable, uint* waitableHandlePtr);
 
         [global::System.Runtime.InteropServices.DllImport("$root", EntryPoint = "[waitable-set-drop]"), global::System.Runtime.InteropServices.WasmImportLinkageAttribute]
-        internal static unsafe extern void WaitableSetDrop(int waitable);
+        internal static extern void WaitableSetDrop(int waitable);
 
         [global::System.Runtime.InteropServices.DllImport("$root", EntryPoint = "[context-set-0]"), global::System.Runtime.InteropServices.WasmImportLinkageAttribute]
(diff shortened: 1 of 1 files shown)


crates/csharp/src/AsyncSupport.cs
```diff
@@ -233,7 +212,7 @@ public static Task TaskFromStatus(uint status)
         }
     }
 
-    public static Task<T> TaskFromStatus<T>(uint status, Func<T> liftFunc)
+    public static unsafe Task<T> TaskFromStatus<T>(uint status, Func<T> liftFunc)
```
dicej, line 219: What makes this function unsafe to call?  Could it be documented?

crates/csharp/src/AsyncSupport.cs
```diff
@@ -389,14 +371,17 @@ internal unsafe Task<int> ReadInternal(Func<GCHandle?> liftBuffer, int length)
         {
             Console.WriteLine("Read Blocked");
             var tcs = new TaskCompletionSource<int>();
-            if(AsyncSupport.WaitableSet == null)
+            ContextTask* contextTaskPtr = AsyncSupport.ContextGet();
+            if(contextTaskPtr == null)
             {
                 Console.WriteLine("FutureReader Read Blocked creating WaitableSet");
-                AsyncSupport.WaitableSet = AsyncSupport.WaitableSetNew();
+                contextTaskPtr = (ContextTask *)Marshal.AllocHGlobal(Marshal.SizeOf<ContextTask>());
```
dicej: These same three lines appear in several places; might want to consider adding a helper function which encapsulates them.

PR #1558, opened 2026-03-05
fix: update submodule

asteurer:
Technically it doesn't affect anything, but I figured it would be best practice to keep the submodule up-to-date

@dicej


asteurer:
I don't yet have merge permissions, so I'll need someone to take care of that for me

PR #1560, opened 2026-03-09
feat: add empty.s to generated directories

### LINKED ISSUE #1555: Go: make bindings testable in native Go
# Description

When attempting to run unit tests I get the following compilation error when `wit-bindgen go` bindings are imported:
```sh
$ go test ./...
# <go-project>/types/wasi_logging_logging
<go-project>/types/wasi_logging_logging/wit_bindings.go:42:6: missing function body
FAIL    <go-project> [build failed]
FAIL
```

With the bindings file showing the following:
```go
...
//go:wasmimport wasi:logging/logging@0.1.0-draft log
func wasm_import_log(arg0 int32, arg1 uintptr, arg2 uint32, arg3 uintptr, arg4 uint32)
...
```

[`wit-bindgen-go`](https://github.com/bytecodealliance/go-modules/tree/main) got round this issue by including a blank `empty.s` assembly file within each of the directories which includes a Go file that uses the `//go:wasmimport` directive. See [here](https://github.com/wasmCloud/go/blob/main/component/gen/wasi/logging/logging/empty.s) for example.

This tricks the Go compile into thinking there is an assembly implementation of the function which has been declared, and so it no longer complains about the missing function body.

Is this something which we could include in these bindings to get around this build error?

asteurer:
Closes #1555

@dicej @jamesstocktonj1

### DIFF
--- a/crates/go/src/lib.rs
+++ b/crates/go/src/lib.rs
@@ -956,6 +956,14 @@ import (
                         .as_bytes(),
                     ),
                 );
+
+                files.push(
+                    &format!("{prefix}{name}/empty.s"),
+                    r#"// This file exists for testing this package without WebAssembly,
+// allowing empty function bodies with a //go:wasmimport directive.
+// See https://pkg.go.dev/cmd/compile for more information."#
+                        .as_bytes(),
+                );
             }
         }
 


jamesstocktonj1:
Checked it out. Looks good to me 👍

PR #1562, opened 2026-03-11
Add map support

yordis:
## Summary
- add end-to-end `map<K, V>` support in `wit-bindgen` core ABI and backend codegen paths (C, C++, C#, Go, MoonBit), plus Markdown type rendering
- add map-focused codegen/runtime coverage in `tests/codegen/map.wit` and `tests/runtime/map/*`
- fix Go test harness module replacement in `crates/test/src/go.rs` so generated map bindings resolve `go.bytecodealliance.org/pkg/wit/*` correctly

## Test plan
- [x] `cargo fmt`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo check`
- [x] `cargo check -p wit-bindgen-core`
- [x] `cargo check -p wit-bindgen-c`
- [x] `cargo check -p wit-bindgen-cpp`
- [x] `cargo check -p wit-bindgen-csharp`
- [x] `cargo check -p wit-bindgen-go`
- [x] `cargo check -p wit-bindgen-moonbit`
- [x] `cargo check -p wit-bindgen-markdown`
- [x] `cargo run test --artifacts target/artifacts --runner cargo --languages go --filter map.wit tests/codegen`

### DIFF
--- a/crates/rust/src/bindgen.rs
+++ b/crates/rust/src/bindgen.rs
@@ -772,7 +772,10 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 let rt = self.r#gen.r#gen.runtime_path().to_string();
                 self.r#gen.needs_wit_map = true;
                 self.r#gen.needs_runtime_module = true;
-                self.r#gen.r#gen.rt_module.insert(crate::RuntimeItem::WitMapTrait);
+                self.r#gen
+                    .r#gen
+                    .rt_module
+                    .insert(crate::RuntimeItem::WitMapTrait);
                 let body = self.blocks.pop().unwrap();
                 let tmp = self.tmp();
                 let map = format!("map{tmp}");


crates/moonbit/src/lib.rs
```diff
@@ -2199,6 +2203,57 @@ impl Bindgen for FunctionBindgen<'_, '_> {
                 }
             }
 
+            Instruction::MapLower {
```
alexcrichton: Could support for other languages be omitted unless tests are also added? I'm not equipped to handle other languages and forgive me if I'm wrong, but if the other languages' support is LLM generated I would perfer someone review the code who knows about the language generation.

crates/test/src/go.rs
```diff
@@ -161,7 +161,7 @@ fn replace_bindings_go_mod(runner: &Runner, bindings_dir: &Path) -> Result<()> {
     super::write_if_different(
         &bindings_dir.join("go.mod"),
         format!(
-            "module wit_component\n\ngo 1.25\n\nreplace go.bytecodealliance.org => {}",
+            "module wit_component\n\ngo 1.25\n\nreplace go.bytecodealliance.org/pkg => {}",
```
alexcrichton: Is this perhaps a stray change? (otherwise I'm not sure where this comes from)

crates/rust/src/lib.rs
```diff
@@ -98,6 +98,7 @@ impl TypeGeneration {
 #[derive(PartialEq, Eq, Clone, Copy, Hash, Debug)]
 enum RuntimeItem {
     AllocCrate,
+    MapType,
```
alexcrichton: This is a bit of a dated enum which shouldn't be neede any more, cousl `Map` refer to `{rt}::Map` directly in various locations instead of requiring a `pub use`?

crates/rust/src/interface.rs
```diff
@@ -1931,6 +1937,32 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
...
+            self.push_str("&");
+            if lifetime != "'_" {
+                self.push_str(lifetime);
+                self.push_str(" ");
+            }
+            self.push_str("[(");
+            self.print_ty(key, key_mode);
+            self.push_str(", ");
+            self.print_ty(value, value_mode);
+            self.push_str(")]");
```
alexcrichton: This looks like it might be copy/pasted from the lists side of things, but the slice syntax I don't think is intended here? Or is this intended? Should maps not be passed as `&MapThing<K, V>`?

crates/rust/src/interface.rs
```diff
@@ -1786,6 +1786,12 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
                         return;
                     }
                 }
+                TypeDefKind::Map(key, value) => {
+                    if mode.lifetime.is_some() {
+                        self.print_map(key, value, mode);
+                        return;
+                    }
+                }
                 _ => {}
```
alexcrichton, line 1802: If adding this here, can you update the comment above?

crates/moonbit/src/pkg.rs
```diff
@@ -217,6 +217,11 @@ impl PkgResolver {
                                 .join(", ")
                         )
                     }
+                    TypeDefKind::Map(key, value) => {
```
alexcrichton: Similar to above, mind leaving this unimplemented for a future PR to add tests and get a review?

crates/go/src/lib.rs
alexcrichton, line 1: Similar to moonbit, could this be deferred to a future PR with tests + review?

crates/csharp/src/function.rs
alexcrichton, line 1: Similar to moonbit/go, could this be deferred for a future PR for C# which has tests + a review?

crates/cpp/src/lib.rs
alexcrichton, line 1: Same for other languages, can this be deferred until there's tests?

crates/core/src/lib.rs
```diff
@@ -166,6 +166,10 @@ pub trait InterfaceGenerator<'a> {
     fn type_alias(&mut self, id: TypeId, name: &str, ty: &Type, docs: &Docs);
     fn type_list(&mut self, id: TypeId, name: &str, ty: &Type, docs: &Docs);
     fn type_fixed_length_list(&mut self, id: TypeId, name: &str, ty: &Type, size: u32, docs: &Docs);
+    fn type_map(&mut self, id: TypeId, name: &str, key: &Type, value: &Type, docs: &Docs) {
+        let _ = (id, name, key, value, docs);
+        todo!("map types are not yet supported in this backend");
+    }
     fn type_builtin(&mut self, id: TypeId, name: &str, ty: &Type, docs: &Docs);
```
alexcrichton: Can this be a required method like all other methods in this trait? (e.g. no default impl)

crates/core/src/abi.rs
```diff
@@ -2115,6 +2183,7 @@ impl<'a, B: Bindgen> Generator<'a, B> {
                 TypeDefKind::Type(t) => self.read_from_memory(t, addr, offset),
 
                 TypeDefKind::List(_) => self.read_list_from_memory(ty, addr, offset),
+                TypeDefKind::Map(_, _) => self.read_list_from_memory(ty, addr, offset),
```
alexcrichton, line 2188: Can this have a comment for why `read_list_from_memory` works?

crates/core/src/abi.rs
```diff
@@ -1907,6 +1975,7 @@ impl<'a, B: Bindgen> Generator<'a, B> {
             Type::Id(id) => match &self.resolve.types[id].kind {
                 TypeDefKind::Type(t) => self.write_to_memory(t, addr, offset),
                 TypeDefKind::List(_) => self.write_list_to_memory(ty, addr, offset),
+                TypeDefKind::Map(_, _) => self.write_list_to_memory(ty, addr, offset),
```
alexcrichton, line 1979: Similar to below, can this have a comment for why `write_list_to_memory` works?

crates/c/src/lib.rs
alexcrichton, line 1: Similar to other languages, I don't think this should be added without tests. Unlike other languages, though, I can review this, so mind adding some tests using the C bits?

crates/guest-rust/src/rt/mod.rs
```diff
@@ -67,6 +67,11 @@ pub mod bitflags {
     pub use crate::bitflags;
 }
 
+#[cfg(feature = "std")]
+pub type Map<K, V> = std::collections::HashMap<K, V>;
+#[cfg(not(feature = "std"))]
+pub type Map<K, V> = alloc::collections::BTreeMap<K, V>;
```
alexcrichton: This I'm hesitant to do myself, specifically switching based on `std`. That's not API-compatible unfortunately and puts this crate in a weird situation.

One possibility perhaps is to force some sort of configuration in `bindgen!` if `map<K, V>` is encountered which specifies which map type to use. Another is to always fall back to the more general `BTreeMap` in terms of availability. Another is to support both a fallback and a configuration. Another yet might be to avoid using `HashMap` and `BTreeMap` entirely and instead use `Vec<(T, U)>` by default. Overall I'm not sure what the best option is...

yordis: @alexcrichton honestly, I do not feel qualify to decide, until now, I have been "pattern matching" and coding based on the List type. Digging more into Rust details, here is the finding thus far based on AI, so I need to triple verify, I may delay this topic and try to make things to work first, and swap this later

----

I think `BTreeMap` is actually the right default, and the current choice is fine for now. Here's my reasoning:

**WIT key types are constrained.** Keys in a WIT map won't be arbitrary Rust types -- they'll be primitives, strings, enums, or simple records. All of these naturally implement `Ord` (and `Hash + Eq`). So the trait bound difference between `BTreeMap` and `HashMap` is largely theoretical for WIT-generated code.

**It provides real map semantics.** Unlike `Vec<(K, V)>`, you get deduplication and O(log n) lookup. Calling something a "map" but handing users a `Vec` would be misleading and push complexity onto every consumer.

**It works everywhere.** `no_std + alloc` is the baseline for Wasm guest code, and `BTreeMap` lives in `alloc`. No extra dependencies, no feature flag gymnastics.

**Configuration can come later.** A `map_type` option in `bindgen!` (similar to `runtime_path`) could be added as a non-breaking enhancement if someone has a concrete need for `HashMap` or a custom type. But it's not needed to land map support.

The one thing I'd avoid is `Vec<(K, V)>` -- even though it matches the ABI, it sacrifices the whole point of having a `map` type in WIT. If users wanted list-of-tuples semantics, they'd write `list<tuple<K, V>>` in their WIT.

So my suggestion: keep `BTreeMap` as-is, and note in the PR that a configurable `map_type` option is a possible future enhancement.

crates/rust/src/bindgen.rs
```diff
@@ -759,6 +763,55 @@ impl Bindgen for FunctionBindgen<'_, '_> {
...
+                let map = format!("map{tmp}");
+                let result = format!("result{tmp}");
+                let layout = format!("layout{tmp}");
+                let len = format!("len{tmp}");
+                let cleanup = format!("_cleanup{tmp}");
+                self.push_str(&format!(
+                    "let {map} = {operand0};\n",
+                    operand0 = operands[0]
+                ));
+                self.push_str(&format!("let {len} = {map}.len();\n"));
```
alexcrichton: Can `.len()` here and `.into_iter()` below also go through the `WitMap` trait?

PR #1563, opened 2026-03-12
[C#] Add resolving of equivalent types.

yowl:
This PR enables the future-same-type-different-names.wit test by using the same logic as Rust for the resolving of equivalent types.

### DIFF
--- a/crates/csharp/src/interface.rs
+++ b/crates/csharp/src/interface.rs
@@ -1459,12 +1459,12 @@ var {async_status_var} = {raw_name}({wasm_params});
         &mut self,
         func_name: &str,
         generic_type_name: &str,
-        ty: &&Option<Type>,
+        ty: Option<Type>,
     ) {
         self.futures.push(FutureInfo {
             name: func_name.to_string(),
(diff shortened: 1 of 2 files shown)


crates/csharp/src/interface.rs
```diff
@@ -1439,19 +1455,29 @@ var {async_status_var} = {raw_name}({wasm_params});
         format!("{access} {modifiers} {result_type} {camel_name}({params})")
     }
 
-    pub(crate) fn add_future(&mut self, func_name: &str, generic_type_name: &str, ty: &TypeId) {
+    pub(crate) fn add_future(
+        &mut self,
+        func_name: &str,
+        generic_type_name: &str,
+        ty: &&Option<Type>,
```
dicej: This is an unusual looking parameter type for Rust.  I'd recommend changing it to either `ty: &Option<Type>` or `ty: Option<Type>` and letting the caller take care of dereferencing as needed.  Same goes for the `add_stream` function below.

yowl: Thanks, changed to `Option<Type>`

Issue #1062, opened 2024-10-02
Import stubs for native targets


rvolosatovs:
Currently, `rust` bindgen inserts stubs for Wasm imports on native targets, which simply panic.

I'd like to discuss possibilities for making this configurable or perhaps define a convention of some sorts?

My exact use case is as follows:
At wasmCloud we've developed https://github.com/wasmCloud/wadge, which lets components be tested as native applications - for Go, it works by replacing each function with a `wasmimport` directive, generated by upstream bindgen, by an implementation, which calls an export on a component running in an embedded Wasmtime instance. E.g. https://github.com/wasmCloud/wadge/blob/c21abaf5487ad79dd17660858127b3232c2623f4/bindings/bindings.go#L22-L36 for `wasi:clocks/monotonic-clock@0.2.1#now`

This works by using [canonical ABI-*ish* ](https://github.com/wasmCloud/cabish) to read and write data, which is essentially just canonical ABI for native targets.

I'd like to be able to follow a similar approach for Rust components and in future - other languages, but Rust is the current focus.

A straightforward way to handle this could be configuring the stub generation for imports on native targets.

Perhaps `wit-bindgen` macro could take in a `NativeGenerator` or something similar as an optional argument, which allow users to configure how these imports are generated?
Perhaps the API could look similar to https://github.com/bytecodealliance/wit-bindgen/blob/66bfda38bee488df27692311e364d79a3cd301a7/crates/rust/src/interface.rs#L457:
```rust
trait NativeGenerator {
   fn generate_import(&mut self, func: &Function)
}
```

The trait could have a default implementation generating the `unreachable!` stub.

Of course there would need to be some API to facilitate the printing, perhaps `self` could expose the `src`, or the function might return a `String`

I'm happy to work on contributing this feature myself if there's an interest for getting it merged.

alexcrichton:
I think having a trait or similar to configure makes sense yeah. Wouldn't really be accessible through the CLI or the procedural macro but that seems ok too

BjornTheProgrammer:
Is there any work being done on this currently? This would be really useful for something like a plugin system that supports both native and wasm plugins. One for speed, the other for security.

alexcrichton:
I don't know of anyone working on this if you'd like to give it a go.

BjornTheProgrammer:
Alright, I've embarked on implementing this. I've thought of a couple of different solutions, but they all have their pitfalls. 

[...]

PR #1565, opened 2026-03-13
feat(rust): import stubs for native targets

BjornTheProgrammer:
This PR adds the ability for the user to provide a macro to rust that then can generate native funciton stubs, as described in https://github.com/bytecodealliance/wit-bindgen/issues/1062.

There were a few explored different alternatives, and the reasoning for using a macro is provided in the original issue.

> Alright, I've embarked on implementing this. I've thought of a couple of different solutions, but they all have their pitfalls.
> 
> Weak links would be ideal, but either require nigthly until [rust-lang/rust#29603](https://github.com/rust-lang/rust/issues/29603) is implemented, or for the end user to have a c compiler on their machine, which I've read in the github action code isn't ideal.
> 
> I then thought to just emit a trait that has every function that then can be defined, but this has limitations on using codegen to implement all of the native functions with a default implementation.
> 
> This left me with an admittedly more complex solution then I'd like of passing in a macro to bindgen that then can generate the implementations.
> 
> It is not great code, but I'll open up a draft PR. Please feel free to give me feedback on the approach that I'm taking, and whether there should be any changes.

There are still a few things I'd like to change, but as a proof of concept you can do something like the following to automatically create a trait for all native stubs.

```rust
pub use crate::wit::Context;

mod wit {
    macro_rules! native_trait_stubs {
        ($stub:ty, $module:literal, $import:literal,
         fn $name:ident($($p:ident: $t:ty,)*)) => {
            <$stub as crate::wit::NativeStub>::$name($($p,)*)
        };
        ($stub:ty, $module:literal, $import:literal,
         fn $name:ident($($p:ident: $t:ty,)*) -> $ret:ty) => {
            <$stub as crate::wit::NativeStub>::$name($($p,)*)
        };
        (@definitions $stub:ty,
         $($module:literal, $import:literal, fn $name:ident($($p:ident: $t:ty),*) $(-> $ret:ty)?;)*
        ) => {
            pub trait NativeStub {
                $(fn $name($($p: $t),*) $(-> $ret)? {
                    unreachable!()
                })*
            }
        };
    }

    wit_bindgen::generate!({
        skip: ["init-plugin"],
        path: "../../wit",
        world: "plugin",
        native_stub_macro: native_trait_stubs!(crate::MyContextStub)
    });

    use super::Component;
    export!(Component);
}

struct MyContextStub {}
impl wit::NativeStub for MyContextStub {
    fn drop(_p0: i32) {
        todo!()
    }

    fn wit_import1(_p0: i32, _p1: *mut u8) {
        todo!()
    }
}
```

I have a minimum testing repo that can be found [here](https://github.com/BjornTheProgrammer/wit-native-plugin-system)

### DIFF
--- a/crates/guest-rust/macro/src/lib.rs
+++ b/crates/guest-rust/macro/src/lib.rs
@@ -9,7 +9,7 @@ use syn::{Token, braced, token};
 use wit_bindgen_core::AsyncFilterSet;
 use wit_bindgen_core::WorldGenerator;
 use wit_bindgen_core::wit_parser::{PackageId, Resolve, UnresolvedPackageGroup, WorldId};
-use wit_bindgen_rust::{Opts, Ownership, WithOption};
+use wit_bindgen_rust::{NativeStubMacro, Opts, Ownership, WithOption};
 
 #[proc_macro]
 pub fn generate(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
@@ -48,6 +48,7 @@ struct Config {
     world: WorldId,
     files: Vec<PathBuf>,
     debug: bool,
+    native_stub_macro: Option<NativeStubMacro>,
 }
 
 /// The source of the wit package definition
@@ -67,6 +68,7 @@ impl Parse for Config {
         let mut features = Vec::new();
         let mut async_configured = false;
         let mut debug = false;
+        let mut native_stub_macro = None;
 
         if input.peek(token::Brace) {
             let content;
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the PR, and I'm trying to page this all back in and understand the motivations/intentions here. Correct me if I'm wrong, but the high level idea is to be able to satisfy, on non-wasm platforms, wit-bindgen-generated bindings with a native implementation. Is that correct?

If so, I'm second-guessing this a bit in the sense of why go through ABI-marshalling at all? For example if you're calling a WIT function it seems like it'd be best to just turn around and immediately call a native function rather than serializing to some ABI-based representation just to deserialize and call a host function. Does that make sense to fit in here at all? Or am I perhaps missing something? (I could very well be)

BjornTheProgrammer:
This PR despite making the possible stubs very modular, suffers from a lot of issues, I'm going to close this in favor of another PR that I've been making. The idea is a bit different, and slightly more complex then just calling a host function, since that would require the host binary to be compiled with "-rdynamic" on linux, and it wouldn't even work on Windows and MacOS.

PR #1569, opened 2026-03-13
[rust] bindgen path supports env!() macro

kvcache:
This makes it possible to refer to a wit path by a path that includes
things like OUT_DIR or CARGO_MANIFEST_DIR. Relative paths that navigate
up, like `../wit/`, don't work well with cargo publish, and symlinks
have difficulties between git and cargo.

Using [macro-string](https://github.com/dtolnay/macro-string), this replaces the `path` option's `syn::LitStr` with MacroString.

See the example `crates/guest-rust/src/examples/_5_import_directory.rs` for
how this can be used. I need this capability so I can write separate crates for
a partially inter-related set of WIT's without having to copy/paste the whole
WIT tree for each package.

Before: `error: expected string literal`
After: Example 5 compiles.


kvcache:
For context, this is the expression I'm able to compile locally with this change:

```rust
wit_bindgen::generate!({
    world: "momento:bytes/imports",
    path: [ concat!(env!("OUT_DIR"), "/wit/host") ],
});
```

alexcrichton (commented):
Thanks! Would it be ok to place this behind a Cargo feature as well since it's an optional dependency. Perhaps an on-by-default optional feature?

kvcache:
> Thanks! Would it be ok to place this behind a Cargo feature as well since it's an optional dependency. Perhaps an on-by-default optional feature?

Thanks @alexcrichton, I added this as a default-on feature. I'm not sure why someone would disable it but the affordance is there now!

alexcrichton:
Thanks! Can you also edit [this dependency](https://github.com/bytecodealliance/wit-bindgen/blob/b858f8e0bfacf371dee1bbf7097b3022339feb93/crates/guest-rust/Cargo.toml#L22) to disable default features, and then also configure the default features on the `wit-bindgen` crate as well? Most users don't depend on the macro crate directly but instead only indirectly though the `wit-bindgen` crate.

kvcache:
@alexcrichton I think I got the feature lifted to `wit-bindgen` now, but I'm not the most experienced with cargo `feature`s. I also merged main, because github couldn't automatically figure out the line diff in cargo.toml with the 0.54.0 update.

PR #1576, opened 2026-03-31
C# macos support

silesmo:
Build and test it ourselves since it's not a published package based on: https://github.com/dotnet/runtimelab/issues/2119

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,11 +80,9 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         # moonbit removed from language matrix for now - causing CI failures
         lang: [c, rust, csharp, cpp, go]
-        exclude:
-          # For now csharp doesn't work on macos, so exclude it from testing.
-          - os: macos-latest
-            lang: csharp
     runs-on: ${{ matrix.os }}
+    env:
+      RUNTIMELAB_COMMIT: '4cac3ab5c8e97fda69c23dfca41ace964babc05e'
     steps:
     - uses: actions/checkout@v4
       with:
@@ -103,7 +101,14 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: '9.x'
-      if: matrix.lang == 'csharp'
+      if: matrix.lang == 'csharp' && runner.os != 'macOS'
+
+    - name: Setup .NET 10 (macOS)
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: '10.x'
+        dotnet-quality: 'preview'
+      if: matrix.lang == 'csharp' && runner.os == 'macOS'
 
     - name: Setup Go
       uses: actions/setup-go@v5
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -80,11 +80,9 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         # moonbit removed from language matrix for now - causing CI failures
         lang: [c, rust, csharp, cpp, go]
-        exclude:
-          # For now csharp doesn't work on macos, so exclude it from testing.
-          - os: macos-latest
-            lang: csharp
     runs-on: ${{ matrix.os }}
+    env:
+      RUNTIMELAB_COMMIT: '4cac3ab5c8e97fda69c23dfca41ace964babc05e'
```
yowl, line 85: Do we need to fix a commit?

.github/workflows/main.yml
```diff
@@ -125,6 +130,57 @@ jobs:
...
+        # Build wasi libs and packages
+        ./build.sh clr.aot+libs -c Release -a wasm -os wasi /p:NuGetAudit=false
+        ./build.sh nativeaot.packages -c Release -a wasm -os wasi /p:NuGetAudit=false
+
+        # Build host compiler, libs, WASM JIT, and packages
+        ./build.sh clr.aot -c Release /p:NuGetAudit=false
+        ./build.sh libs -c Release /p:NuGetAudit=false
+        LLVM_CMAKE_CONFIG_RELEASE=$(brew --prefix llvm@18)/lib/cmake/llvm \
+        src/coreclr/build-runtime.sh -arm64 -release -os osx -outputrid osx-arm64 -component llvmjit
+        cp artifacts/bin/coreclr/osx.arm64.Release/libclrjit_universal_llvm32_arm64.dylib artifacts/bin/coreclr/osx.arm64.Release/ilc-published/
```
yowl, line 171: Interesting, something must be missing in the runtimelab build to make this necessary.  Not saying anything needs changing, just if we want to push this to runtimelab, we will have to make the build copy these files.

crates/csharp/src/csproj.rs
```diff
@@ -103,21 +103,36 @@ impl CSProjectLLVMBuilder {
...
+            let ilc_version = env::var("ILC_VERSION").unwrap_or_else(|_| "10.0.0-*".to_string());
+
             csproj.push_str(
                 &format!(
                     r#"
         <ItemGroup>
-            <PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM" Version="10.0.0-*" />
-            <PackageReference Include="runtime.{os}-x64.Microsoft.DotNet.ILCompiler.LLVM" Version="10.0.0-*" />
+            <PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM" Version="{ilc_version}" />
+            <PackageReference Include="runtime.{os}-{arch}.Microsoft.DotNet.ILCompiler.LLVM" Version="{ilc_version}" />
```
yowl, line 123: This might be a problem for arm64 on windows as we don't have that either.

yowl (commented):
Just curious about fixing the commit.

yowl:
What is the effect on the build time?  An alternative is to push whatever might be needed to runtimelab and will build the nuget there.

.github/workflows/main.yml
silesmo: Maybe not, was just for consistency in case the latest version caused something to break. But we can remove it if you want?

crates/csharp/src/csproj.rs
silesmo: Yes this is true. I can add a not supported error for now?

silesmo:
> What is the effect on the build time? An alternative is to push whatever might be needed to runtimelab and will build the nuget there.

It took 36m 36s
https://github.com/bytecodealliance/wit-bindgen/actions/runs/23816345703/job/69416768215?pr=1576

But the build of runtimelab took 27m and 43s of that time and it gets cached so it wont take so long for following runs (This is actually a reason to pin it to a specific hash so we don't have to rebuild every time @yowl)

OptimusPi:
I'm all set up for compiling my exciting new project, a Balatro seed searcher in wasm. 

Windows build is great, I am set up on MacOS and just noticed all this stuff.

What should I do, I'd like to test something you have here just to offer another box to test on, 

Mac M1 Pro 32GB ram

Thanks!

.github/workflows/main.yml
yowl: On second thoughts, I agree, lets leave it.

crates/csharp/src/csproj.rs
yowl: host arm64 we can do on windows , we use the x64 version and emulation, so maybe condition the `{arch}` on macOS for now (until we get a native Windows Arm64 build)

yowl:
> What should I do, I'd like to test something you have here just to offer another box to test on,
> 
Up to you really, you can take this branch and build yourself, or wait until this is merged.

OptimusPi:
Thanks all, it works great...it's really lovely!

yowl:
@dicej @alexcrichton  What do you think about this and the extra time for the build, apparently it should cache the longest bit?

dicej:
> @dicej @alexcrichton What do you think about this and the extra time for the build, apparently it should cache the longest bit?

Seems fine to me if the extra time is minimal when the cache is populated.  Is there a plan to move this step over to runtimelab eventually like you mentioned above?

yowl:
There is no plan to move this to runtimelab due to no one able to support it there.


________________________________

alexcrichton:
I'm rerunning the macos job in CI to double-check that caching works. So long as this is reliable and at least not slower than the Windows C# job I think it's fine.

One minor ask: could the C#-specific setup bits of the CI configuration be moved to something like `.github/actions/setup-csharp/action.yml`? It's getting pretty large and it seems about the time to refactor that out to a dedicated script.

alexcrichton:
Hm, and it looks like the caching may not be working? On the retry it's rebuilding everything it looks like

.github/workflows/main.yml
OptimusPi, line 119: is the double build stuff able to be fixed from the pr being closed?

yowl: If you mean that if we fix the runtimelab issue then we can merge this one, then I would say yes.   At runtimelab we have limited resource to maintain an ios build, but we would accept a PR there on the basis that if it breaks in the future we would remove it if it was not trivial to fix.

PR #1580, opened 2026-04-04
[C#] Add cancel support for futures and streams

yowl:
This PR adds some support for canceling futures and streams..While dotnet `Task` are cancelable via tokens, that is not the design we want for components, so in this PR there is introduced a `ComponentTask` which is a merging of a `Task` and a `TaskCompletionSource` (and their generic counterparts) which adds a `Cancel` method.

In place of Rust's `Result<>` this uses a `ComponentTask` `Result` together with a `CancelCode` enum.
`LiftingTaskCompletionSource` is a placeholder for returning values from async functions.  It requires the lifting code, but hopefully can be used as a starting point.

Changes all task continuations to be `ExecuteSynchronously` as we only have one thread, and this is more predicable.  This should be removed if we get true multithreading.

The `index` suffix on imports is now created sequentially, no longer hard coded to `0`.

Adds the runtime test, `future-cancel-read`

### DIFF
--- a/crates/csharp/src/AsyncSupport.cs
+++ b/crates/csharp/src/AsyncSupport.cs
@@ -433,7 +433,7 @@ public abstract class ReaderBase : IFutureStream
         }
         else
         {
-            // TODO: crete buffers for lowered stream types and then lift
+            // TODO: create buffers for lowered stream types and then lift
             throw new NotImplementedException("reading from futures types that require lifting");
         }
     }


crates/csharp/src/AsyncSupport.cs
```diff
@@ -364,10 +424,24 @@ internal int TakeHandle()
...
+    protected GCHandle LiftBuffer<T>(T[] buffer)
+    {
+        // For primitive, blittable types
+        if (typeof(T).IsPrimitive || typeof(T).IsValueType)
+        {
+            return GCHandle.Alloc(buffer, GCHandleType.Pinned);
+        }
+        else
+        {
+            // TODO: crete buffers for lowered stream types and then lift
```
dicej:
```suggestion
            // TODO: create buffers for lowered stream types and then lift
```

crates/csharp/src/AsyncSupport.cs
```diff
@@ -467,22 +537,47 @@ public class FutureReader<T>(int handle, FutureVTable vTable) : ReaderBase(handl
...
-            // TODO: crete buffers for lowered stream types and then lift
-            throw new NotImplementedException("reading from futures types that require lifting");
-        }
+            if (it.IsCompletedSuccessfully)
+            {
+                readTask.SetResult(buf[0]);
+            }
+            else if (it.IsCanceled)
+            {
+                // readTask.SetCanceled();
```
dicej: Why is this commented out?

crates/csharp/src/AsyncSupport.cs
```diff
@@ -594,25 +677,23 @@ internal unsafe Task<int> WriteInternal(Func<GCHandle?> lowerPayload, int length
         bufferHandle = lowerPayload();
 
         var status = new WaitableStatus(VTableWrite(bufferHandle == null ? IntPtr.Zero : bufferHandle.Value.AddrOfPinnedObject(), length));
+        canDrop = true;  // We can only call drop once something has been written.
```
dicej, line 676: We should set this back to false if the write is cancelled.

yowl:
> > While dotnet Task are cancelable via tokens, that is not the design we want for components
> 
> Can you elaborate on this? What makes `CancellationToken` a bad fit here?

Its a cooperative mechanism: the long running task periodically checks the token to see if a cancellation has been requested, while typically another thread will issue the cancel on the token,  We don't have a loop in C# where we could periodically check the cancellation status of the token.

crates/csharp/src/AsyncSupport.cs
yowl: I'll delete this, we dont actually need to do anything in here, the cancellation is on the outer task.  We do call Cancel on this inner task, but its just for correctness rather than anything functional.

crates/csharp/src/AsyncSupport.cs
yowl: Thanks, done.

PR #1548, opened 2026-02-18
feat: migrate to vanity imports

asteurer:
@dicej @ricochet

PR #1583, opened 2026-04-10
feat(go): add map type support

yordis:


crates/test/src/go.rs
```diff
@@ -162,7 +161,7 @@ fn replace_bindings_go_mod(runner: &Runner, bindings_dir: &Path) -> Result<()> {
     super::write_if_different(
         &bindings_dir.join("go.mod"),
         format!(
-            "module wit_component\n\ngo 1.25\n\nreplace go.bytecodealliance.org => {}",
+            "module wit_component\n\ngo 1.25\n\nreplace go.bytecodealliance.org/pkg => {}",
```
yordis, line 164: related to https://github.com/bytecodealliance/wit-bindgen/pull/1548

As far as I can tell, it suppose to be `/pkg`?

alexcrichton:
cc @asteurer, you might have thoughts on this too

crates/go/src/lib.rs
```diff
@@ -2546,6 +2558,64 @@ lifters = append(lifters, func() {{
...
+        {body}
+        {result}[{body_key}] = {body_value}
+}}
+"
+                );
+                results.push(result);
+            }
+            Instruction::IterMapKey { .. } => results.push("map_key".into()),
+            Instruction::IterMapValue { .. } => results.push("map_value".into()),
+            Instruction::GuestDeallocateMap { .. } => {
```
dicej: Nit: could combine these cases to reduce duplication a bit, e.g.:

```
Instruction::GuestDeallocateMap { .. } | Instruction::GuestDeallocate { .. } => {
    // Nothing to do here; should be handled when calling `pinner.Unpin()`
}
```

crates/test/src/go.rs
dicej: Seems reasonable, but I wonder how the tests have been passing so far without that.  @asteurer might have ideas.

dicej (approved):
LGTM, thanks!

crates/go/src/lib.rs
yordis: @dicej done

crates/test/src/go.rs
asteurer: Yeah, that was definitely a mistake. I'm not sure why Go allows an incomplete prefix for a replace directive like that. Great catch!

asteurer (approved):
Seems good to me! Thanks!

PR #1586, opened 2026-04-13
Add `--enable-method-chaining` for the Rust generator.

### LINKED ISSUE #1559: Make resource function default return reference to self
In the rust guest implementaiton, it would be interesting to explore the idea of having the function return a reference to self if there isn't a return value. This would enable chaining like the following.

```wit
interface text {
    resource text-component {
        text: static func(plain: string) -> text-component;
        add-text: func(text: string);
        underlined: func(value: bool);
    }
}
```

```rust
fn main() {
    let text = TextComponent::text("My Component").underlined(true).add_text(" - my additional text here");
}
```

This could also be put under a feature flag in wit-bindgen.

QuantumSegfault:
Resolves #1559

Implements an option for Rust guest bindings generation to have methods with no return (`-> ()`) instead pass-through `-> &Self` to allow chaining.

For consistency, this also applies to exports, so all exports normally expecting empty returns must now return `self`.

### DIFF
--- a/crates/guest-rust/macro/src/lib.rs
+++ b/crates/guest-rust/macro/src/lib.rs
@@ -171,6 +171,9 @@ impl Parse for Config {
                         }
                         opts.async_ = val;
                     }
+                    Opt::EnableMethodChaining(enable) => {
+                        opts.enable_method_chaining = enable.value();
+                    }
                 }
             }
         } else {
@@ -324,6 +327,7 @@ mod kw {
     syn::custom_keyword!(disable_custom_section_link_helpers);
     syn::custom_keyword!(imports);
     syn::custom_keyword!(debug);
+    syn::custom_keyword!(enable_method_chaining);
 }
 
 #[derive(Clone)]
@@ -404,6 +408,7 @@ enum Opt {
     DisableCustomSectionLinkHelpers(syn::LitBool),
     Async(AsyncFilterSet, Span),
     Debug(syn::LitBool),
+    EnableMethodChaining(syn::LitBool),
 }
 
 impl Parse for Opt {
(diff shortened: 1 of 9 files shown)


QuantumSegfault:
@alexcrichton @BjornTheProgrammer @Snowiiii

QuantumSegfault:
Thinking about this more, this feels more like a hack/workaround than an actual solution.

It works, but it's invasive (makes ALL void returns chainable; even if there's no good reason).

It's also language dependent (each language needs to do something like this themselves).

WebAssembly/component-model#334 is basically what we are trying to solve. But conversation on that stagnated.

BjornTheProgrammer:
It is a bit of a hack due to the fact that the wit doesn't have the semantics to return the self parameter easily from a resource, but despite that I think it should still be included.

I think it is fine if this is something only used in the rust generator as well, since builder patterns like this are much more idiomatic in rust than in other languages. 

Even though the function makes all void functions return references to self, I don't really believe that this creates much tangible harm. Also since it is an optional binding feature, it makes it only opt in for projects that are willing to adopt it. Eventually when wit is extended to allow some more idiomatic way of doing this without such a hack, this could always be removed. 

The quality of life improvements for allowing this to be toggled in a rust project is immense. Currently a huge annoyance for people with our plugin system is that they need to write out each modification to a resource on a new line. 

These are just my thoughts, and I really appreciate your great work!

alexcrichton (commented):
Thanks! I agree this is a workaround at best but it's also the best we have right now, and I also agree that because this is opt-in the bar isn't too high for inclusion.

Some high level comments though:

* Can this expose the option via the `generate!` macro as well?
* There's a lot of duplication of the deduction of whether to return `self`, could that be refactored into a helper method on the `Rust` type or similar? Somethign which would probably take a `&Function` as input for example. That helper should then be called ideally instead of duplicating the logic and/or passing around a `bool`

QuantumSegfault:
That should do it on the macro side?

But how do I actually run the `crates/rust/tests/codegen.rs` tests?

alexcrichton:
You can run the tests locally with `cargo test -p wit-bindgen-rust` (see [CI here](https://github.com/bytecodealliance/wit-bindgen/actions/runs/24417721006/job/71331297221?pr=1586)).

Everything here looks reasonable to me, thanks! Was there anything else you wanted to test or do before landing?

QuantumSegfault:
Ah! Got it.

Nope. This PR is ready for merge. 👍

alexcrichton (approved):

Snowiiii:
@QuantumSegfault Thanks for the implementation 👍, @alexcrichton when do you plan to do a new release would appreciate to have this as soon as possible

alexcrichton:
I'll try to do a release by tomorrow or next monday at the latest

PR #1584, opened 2026-04-11
feat(moonbit): add map type support

yordis:
This PR adds map type support to the MoonBit backend, following the same pattern established in #1562 (core) and #1583 (Go).

PR #1590, opened 2026-04-14
feat(cpp): add map type support

yordis:
This PR adds map type support to the C++ backend, following the same pattern established in #1562 (core), #1583 (Go), and #1584 (MoonBit).


alexcrichton:
cc @cpetig and @TartanLlama

crates/cpp/src/lib.rs
```diff
@@ -3540,12 +3549,102 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
...
+                let val = format!("map{tmp}");
+                let ptr = format!("ptr{tmp}");
+                let len = format!("len{tmp}");
+                let idx = format!("idx{tmp}");
+                let entry_name = format!("entry{tmp}");
+                let entry = self.r#gen.sizes.record([*key, *value]);
+                let size = entry.size.format(POINTER_SIZE_EXPRESSION);
+                let align = entry.align.format(POINTER_SIZE_EXPRESSION);
+                self.push_str(&format!("auto&& {val} = {};\n", operands[0]));
+                self.push_str(&format!("auto {len} = (size_t)({val}.size());\n"));
```
TartanLlama: Can we avoid the C-style cast here?

crates/cpp/src/lib.rs
```diff
@@ -3540,12 +3549,102 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
...
+                uwriteln!(
+                    self.src,
+                    "auto {ptr} = ({ptr_type})({len} > 0 ? cabi_realloc(nullptr, 0, {align}, {len} * {size}) : nullptr);",
```
TartanLlama: Can we avoid the C-style cast here?

crates/cpp/src/lib.rs
```diff
@@ -3540,12 +3549,102 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
...
+                    uwriteln!(self.src, "if ({len}>0) _deallocate.push_back({base});");
+                }
+                uwriteln!(self.src, "for (unsigned i=0; i<{len}; ++i) {{");
+                uwriteln!(self.src, "auto base = {base} + i * {size};");
+                uwrite!(self.src, "{}", body.0);
+                let body_key = &body.1[0];
+                let body_value = &body.1[1];
+                uwriteln!(
+                    self.src,
+                    "{result}.insert(std::make_pair({}, {}));",
```
TartanLlama: I think `.emplace` would be better here

crates/cpp/src/lib.rs
```diff
@@ -435,6 +436,9 @@ impl Cpp {
         if self.dependencies.needs_bit {
             self.include("<bit>");
         }
+        if self.dependencies.needs_map {
+            self.include("<map>");
```
TartanLlama: I wonder if it would be better to have some configurability here? As a user, I'd often want to avoid `std::map` for performance reasons, and if I'm really perf-focused, I'd also avoid `std::unordered_map` and use some custom cache-friendly hash map type. Do you think we can enable this use-case without mandating an additional copy of the data?

crates/cpp/src/lib.rs
```diff
@@ -3540,12 +3549,102 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
...
+                self.push_str(&format!("auto&& {val} = {};\n", operands[0]));
+                self.push_str(&format!("auto {len} = (size_t)({val}.size());\n"));
+                uwriteln!(
+                    self.src,
+                    "auto {ptr} = ({ptr_type})({len} > 0 ? cabi_realloc(nullptr, 0, {align}, {len} * {size}) : nullptr);",
+                    ptr_type = self.r#gen.r#gen.opts.ptr_type()
+                );
+                uwriteln!(self.src, "size_t {idx} = 0;");
+                uwriteln!(self.src, "for (auto& {entry_name} : {val}) {{");
+                uwriteln!(self.src, "auto iter_map_key = {entry_name}.first;");
```
TartanLlama: Is there a reason we're copying the key and taking a reference to the value? Can we at least move the key?

crates/cpp/src/lib.rs
```diff
@@ -3540,12 +3549,102 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
...
+                uwriteln!(self.src, "auto& iter_map_value = {entry_name}.second;");
+                uwriteln!(self.src, "auto base = {ptr} + {idx} * {size};");
+                uwrite!(self.src, "{}", body.0);
+                uwriteln!(self.src, "++{idx};");
+                uwriteln!(self.src, "}}");
+                if realloc.is_some() {
+                    // ownership transfers
```
TartanLlama: I'm not sure what this comment means or what this block is for

crates/cpp/src/lib.rs
```diff
@@ -3540,12 +3549,102 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
...
+                let ptr = self.tempname("ptr", tmp);
+                let len = self.tempname("len", tmp);
+                uwriteln!(self.src, "uint8_t* {ptr} = {};", operands[0]);
+                uwriteln!(self.src, "size_t {len} = {};", operands[1]);
+                let i = self.tempname("i", tmp);
+                uwriteln!(self.src, "for (size_t {i} = 0; {i} < {len}; {i}++) {{");
+                let entry = self.r#gen.sizes.record([*key, *value]);
+                let size = entry.size.format(POINTER_SIZE_EXPRESSION);
+                uwriteln!(self.src, "uint8_t* base = {ptr} + {i} * {size};");
+                uwriteln!(self.src, "(void) base;");
```
TartanLlama: Why do we need this line? In case the body doesn't use `base`?

crates/cpp/src/lib.rs
TartanLlama: It seems that all the allowed key types from the spec would be hashable, so at least it may be better to default to `std::unordered_map`

yordis:
@TartanLlama addressing your comments, the "why" behind things, I am effectively Junior in C++, please be patience with me here 🙏🏻 I am try to understand the intent and I will give it a second path

cpetig:
If I were to implement this feature I would introduce a wit::map type which has the same layout as the lowered data. This way the data doesn't need to be copied unless the user code requests it (making map a lower cost abstraction).

On the other hand I just ran into an adjacent problem with list lowering (and avoiding re-allocation) in https://github.com/bytecodealliance/wit-bindgen/issues/1592 , so avoiding a copy in case of complex types might still be impossible - due to C++ type layout restrictions.

A `std::map` is of course more easy to use if you don't care about allocation and copy costs.

yordis:
@cpetig @TartanLlama can you give it a pass again? 🙏🏻

Ideally, please share the intent of what you would like to see, I can figure out eventually the implementation but I am not qualify to the best practices or insights of C++. I can figure that out for sure, it would help to declare the intent so I can go off, study, and figure out what you want 😄

crates/cpp/helper-types/wit.h
```diff
@@ -170,6 +171,70 @@ template <class T> class vector {
...
+
+private:
+  entry_type *data_;
+  size_t length;
+
+  static entry_type* empty_ptr() { return (entry_type*)alignof(entry_type); }
+
+public:
+  map(map const &) = delete;
+  map(map &&b) : data_(b.data_), length(b.length) { b.data_ = nullptr; }
```
TartanLlama, line 193: Could zero the length as well

crates/cpp/helper-types/wit.h
```diff
@@ -170,6 +171,70 @@ template <class T> class vector {
...
+  // WARNING: map contains uninitialized entries; caller must construct them via
+  // `initialize` before the map is observed or destroyed.
+  static map<K, V> allocate(size_t len) {
+    if (!len) return map<K, V>(empty_ptr(), 0);
+    return map<K, V>((entry_type*)malloc(sizeof(entry_type) * len), len);
+  }
+  void initialize(size_t n, entry_type&& entry) {
+    new ((void*)(data_ + n)) entry_type(std::move(entry));
+  }
+  entry_type* leak() { entry_type* result = data_; data_ = nullptr; return result; }
```
TartanLlama, line 228: Very long line, can you run clang-format on this?

TartanLlama:
Hmm I think that providing indexing functions for `wit::map` is a potential footgun. If the key type is an integral type, users who don't check the docs would likely assume that the argument to `operator[]` would be a key, rather than an index, but the code would compile fine. I think if we provide a zero-additional-copy `wit::map` type, it would have to have no indexing functions. 

But I think in general there's a big tradeoff in performance vs ergonomics here. Ideally, I think we'd be able to configure whether we wanted a low-copy, low-usability type like `wit::map` vs a high-usability type that requires extra copies like `std::unordered_map`. What do you think @cpetig ?

yordis:
I am gonna wait until @TartanLlama and @cpetig agree on the direction

crates/cpp/helper-types/wit.h
```diff
@@ -170,6 +171,70 @@ template <class T> class vector {
...
+    data_ = b.data_;
+    length = b.length;
+    b.data_ = nullptr;
+    return *this;
+  }
+  map(entry_type *d, size_t l) : data_(d), length(l) {}
+  map() : data_(empty_ptr()), length() {}
+  entry_type const *data() const { return data_; }
+  entry_type *data() { return data_; }
+  entry_type &operator[](size_t n) { return data_[n]; }
```
cpetig: I agree, this should not exist. 

Indexing into a map would be expected to take a `K const&` argument and then do a linear search.

crates/cpp/helper-types/wit.h
```diff
@@ -170,6 +171,70 @@ template <class T> class vector {
...
+  static map<K, V> allocate(size_t len) {
+    if (!len) return map<K, V>(empty_ptr(), 0);
+    return map<K, V>((entry_type*)malloc(sizeof(entry_type) * len), len);
+  }
+  void initialize(size_t n, entry_type&& entry) {
+    new ((void*)(data_ + n)) entry_type(std::move(entry));
+  }
+  entry_type* leak() { entry_type* result = data_; data_ = nullptr; return result; }
+  static void drop_raw(void *ptr) { if (ptr != empty_ptr()) free(ptr); }
+  std::span<entry_type> get_view() const { return std::span<entry_type>(data_, length); }
```
cpetig: The `std::span` functions are only meaningful for a `vector`, as far as I can foresee. I would not expect a map to constructed from a span of pairs (or turned into a span of pairs), but I haven't written a test program, yet.

crates/cpp/src/lib.rs
```diff
@@ -1741,7 +1741,33 @@ impl CppInterfaceGenerator<'_> {
...
+                            Flavor::BorrowedArgument
+                        }
+                        _ => Flavor::InStruct,
+                    };
+                    let k = self.type_name(key, from_namespace, element_flavor);
+                    let v = self.type_name(value, from_namespace, element_flavor);
+                    match flavor {
+                        Flavor::BorrowedArgument => {
+                            self.r#gen.dependencies.needs_span = true;
+                            format!("std::span<std::pair<{k}, {v}> const>")
```
cpetig: :thinking: a span works like a vector, but here you would like a span like object (spans only borrow) which mimics a map (likely unordered)

cpetig (commented):
Don't feel disheartened by the complexity of designing proper language bindings, it took me more than a year to come up with something I mostly liked - and then @TartanLlama made a significant improvement by pointing out that moved arguments can be received into normal objects.

Sorry, but writing proper bindings is simply complex, and mapping map is more so.

cpetig:
> Hmm I think that providing indexing functions for `wit::map` is a potential footgun. If the key type is an integral type, users who don't check the docs would likely assume that the argument to `operator[]` would be a key, rather than an index, but the code would compile fine. I think if we provide a zero-additional-copy `wit::map` type, it would have to have no indexing functions.

I would not want to have functions which aren't part of [the standard (unordered_)map](https://cppreference.com/cpp/container/unordered_map) - unless they are needed to construct the object within the bindings (or help construct these types in user code). 

> But I think in general there's a big tradeoff in performance vs ergonomics here. Ideally, I think we'd be able to configure whether we wanted a low-copy, low-usability type like `wit::map` vs a high-usability type that requires extra copies like `std::unordered_map`. What do you think @cpetig ?

I don't know which to prefer. This usually indicates that a generator option is a good solution. But I am quite sure that `wit::unordered_map` would be a better name for this thing. For C/C++/Rust I typically go with avoiding copying in the bindings even if the types get a little more difficult to use. In user code you can always add copying to more convenient types, but you can never remove it once it is done in the bindings.

Also I feel that wit::vector vs std::vector and wit::string vs std::string would be selected by the same option. This feels out of scope for just implementing https://github.com/WebAssembly/component-model/pull/554

yordis:
@TartanLlama @cpetig give it another pass please 🙏🏻

yordis:
I can't retry CI @alexcrichton

yordis:
@TartanLlama @cpetig pinging just in case you missed it, CI is green

crates/cpp/helper-types/wit.h
```diff
@@ -170,6 +171,107 @@ template <class T> class vector {
...
+/// @brief A map stored as a contiguous array of key-value pairs in linear
+/// memory, freed unconditionally using free.
+///
+/// Mirrors the canonical ABI representation of `map<K, V>` (`list<tuple<K, V>>`)
+/// to enable lift without per-entry tree allocation. The container has no
+/// ordering or hashing guarantees and exposes only the subset of the
+/// `std::unordered_map` API that's meaningful over a flat pair buffer plus
+/// the bindings-construction primitives (`allocate`, `initialize`, `leak`,
+/// `drop_raw`, `data`).
+template <class K, class V> class unordered_map {
```
TartanLlama, line 183: Why `unordered_map` when the entries are ordered?

crates/cpp/helper-types/wit.h
```diff
@@ -170,6 +171,107 @@ template <class T> class vector {
...
+
+/// @brief A non-owning, read-only view over a contiguous run of key-value
+/// pairs in linear memory.
+///
+/// Counterpart to `wit::unordered_map<K, V>` for borrowed arguments: a
+/// borrow-only handle that mimics map semantics rather than vector semantics
+/// — entries are reachable via range-for or `data()`/`size()`, with no
+/// positional `operator[]`. Construct from a `wit::unordered_map<K, V>` or
+/// an explicit `(data, size)` pair.
+template <class K, class V> class unordered_map_view {
```
TartanLlama: I wonder if this would be better as simply `std::span<const std::pair<K,V>>`

crates/cpp/helper-types/wit.h
yordis: The name reflects WIT `map<K, V>` semantics, not the layout: it's a set of key/value associations with no ordering guarantee, same as `std::unordered_map`.

The contiguous, ordered-looking buffer is incidental, the canonical ABI is `list<tuple<K, V>>`, so entries land in whatever order the producer serialized them. We keep that array (instead of a tree) only so lift is allocation-free; the order isn't meaningful or guaranteed.

Calling it `map` would wrongly imply `std::map`'s key-sorted semantics, which we don't provide.

yordis:
@TartanLlama ready again

yordis:
Hey @cpetig @TartanLlama 👋🏻 alo! Can you share some love here?

TartanLlama (approved):

alexcrichton (approved):

PR #1591, opened 2026-04-16
Remove dependency on `std` for `async` feature

alexcrichton:
This commit adds `no_std` support for the `async` feature of the `wit-bindgen` crate, notably enabling this to be used as part of the standard library for wasip3.


crates/guest-rust/src/rt/async_support/future_support.rs
```diff
@@ -550,15 +561,16 @@ impl<O: FutureOps> RawFutureWriter<O> {
...
+                    unsafe { Pin::new_unchecked(&mut *self.write.get()).poll(&mut cx) }
                 };
                 if poll.is_ready() {
                     assert_eq!(Arc::strong_count(&self), 1);
                 } else {
                     assert!(Arc::strong_count(&self) > 1);
                 }
                 assert_eq!(Arc::weak_count(&self), 0);
+
+                self.in_use.store(false, Release);
```
dicej: Should we use RAII for this (or the `with` pattern you used for `WakerState`)?  I get a little nervous about someone later adding an early return above and not noticing this.

dicej (approved):

crates/guest-rust/src/rt/async_support/future_support.rs
alexcrichton: Seems reasonable yeah 👍

PR #1601, opened 2026-04-18
fix(rust): qualify world-owned type paths in future/stream payload vtables

### LINKED ISSUE #1598: Bug: `future<result<T, ...>>` where T is from a `use` produces unresolved type references
# Bug: `future<result<T, ...>>` where T is from a `use` produces unresolved type references

## Summary

Using a record type via `use types.{T}` at world level, then embedding it in `future<result<T, ...>>` on an exported function, generates Rust bindings that reference `T` without the necessary `use super::super::T;` in the emitted `wit_future::vtableN` module.

Related prior issue: #1433 (closed, fixed by #1528) — that fix covered `stream<T>` but the same code path for `future<result<T, _>>` was missed.

## Environment

- `wit-bindgen` 0.57.1 (current release)
- rustc nightly, edition 2024
- Reproduced on `wasm32-wasip2`; the macro emits the same invalid code path regardless of target, so `wasm32-wasip3` with `-Zbuild-std=core,alloc` will hit the same compilation failure (the bug is in code generation, not target-specific).

## Reproducer

```rust
wit_bindgen::generate!({
    inline: r#"
        package a:b;

        world w {
            use t.{r};
            export f: async func() -> future<result<r, string>>;
        }

        interface t {
            record r {
                x: u32,
            }
        }
    "#,
});

fn main() {}
```

Build (any Wasm target; `wasm32-wasip2` shown for brevity):

```console
$ cargo build --target wasm32-wasip2
error[E0425]: cannot find type `R` in this scope
 --> src/main.rs:1:1
  |
1 | / wit_bindgen::generate!({
...|
16| | });
  | |__^ not found in this scope
  |
  = help: consider importing one of these items:
          crate::R
          crate::a::b::t::R
```

## Root cause

The generated code contains (pared down):

```rust
pub type R = a::b::t::R;   // at crate root

pub trait Guest {
    async fn f(...) -> Result<(..., wit_bindgen::rt::async_support::FutureReader<Result<R, _rt::String>>), _rt::String>;
    //                                                                              ^ resolves fine via the type alias
}

pub mod a { pub mod b { pub mod t { pub struct R { ... } } } }

pub mod wit_future {
    pub mod vtable1 {
        // ↓ HERE — `R` is bare, but we are two modules deep.
        unsafe fn lift(ptr: *mut u8) -> Result<R, super::super::_rt::String> { ... }
        //                                     ^ unresolved
    }
}
```

Inside `wit_future::vtable1`, `_rt` is correctly qualified as `super::super::_rt`, but `R` is left bare. The generator appears to treat `R` as living in the enclosing module without emitting the corresponding `use super::super::R;` (or using a fully qualified path).

Replacing `future<result<r, string>>` with `stream<r>` — the pattern fixed by #1528 — generates correctly qualified code, so the fix clearly addressed the stream path but not the future-of-result path.

## Expected

Either the emitted `lift` signature should read `-> Result<super::super::R, super::super::_rt::String>`, or the `vtable1` module should start with `use super::super::R;` (matching how `_rt` is handled).

mattwilkinsonn:
Fixes #1598.

Reworked per review: instead of injecting `use` statements into the generated `vtable{N}` module, extend `type_path_with_name` to qualify world-owned types with `path_to_root()`. That's `""` normally and `"super::super::"` in `Identifier::StreamOrFuturePayload` mode, so a world-level `use t.{r};` inside `future<result<r, _>>` now resolves inside the vtable scope.

Regression test in `tests/codegen.rs`: fails on main with `cannot find type R in this scope`.

### DIFF
--- a/crates/rust/src/interface.rs
+++ b/crates/rust/src/interface.rs
@@ -1887,6 +1887,11 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8) 
                 return format!("{path}::{name}");
             }
         }
+        // World-level aliases live at macro root. `path_to_root()` is
+        // `""` at root, `"super::super::"` in stream/future payload mode.
+        if let TypeOwner::World(_) = self.resolve.types[id].owner {
+            return format!("{}{name}", self.path_to_root());
+        }
         name
     }
 
--- a/crates/rust/tests/codegen.rs
+++ b/crates/rust/tests/codegen.rs
@@ -235,3 +235,25 @@ mod method_chaining {
         enable_method_chaining: true
     });
 }
+
+// Regression for #1598: world-level `use t.{r};` inside
+// `future<result<r, _>>` used to emit unresolved bare `R` in vtable{N}.
+#[allow(unused, reason = "testing codegen, not functionality")]
+mod issue_1598_future_result_use {
+    wit_bindgen::generate!({
+        inline: r#"
+        package a:b;
+
+        world w {
+            use t.{r};
+            export f: func() -> future<result<r, string>>;
+        }
+
+        interface t {
+            record r {
+                x: u32,
+            }
+        }
+        "#,
+    });
+}
--- a/crates/rust/tests/codegen.rs
+++ b/crates/rust/tests/codegen.rs
@@ -235,25 +235,3 @@ mod method_chaining {
         enable_method_chaining: true
     });
 }
-
-// Regression for #1598: world-level `use t.{r};` inside
-// `future<result<r, _>>` used to emit unresolved bare `R` in vtable{N}.
-#[allow(unused, reason = "testing codegen, not functionality")]
-mod issue_1598_future_result_use {
-    wit_bindgen::generate!({
-        inline: r#"
-        package a:b;
-
-        world w {
-            use t.{r};
-            export f: func() -> future<result<r, string>>;
-        }
-
-        interface t {
-            record r {
-                x: u32,
-            }
-        }
-        "#,
-    });
-}
--- a/crates/test/src/cpp.rs
+++ b/crates/test/src/cpp.rs
@@ -45,6 +45,10 @@ impl LanguageMethods for Cpp {
         config: &crate::config::WitConfig,
         _args: &[String],
     ) -> bool {
+        // Compiles on C++ despite the blanket async exclusion below.
+        if name == "issue-1598.wit" {
+            return false;
+        }
         return match name {
             "issue1514-6.wit" | "named-fixed-length-list.wit" | "map.wit" => true,
             _ => false,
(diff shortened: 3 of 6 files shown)


alexcrichton (commented):
Thanks for the PR! I'm not sure though that the fix here is the best fix, though. I would expect this to not import more types (as that can clash), but instead to update the path used to refer to a type (e.g. a missing `super::` or something like that). IIRC these type paths are printed under a special mode which understands that it's in this nested module, so I think this may be relatively easily possible?

Another possible alternative is to use `const _: () = { /* ... stuff ... */ };` as a scoping mechanism instead of a module. I'm not sure if this is exactly applicable though and I'd have to dig in.

Also, FWIW, I'd appreciate to leave out the AI summaries/generation here. I'm happy to work with you, but wading through reams of AI-generated text is not always easy to do.

mattwilkinsonn:
> Thanks for the PR! I'm not sure though that the fix here is the best fix, though. I would expect this to not import more types (as that can clash), but instead to update the path used to refer to a type (e.g. a missing `super::` or something like that). IIRC these type paths are printed under a special mode which understands that it's in this nested module, so I think this may be relatively easily possible?
> 
> Another possible alternative is to use `const _: () = { /* ... stuff ... */ };` as a scoping mechanism instead of a module. I'm not sure if this is exactly applicable though and I'd have to dig in.
> 
> Also, FWIW, I'd appreciate to leave out the AI summaries/generation here. I'm happy to work with you, but wading through reams of AI-generated text is not always easy to do.

Sure, I'll look into changing it to update the path.

Also noted on the too long AI description - i'll cut it down and rewrite it.

mattwilkinsonn:
> Thanks for the PR! I'm not sure though that the fix here is the best fix, though. I would expect this to not import more types (as that can clash), but instead to update the path used to refer to a type (e.g. a missing `super::` or something like that). IIRC these type paths are printed under a special mode which understands that it's in this nested module, so I think this may be relatively easily possible?
> 
> Another possible alternative is to use `const _: () = { /* ... stuff ... */ };` as a scoping mechanism instead of a module. I'm not sure if this is exactly applicable though and I'd have to dig in.
> 
> Also, FWIW, I'd appreciate to leave out the AI summaries/generation here. I'm happy to work with you, but wading through reams of AI-generated text is not always easy to do.

@alexcrichton updated the fix to extend type_path_with_name to qualify world-owned types with path_to_root(). Rewrote PR desc as well. Let me know if there's any other changes you want here

alexcrichton (approved):
Looks great, thanks! As one final request, could the test move to the `tests/codegen/*.wit` location? That way it'll be tested under a number of modes and easily made available to test for other languages which can be beneficial

alexcrichton (approved):

mattwilkinsonn:
@alexcrichton some of the other languages break on this test - looks like go & c++. Ok to revert to just being a Rust test for now?

alexcrichton:
Could exceptions for C++/Go be added as well in a similar manner as C#?

mattwilkinsonn:
> Could exceptions for C++/Go be added as well in a similar manner as C#?

Oh right, that would be much better, just pushed that up will see if it builds now. I had to do the C++ one slightly differently b/c async is supposed to be rejected but this wit file compiled

PR #1603, opened 2026-04-21
feat(csharp): add map type support

yordis:
- `Dictionary<K,V>` provides a familiar, zero-dependency host representation for WIT maps in C# and unblocks runtime interop with components that use them without requiring a custom collection type.
- Parity with the Rust, Go, and MoonBit backends that already ship map support, reducing the per-language divergence in `should_fail_verify` as new features land.


yowl (commented):
Thanks!  Do we need the cargo.lock changes?

crates/csharp/src/world_generator.rs
```diff
@@ -856,6 +856,18 @@ impl WorldGenerator for CSharp {
                 );
             }
 
+            files.push(
+                &format!("{world_namespace}_component_type.o"),
+                wit_bindgen_c::component_type_object::object(
```
yowl: Why does adding map support mean we need this library?

crates/csharp/src/world_generator.rs
```diff
@@ -814,14 +814,6 @@ impl WorldGenerator for CSharp {
                 );
             }
 
-            // For the time being, we generate both a .wit file and a .o file to
```
yordis, line 828: @yowl this got me foncused which is related to your question, I removed the .o file

yordis: Should I put this comment back? Maybe I misunderstood, if so, my apologies

yordis:
@yowl can you give it a second path?

yowl:
Looks good, one last thing, is the cargo.lock change required now?

yordis:
@yowl fixed

yowl (approved):
Looks good, thanks!

PR #1606, opened 2026-04-28
[C#] Add support for the ping pong test, bring support for futures that return strings, i.e. types returned via pointers.

yowl:
As per the title, this PR adds more support to the C# wit-bindgen to get the ping-pong test to pass.  As part of this I've:

* Added more things to the VTable: size, alignment, lower and lift.
* Added support for lifting and lowering, tested for strings only.
* Removed `ICancelableRead` and `ICancelableWrite` and replaced with `IVTable`
* Removed the `FreeBuffer` method and replaced with `cleanup` actions

### DIFF
--- a/crates/csharp/src/AsyncSupport.cs
+++ b/crates/csharp/src/AsyncSupport.cs
@@ -133,9 +133,8 @@ public static class AsyncSupport
     }
 
     // unsafe because we are using pointers.
-    public static unsafe int Callback(EventWaitable e, ContextTask* contextPtr)
+    public static unsafe int Callback(EventWaitable e)
     {
-        Console.WriteLine("Callback");
         ContextTask* contextTaskPtr = ContextGet();
 
         var waitables = pendingTasks[contextTaskPtr->WaitableSetHandle];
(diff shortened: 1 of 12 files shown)


crates/csharp/src/AsyncSupport.cs
```diff
@@ -130,16 +133,17 @@ public static unsafe void ContextSet(ContextTask* contextTask)
     }
 
     // unsafe because we are using pointers.
-    public static unsafe uint Callback(EventWaitable e, ContextTask* contextPtr)
+    public static unsafe int Callback(EventWaitable e, ContextTask* contextPtr)
     {
+        Console.WriteLine("Callback");
```
dicej: Looks like a stray debug print statement.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -148,13 +152,81 @@ public static unsafe uint Callback(EventWaitable e, ContextTask* contextPtr)
...
-                // TODO: Handle/lift async function return values. 
-                waitableInfoState.SetResult(0 /* not used */);
+                if (e.SubtaskStatus.IsStarting)
+                {
+                    throw new Exception("unexpected subtask status Starting " + e.Code);
+                }
+                if (e.SubtaskStatus.IsStarted || e.SubtaskStatus.IsReturned)
+                {
+                    waitableInfoState.SetResult(e.WaitableCount);
+                    Interop.SubtaskDrop(e.Waitable);
```
dicej: I'm surprised to see a `SubtaskDrop` call here for `e.SubtaskStatus.IsStarted`.  Don't we want to wait until it reaches the `e.SubtaskStatus.IsReturned` state?

dicej: Also, the line above seems like it should only apply to `e.SubtaskStatus.IsReturned` as well.

crates/csharp/src/AsyncSupport.cs
```diff
@@ -148,13 +152,81 @@ public static unsafe uint Callback(EventWaitable e, ContextTask* contextPtr)
...
+            }
+
+            return (int)CallbackCode.Wait | (int)(contextTaskPtr->WaitableSetHandle << 4);
+        }
+
+        throw new NotImplementedException($"WaitableStatus not implemented {e.WaitableStatus.State} in set {contextTaskPtr->WaitableSetHandle}");
+    }
 
+    // unsafe because we are using pointers.
+    public static unsafe int Callback<T>(EventWaitable e, Func<T> liftFunc)
```
dicej: There's a lot of code duplicated between the two overloads of `Callback`.  Would it be practical to combine them, e.g. by having the non-generic one defer to the generic one using a dummy type of some sort?

crates/csharp/src/function.rs
```diff
@@ -1768,6 +1873,13 @@ impl Bindgen for FunctionBindgen<'_, '_> {
     }
 }
 
+// TODO: this is not great, we want the underlying parameter, but it is passed in operands already lifted.
+pub fn strip_lift(lifted_param: &String) -> String {
```
dicej, line 1878: I'll admit I don't understand why this is needed or what it is doing.  Could you elaborate?

crates/csharp/src/interface.rs
```diff
@@ -876,6 +963,7 @@ var {async_status_var} = {raw_name}({wasm_params});
             [global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "[callback]{export_name}")]
             public static unsafe uint {camel_name}Callback(int eventRaw, uint waitable, uint code)
             {{
+            Console.WriteLine("Callback for {export_name} creating EventWaitable with code " + eventRaw + " for waitable " + waitable + " with code " + code);
```
dicej: Did you mean to leave this in here?

crates/csharp/src/interface.rs
```diff
@@ -1781,6 +1961,25 @@ impl<'a> CoreInterfaceGenerator<'a> for InterfaceGenerator<'a> {
...
+        | Type::S16
+        | Type::U16
+        | Type::S32
+        | Type::U32
+        | Type::S64
+        | Type::U64
+        | Type::Char
+        | Type::F32
+        | Type::F64 => false,
+        _ => true,
```
dicej, line 1967: FYI, this will probably need to be refined; e.g. resource handles and other types may not need pointers either.

dicej (commented):
Thanks for continuing to push this forward!

crates/csharp/src/function.rs
yowl: ah yes, I meant to talk about this in the description.  This stems from the chat we had and the idea to use an async function for the exported function.  In C# `UnmanagedCallersOnly` functions cannot be marked `async` so what I did was from the UCO function, we call to an `async` function so we can do the cleanups easily.  Without this `strip_lift` function, that pattern ends up like this
```C#
[global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "[async-lift]a:b/i#one-argument")]
public static unsafe int wasmExportOneArgument(int p0) {

    var task = OneArgumentAsync((unchecked((uint)(p0))));
    if (task.IsCompletedSuccessfully)
    {
        return (int)CallbackCode.Exit;
    }

    // TODO: Defer dropping borrowed resources until a result is returned.
    ContextTask* contextTaskPtr = AsyncSupport.ContextGet();

    return (int)CallbackCode.Wait | (int)(contextTaskPtr->WaitableSetHandle << 4);
}

public static async Task OneArgumentAsync(uint unchecked((uint)(p0)))
{
    var cleanups = new global::System.Collections.Generic.List<global::System.Action>();

    await IExportsImpl.OneArgument((unchecked((uint)(p0))));
    // TODO: task_cancel.forget();
    OneArgumentTaskReturn();

    {
        cleanup();
    }

}
```
The signature to `OneArgumentAsync` has the lifted parameter expression and is not valid.

crates/csharp/src/AsyncSupport.cs
yowl: Ah, yes, I misread the `go` equivalent, not realising it doesn't use a `break` statement between cases.

crates/csharp/src/interface.rs
yowl: Thanks, added a TODO for now.

crates/csharp/src/interface.rs
yowl: Removed this and other debugs

crates/csharp/src/AsyncSupport.cs
yowl: Turns out that we don't need the generic version at all as the lift func parameter is not used here.  Was an idea that later got dropped.

crates/csharp/src/AsyncSupport.cs
yowl: removed

crates/csharp/src/function.rs
dicej: I see, thanks for explaining.  Perhaps you could expand the comment to include an example of what the input and output of the regex would be, e.g. `unchecked((uint)(p0))`->`p0`.

yowl: done.

dicej (approved):

yowl:
going to try again as looked like connectivity issue

PR #1611, opened 2026-05-13
feat(c): add map type support

yordis:
- The flat `{ entry_t *ptr; size_t len; }` representation matches the canonical ABI's lowered layout for `map<K, V>` exactly, so reusing the existing list lower/lift fast-path keeps the FFI boundary copy-free and the implementation stays consistent with `crates/c`'s existing list/string idiom.
- Parity with the Rust, Go, MoonBit, and C# backends that already ship map support, removing the last `TypeDefKind::Map(..) => todo!()` arms from `crates/c` and the `map.wit` entry from `crates/test/src/c.rs`'s `should_fail_verify` skip list.

### DIFF
--- a/crates/test/src/c.rs
+++ b/crates/test/src/c.rs
@@ -56,9 +56,7 @@ impl LanguageMethods for C {
         config: &crate::config::WitConfig,
         _args: &[String],
     ) -> bool {
-        config.error_context
-            || name.starts_with("named-fixed-length-list.wit")
-            || name.starts_with("map.wit")
+        config.error_context || name.starts_with("named-fixed-length-list.wit")
     }
 
     fn codegen_test_variants(&self) -> &[(&str, &[&str])] {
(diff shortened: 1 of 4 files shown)


crates/test/src/c.rs
```diff
@@ -56,6 +56,12 @@ impl LanguageMethods for C {
         config: &crate::config::WitConfig,
         _args: &[String],
     ) -> bool {
+        // `map.wit` exercises `tuple<names-by-id, u64>` where `names-by-id`
```
yordis: I need help here

alexcrichton: Could you test out https://github.com/bytecodealliance/wit-bindgen/pull/1624, which came up with a totally different issue but seems related, and see if it fixes this?

alexcrichton (commented):
Thanks again for this! This all looks reasonable to me and I'd be ok landign this as-is without the `map.wit` test comment (that seems like more of a transient comment)

alexcrichton (approved):

PR #1613, opened 2026-05-15
fix: add defer pinner.Unpin()

asteurer:
This fixes the following panic I'm getting when using async WIT functions:

```
 panic: runtime error: runtime.Pinner: found leaking pinned pointer; forgot to call Unpin()?

goroutine 6 [running]:
runtime.runFinalizers()
    /home/asteurer/.cache/componentize-go/v2/go-linux-amd64-bootstrap/src/runtime/mfinal.go:272 +0x5e
error while executing at wasm backtrace:
0: 0x11d931 - <unknown>!runtime.abort
1: 0x95f45 - <unknown>!runtime.fatalpanic
2: 0x113d55 - <unknown>!runtime.gopanic
3: 0x1201a8 - <unknown>!wasm_pc_f_loop_export
4: 0x18c0de - <unknown>!_async_lift_my_func
```

@dicej

### DIFF
--- a/crates/go/src/lib.rs
+++ b/crates/go/src/lib.rs
@@ -1326,7 +1326,10 @@ func {camel}({go_params}) {go_results} {{
 
             (
                 if abi::guest_export_needs_post_return(resolve, func) {
-                    format!("{PINNER} := &runtime.Pinner{{}}")
+                    format!(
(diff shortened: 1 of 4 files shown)


dicej (approved):
Oops, thanks for catching this.

Would you be up for adding a test case that covers this?

PR #1619, opened 2026-05-27
C++: sort types topologically before emitting definitions

### LINKED ISSUE #1615: C++: `undeclared identifier` errors building generated code
The code for this issue is at: https://github.com/itowlson/spin-cpp-wasip2-test

I am trying to build an HTTP handler using WASI 0.2.  I used the C++ binding generator from wit-bindgen 0.57.1:

```
wit-bindgen cpp ./wit/ --out-dir bindings -w http-trigger
```

and the WASI-SDK 33 compiler:

```
$(WASI_SDK_PATH)/bin/clang++ -std=c++20 -I bindings -I expected -target wasm32-wasip2 test.cpp bindings/http_trigger.cpp bindings/http_trigger_component_type.o -mexec-model=reactor -o test.wasm
```

I have patched up `expected` and `wit.h`, which has gotten rid of most the first tranche of errors.  However, I still get a lot of build errors for generated types for some of my WITs (I think for the Spin APIs, rather than part of WASI), e.g.

```
./bindings/http_trigger_cpp.h:300:42: error: use of undeclared identifier 'Error'
  300 |         static std::expected<Connection, Error> Open(std::string_view database);
```

(The WIT interface here, an API for SQLite, defines an `error` variant.  The `.h` file _does_ contain an `Error` type, albeit _after_ this line of code.)

Am I doing something wrong (e.g. wit-bindgen or clang options)?  How can I get this generated code to build?  Thanks!

(And I'm happy to offer a README for the C++ generator if I can get this working.  I could only find one for C so far.)

dicej:
Specifically, we sort the types while taking the parameter and return types of resource functions into consideration.  This ensures that types are declared (and, if necessary, defined) before they are used in the emitted source code.

Fixes #1615

### DIFF
--- a/crates/cpp/src/lib.rs
+++ b/crates/cpp/src/lib.rs
@@ -21,6 +21,7 @@ use wit_bindgen_core::{
         WorldKey,
     },
 };
+use wit_parser::TypeIdVisitor;
 
 // mod wamr;
 mod symbol_name;
@@ -879,11 +880,10 @@ impl CppInterfaceGenerator<'_> {
     fn types(&mut self, iface: InterfaceId) {
         let iface_data = &self.resolve().interfaces[iface];
 
-        // Second pass: emit full type definitions for all types.
-        //
-        // Here we sort the types topologically, taking into consideration the
-        // parameter and return types of functions associated with resource
-        // types. This ensures each type is declared before any uses of it.
+        // Here we sort the types topologically before emitting code for them,
+        // taking into consideration the parameter and return types of functions
+        // associated with resource types. This ensures each type is declared
+        // before any uses of it.
         for (name, id) in sort_types(self.resolve, &iface_data.types) {
             self.define_type(name, id);
         }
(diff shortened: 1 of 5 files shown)


crates/cpp/src/lib.rs
```diff
@@ -3629,3 +3634,126 @@ fn is_special_method(func: &Function) -> SpecialMethod {
...
+        .iter()
+        .map(|(k, v)| (*v, k.as_str()))
+        .collect::<HashMap<_, _>>();
+    sorted
+        .into_iter()
+        .filter_map(|v| names.get(&v).map(|k| (*k, v)))
+        .collect()
+}
+
+fn sort(resolve: &Resolve, ty: Type, sorted: &mut IndexSet<TypeId>, visited: &mut HashSet<TypeId>) {
```
alexcrichton: Could this use [`TypeIdVisitor`](https://docs.rs/wit-parser/latest/wit_parser/trait.TypeIdVisitor.html) to simplify this?

crates/cpp/src/lib.rs
```diff
@@ -891,9 +892,13 @@ impl CppInterfaceGenerator<'_> {
             }
         }
 
-        // Second pass: emit full type definitions
-        for (name, id) in iface_data.types.iter() {
-            self.define_type(name, *id);
+        // Second pass: emit full type definitions for all types.
```
alexcrichton: With this sort is the first pass above still required?

alexcrichton (approved):

crates/cpp/src/lib.rs
dicej: Apparently not; I'll remove it.

crates/cpp/src/lib.rs
dicej: I'll try that.

alexcrichton (approved):

cpetig:
Thank you, Joel, for fixing this!

PR #1620, opened 2026-05-27
C++: fix compile errors when lowering `result<string>`, etc.

dicej:
Previously, the code generator got confused about when to use `wit::string` vs. `std::string_view` when lowering variants with string payloads, tuple-with-string payloads, etc.

I'll admit my C++ skills are pretty out-of-date, so I'm not 100% confident I've got everything lined up correctly here, but it fixes the scenarios I've run into so far.  Feedback from experts in modern C++ would be very welcome.

### DIFF
--- a/crates/cpp/src/lib.rs
+++ b/crates/cpp/src/lib.rs
@@ -3161,23 +3161,28 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
                 }
 
                 let op0 = &operands[0];
-                let flavor = if matches!(self.variant, AbiVariant::GuestImport) {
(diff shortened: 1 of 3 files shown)


dicej:
Interesting; the _Rust_ codegen is choking on the new test case.  I'll investigate.

alexcrichton:
Feel free to ignore the Rust test failure if it takes too much to investigate, it's got to do with trying to get everything working in all borrowing modes which while useful isn't always the most urgent

dicej:
Okay, I'll add it to the `should_fail_verify` list for Rust and open an issue for it.

alexcrichton (approved):

PR #1625, opened 2026-06-06
Rust: Provide a trait method to (optionally) control resource allocation

cpetig:
This enables resource allocation from a user provided arena instead of the heap.

Arena implementation in test case generated by genAI.


crates/rust/src/interface.rs
```diff
@@ -239,6 +239,30 @@ fn _resource_rep(handle: u32) -> *mut u8
...
+#[doc(hidden)]
+/// Consumes the value from the handle, handle is invalid afterwards.
+/// 
+/// # Safety
+/// 
+/// See Box::from_raw
+unsafe fn _resource_from_raw(handle: *mut Option<Self>) -> Option<Self> where Self: Sized
+{{
+    *unsafe {{ {box_path}::from_raw(handle) }}
+}}
```
alexcrichton: There's a few aspects of this I'd like to bikeshed if that's ok with you as well. One is that I'd rather not open-code the `Option<Self>` implementation if possible. Given how everything's moving to a `trait`-based definition, do you think it'd be possible to have, for example, a default associated type or something like that which represents the container here? Something like `type RawContainer: SomeOtherTrait = Option<Self>;` or something like that, where `Option<T>` by default implements `SomeOtherTrait`. That wouldn't be directly related to the allocation here, and would involve abstracting the preexisting `.as_mut().unwrap()` operations in a few places, but it'd be in the same trend as this.

Orthogonally I'd bikeshed these exact shapes/docs a bit. The `_`-prefixed methods are probably overkill at this point. It might make sense to rename them to maybe a `_`-suffix rather than a `_` prefix (which typically means "yes I know this is unused ignore that" which isn't the case here). The implementations would stay the same, however. The documentation, though, I think will want to be expanded to explain a bit more about the allocation lifecycle and what's expected of the pointer (e.g. it lives for the entire duration of the resource). I think `resource_into_raw` will need to be `unsafe` as well because one of the contractual invariants of `resource_from_raw` will be that it only consumes values previously produced by `resource_into_raw`. Finally, the `where Self: Sized` I think is fine to move to the trait definition (make `Sized` a supertrait) and avoid the syntactic overhead here.

tests/runtime/rust/arena-allocated-resources/test.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        /// Gets mutable access to the arena.
+        ///
+        /// # Safety
+        ///
+        /// This is safe in single-threaded contexts. In multi-threaded contexts,
+        /// the caller must ensure exclusive access.
+        #[inline]
+        pub fn get_mut(&self) -> &mut Arena<T, SIZE> {
```
alexcrichton: FWIW this is fundamentally unsafe and more-or-less UB. This'll want to be protected by some sort of lock or similar.

For this test though I think it's ok to avoid being fancy and just manually call `Box` maybe with a slightly different wrapper just to showcase that it can be done (e.g. wrap it up in `Box<Thing<T>>` instead of `Box<T>` or something like that)

alexcrichton (commented):
I like the idea, thanks! I've got some comments below, but let me know if you'd prefer to not handle anything in particular.

crates/rust/src/interface.rs
cpetig: I agree with hiding the `Option<T>` implementation detail by an associated type or similar (I think there exists an `FooImp` type or similar already for this). I am not sure about creating another trait here (feels overkill to me for now).

I also agree on the `_` suffix and the better documentation and Sized in a different place.

But ..  I consider the argumentation about `into_raw` better be `unsafe` not compelling. There is no invariant on `into_raw`, only `from_raw` has invariants (only called once and only on a result from `into_raw`). Similarly only one of these member functions is unsafe on Box as well.

tests/runtime/rust/arena-allocated-resources/test.rs
cpetig: I agree. Turning `&` into `&mut` is instant UB - unless there is some sort of internal mutability involved (which probably needs an UnsafeCell to protect against compiler optimizations).

I didn't look that closely into this third re-implementation of the arena, my goal was to create a minimalist wasm component without any reference to malloc - which involves avoiding panic and libc, but can be achieved by wat-patching.

I guess I should look a bit more closely into how to make the arena throwaway example code opsem safe and not an evil example buried in a high-quality code base.

cpetig: Such a thing (&self to &mut T) exists ... https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html#method.as_mut_unchecked but I agree that the unsafety should become better capsuled.

crates/rust/src/interface.rs
alexcrichton: That sounds reasonable to me, yeah, and good points!

cpetig: I tried my best, but a lot of the really nice options require unstable associated type features.

tests/runtime/rust/arena-allocated-resources/test.rs
cpetig: Here I simplified the whole arena into an implementation I would be willing to defend the soundness of. I think it should be sound and not cause UB.

crates/rust/src/interface.rs
cpetig: Please note that all the other types and functions in the trait are named `_foo` instead of `foo_`, so it feels a bit incoherent now.

tests/runtime/rust/arena-allocated-resources/test.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+use arena::Arena;
+
+#[derive(Clone)]
+struct MyThing {
+    contents: u32,
+}
+
+static ARENA: Arena<Option<MyThing>, 4> = Arena::new();
```
cpetig, line 72: Oh, this should become ThingStorage instead of Option

alexcrichton:
@cpetig ok I tried my hand a this a bit and resulted in the last commit here. Everything looks good to me now but I'd want to run that by you to confirm.

I realized that ideally what I want is:

```rust
trait GuestThing {
    type Rep: ResourceRep = Box<Option<T>>;
    // ...
}
```

however  defaults for associated bounds aren't stable in Rust. In lieu of that I did a, now unrelated, refactoring to the `Option<T>` "rep" which abstracts operations behind a trait. This is sort of like the `_FooStorage<T>` from prior but that didn't do what I want because it wasn't customizable. It's still not customizable due to the blanket impl, but I think it's a bit closer to being customizable (subjectively).

I also realized that the `resource_into_raw_` method needed to be `unsafe` mostly because otherwise it would be possible to override just `resource_into_raw_`, nothing else, and you'll have written no `unsafe` code but that would be theoretical UB. This is best modeled as an `unsafe trait` but in lieu of that with associated types and such, I'll pass on that.

One possible future option is to basically require all Rust exported resources to manually say `type Rep = Box<Option<Self>>` as a form of boilerplate. That way it could be fully customizable at the cost of being a bit more verbose.

cpetig:
> however defaults for associated bounds aren't stable in Rust. In lieu of that I did a, now unrelated, refactoring to the `Option<T>` "rep" which abstracts operations behind a trait. This is sort of like the `_FooStorage<T>` from prior but that didn't do what I want because it wasn't customizable. It's still not customizable due to the blanket impl, but I think it's a bit closer to being customizable (subjectively).

It amuses me that you took my frustration with the incompleteness of traits in stable Rust to the next level, yesterday I was just sad that defaults for associated types were not possible in this PR, but a bound on an associated type is a concept I didn't miss until today.

> I also realized that the `resource_into_raw_` method needed to be `unsafe` mostly because otherwise it would be possible to override just `resource_into_raw_`, nothing else, and you'll have written no `unsafe` code but that would be theoretical UB. This is best modeled as an `unsafe trait` but in lieu of that with associated types and such, I'll pass on that.

Oh, another facet of unsafe which I didn't see before. An unsafe trait would make sense, because you rely on the correctness of the implementation of two particular functions, but you don't want to make the trait implementation unsafe for all people not overriding the allocation method.

> One possible future option is to basically require all Rust exported resources to manually say `type Rep = Box<Option<Self>>` as a form of boilerplate. That way it could be fully customizable at the cost of being a bit more verbose.

That feels excessive, but also missing specialization has made one of my API designs less elegant than I wished for before. 🤷 

Thank you for teaching me new aspects of Rust, I will need some hours to read and understand your new design.

cpetig:
Looks elegant to me. I think documentation always improves when it written by more than a single individual, you pointed out so many things I have forgotten to mention.

crates/rust/src/interface.rs
```diff
@@ -2825,12 +2844,14 @@ impl {camel} {{
     }}
 
     #[doc(hidden)]
-    pub unsafe fn dtor<T: 'static + Guest{camel}>(handle: *mut u8) {{
+    pub unsafe fn dtor<T: Guest{camel}>(handle: *mut u8) {{
         Self::type_guard::<T>();
-        let _ = unsafe {{ T::resource_from_raw_(handle as *mut _{camel}Rep<T>) }};
+        unsafe {{
+            let _rep = T::resource_from_raw_(handle.cast());
```
cpetig, line 2850: This is the one change which makes me curious, I tried to minimize the number of characters in the `unsafe` block. Why would you put the `let` into the unsafe block instead?

alexcrichton: Ah mostly just subjective on this. It's all generated code anyway and doesn't really get `// SAFETY: ...` comments regardless so I basically just took the expedient route here. Otherwise though I'd agree that this would probably best be served by 2 blocks.

alexcrichton (approved):
Ok I'll go ahead and flag this for merge, thanks for putting up with me here @cpetig and thanks again!

PR #1627, opened 2026-06-07
support "--with" parameter to include already generated types

cpetig:
This implements --with for C in a way similar to C++ (include) and Rust (external type path).


cpetig:
I guess a test would be the ideal next step ...

:thinking: The runner would import (half of) the exports by the test via a library, but this library design means that the test runner would need to run wit-bindgen twice for generating one component, also the test case linking becomes much more complex than now. 

I feel this is beyond the capabilities of the current test runner :thinking:

alexcrichton:
I think this would be best to have a test, yeah, and I agree the current infrastructure isn't sufficient to implement it. Would you be up for extending the test runner with support for this? It might make sense to reimplement the preexisting `--with` tests for Rust in terms of this too?

PR #1629, opened 2026-06-08
Rename `spawn` to `spawn_local`

### LINKED ISSUE #1626: rust: Should `spawn` be renamed `spawn_local`?
`spawn` is documented to be single-threaded, but multi-threaded async execution is ubiquitous enough in the Rust ecosystem that it is easy to miss (or forget, even for people who really ought to know better 😐) the important distinction. Renaming the function could be a nice signal to users that this may not be the `spawn` they are ~looking for~ used to, and the semantics of the operation are significantly closer to what other Rust async executors call `spawn_local`.

alexcrichton:
Rationale is listed in #1626.

Closes #1626

### DIFF
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -881,7 +881,7 @@ pub mod rt;
 #[cfg(feature = "inter-task-wakeup")]
 pub use rt::async_support::UnitStreamOps;
 #[cfg(feature = "async-spawn")]
-pub use rt::async_support::spawn;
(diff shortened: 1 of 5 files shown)


dicej (approved):

itowlson:
Presumably too late now, but it would have been nice to keep `spawn` for a few versions with a deprecation note saying "use `spawn_local` instead."  I know this is an 0.x crate with no semver obligations - but users are having to upgrade to do async or work around bugs, so it would be good to provide some kind of help with continuity.  Something to consider for next time at least - thanks!

Issue #1078, opened 2024-11-06
c#: only use version info if required


jsturtevant:

sano-suguru:
Proposing to port the `pkg_has_multiple_versions` guard from `crates/c/src/lib.rs::interface_identifier` to `interface_name` in `csharp/src/world_generator.rs` — the sole version-emit site in the C# backend. The fix makes C# match C's behavior exactly (version segment only when multiple versions of the same package coexist).

Before opening a PR: this is a silent default change that breaks source compatibility for users who authored code into versioned namespaces. The C backend precedent suggests "default, no flag" — but wanted to confirm whether a flag is preferred here.

yowl:
I think we still have lots of changes to come and we don't need to worry about breaking code yet.  There is certainly no expectation on my side. So I would go for consistency with c.

jsturtevant:
Agreed breaking changes are fine.  Only using versions when needed is how rust does it too and make it more developer friendly i believe

sano-suguru:
Opened #1635 for this — CI is green, including the C# suites on Linux and Windows.

PR #1635, opened 2026-06-13
csharp: omit version namespace segment when package is unambiguous (#1078)

sano-suguru:
Ports the `pkg_has_multiple_versions` guard from the C generator (`crates/c/src/lib.rs`, `interface_identifier`) to the C# backend, closing #1078.

## What changes

Previously the C# generator unconditionally appended a version segment to the namespace of every versioned WIT package:

```
my.dep.v0_1_0.wit.imports.a.Ia   // always, even when only one version exists
```

After this change the segment is only emitted when the same package `namespace:name` appears at more than one version in the Resolve — the only situation where omitting it would cause a collision:

```
my.dep.wit.imports.a.Ia          // single version → no segment
my.dep.v0_1_0.wit.imports.a.Ia   // coexists with v0_2_0 → segment kept
```

The guard logic is identical to the C backend's. The mangling format intentionally differs by design: C# emits `v{major}_{minor}_{patch}.` while C uses a trailing-underscore snake-case segment. The Rust and Go backends already do the same via the shared `name_package_module` helper (`crates/core/src/path.rs`).

## Tests

- **Unit tests** on `interface_name` assert the generated namespace directly: the segment is dropped for a single-version package and kept when two versions coexist. I verified they fail if the guard is reverted to the unconditional form, so they lock the behavior (not just compilation) and run on every platform without the dotnet toolchain.
- **Codegen fixture** `tests/codegen/single-version-package` (one versioned import, single version in the Resolve) builds the drop path across all backends, alongside the existing `multiversion` fixture (two versions of the same package) which guards the must-keep path.

C# codegen build tests don't run on my machine (macOS), so the `dotnet build` step is left to CI; `cargo build` and the unit tests pass locally.

## Design decision

I asked on the issue whether this should be a silent default or gated behind a flag. @yowl preferred the default/no-flag approach for consistency with the C backend (https://github.com/bytecodealliance/wit-bindgen/issues/1078#issuecomment-4687002167). This changes generated C# namespace names for single-version packages — a breaking change that's acceptable pre-1.0 per that discussion.

### DIFF
--- a/crates/csharp/src/world_generator.rs
+++ b/crates/csharp/src/world_generator.rs
@@ -1049,8 +1049,18 @@ fn interface_name(
             );
 
             if let Some(version) = &name.version {
-                let v = version.to_string().replace(['.', '-', '+'], "_");
-                ns = format!("{}v{}.", ns, &v);
+                // Only include the version segment when the same package name exists at
+                // multiple versions in this Resolve; omit it when unambiguous.
+                // Mirrors the equivalent guard in the C generator (crates/c/src/lib.rs).
+                let pkg_has_multiple_versions = resolve.packages.iter().any(|(_, p)| {
+                    p.name.namespace == name.namespace
+                        && p.name.name == name.name
+                        && p.name.version != name.version
+                });
+                if pkg_has_multiple_versions {
+                    let v = version.to_string().replace(['.', '-', '+'], "_");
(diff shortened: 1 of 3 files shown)


yowl (approved):
Looks good, thanks!

PR #1637, opened 2026-06-17
Evolve the wasip3 async C ABI for tasks

### LINKED ISSUE #1618: assertion `left == right` failed at `src/rt/async_support/waitable.rs:201`
I'm hitting [this assertion](https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/guest-rust/src/rt/async_support/waitable.rs#L201).

I ran into this while working on a branch that uses `wit_bindgen::block_on` to drive stream I/O synchronously from within an async export.

With some LLM help I was able to reduce it to: https://github.com/badeend/wasmtime/blob/waitable-panic-repro/crates/test-programs/src/bin/p3_tls_sample_application.rs

Noet that the assertion only fires if the two write futures are allocated at _different_ heap addresses. If the allocator happens to reuse the same slot (`ptr == prev`) then the assert passes by coincidence, possibly causing memory corruption (?)

I have no idea what the correct behavior should be here. On the surface I'd expect this to work. Or, if `block_on` is not intended to be used like this, I don't think comparing memory addresses is the correct line of defense.

---

Stack trace:

```
thread '<unnamed>' (1) panicked at /home/dave/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wit-bindgen-0.57.0/src/rt/async_support/waitable.rs:201:17:
assertion `left == right` failed
  left: 0x104fd0
 right: 0x104e50
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test p3::rustls::p3_tls_sample_application ... FAILED

failures:

---- p3::rustls::p3_tls_sample_application stdout ----
Error: error while executing at wasm backtrace:
    0:  0x1a3c7 - p3_tls_sample_application-57673490efd69ec5.wasm!abort
    1:  0x155ed - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN3std3sys3pal6wasip17helpers14abort_internal17h74119267dbd11beeE
    2:  0x140af - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN3std7process5abort17h49844fb4bec76201E
    3:  0x1412b - p3_tls_sample_application-57673490efd69ec5.wasm!_RNvCsdBezzDwma51_7___rustc12___rust_abort
    4:  0x13ecf - p3_tls_sample_application-57673490efd69ec5.wasm!_RNvCsdBezzDwma51_7___rustc18___rust_start_panic
    5:  0x13ef2 - p3_tls_sample_application-57673490efd69ec5.wasm!_RNvCsdBezzDwma51_7___rustc10rust_panic
    6:  0x163c1 - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN3std9panicking15panic_with_hook17h7a3e5eaee7fad80cE
    7:  0x158b9 - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN3std9panicking13panic_handler28_$u7b$$u7b$closure$u7d$$u7d$17h36465fe3c8c216bdE
    8:  0x1421a - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN3std3sys9backtrace26__rust_end_short_backtrace17hfc6d834ad5c88707E
    9:  0x1420e - p3_tls_sample_application-57673490efd69ec5.wasm!_RNvCsdBezzDwma51_7___rustc17rust_begin_unwind
   10:  0x1ba41 - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN4core9panicking9panic_fmt17h2b55671c70ea4e21E
   11:  0x1d0ce - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN4core9panicking19assert_failed_inner17h413a92926c49224dE
   12:   0x731a - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN4core9panicking13assert_failed17ha9e7c539c84bc32eE
   13:   0xb3a7 - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN11wit_bindgen2rt13async_support8waitable26WaitableOperation$LT$S$GT$14register_waker17h24054f043d98c2e5E
   14:   0xa812 - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN11wit_bindgen2rt13async_support8waitable26WaitableOperation$LT$S$GT$23poll_complete_with_code17h52b4388adf3a6c5fE
   15:   0xa35f - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN11wit_bindgen2rt13async_support8waitable26WaitableOperation$LT$S$GT$13poll_complete17h574be3c29c8fe313E
   16:   0x85a7 - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN120_$LT$wit_bindgen..rt..async_support..stream_support..RawStreamWrite$LT$O$GT$$u20$as$u20$core..future..future..Future$GT$4poll17hdff73afdb24c0bbdE
   17:   0x2123 - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN106_$LT$p3_tls_sample_application..Component$u20$as$u20$test_programs..p3..exports..wasi..cli..run..Guest$GT$3run28_$u7b$$u7b$closure$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$17h458754691bc5ca2bE
   18:   0x1b4f - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN106_$LT$p3_tls_sample_application..Component$u20$as$u20$test_programs..p3..exports..wasi..cli..run..Guest$GT$3run28_$u7b$$u7b$closure$u7d$$u7d$17h6c6098462d6d7272E
   19:   0x349d - p3_tls_sample_application-57673490efd69ec5.wasm!_ZN13test_programs2p37exports4wasi3cli3run16_export_run_cabi28_$u7b$$u7b$closure$u7d$$u7d$17h808d1ee9b878ecc9E

Caused by:
    wasm trap: wasm `unreachable` instruction executed
```

alexcrichton:
This commit is an evolution of the C ABI used to managed task-related infrastructure in WASIp3 with the goal of solving #1618. The basic problem of #1618 is that waitables in Rust aren't guaranteed to be polled within the context of the original task. For example by mixing an `async` Rust export and `block_on` it's possible to "cross the wires" and poll in one context while dropping/completing in another context. This can lead to buggy situations where a waitable is left in a set, not added to an appropriate set, or generally mis-managed.

The solution here is to enhance the current C ABI of task management with clone/drop operations. Notably this enables waitables to retain a strong reference to the task state as opposed to always consulting what the current task in. This fixes a few situations such as:

* When dropping a half-finished waitable it no longer needs to be dropped in the context of the original task. Dropping will unregister the waitable from a task that it was originally registered with.
* When a waitable is moved from one task to another it needs to implicitly de-register with the previous task, and this was not previously done. Now with a retained strong reference it's able to clear out previous state upon re-registering with a new task.

This change requires some finesse as this needs to be ABI-stable to work with previous versions of the `wit-bindgen` crate. The runtime support additionally can't assume that the new ABI bits are available and instead needs to handle the previous ABI as well. Not too too bad, in the end, though.

This additionally did some refactoring of the state associated with async tasks to juggle things around and better represent the raw pointers/`Arc`/etc from before.

Closes #1618

### DIFF
--- a/crates/guest-rust/src/rt/async_support/cabi.rs
+++ b/crates/guest-rust/src/rt/async_support/cabi.rs
@@ -66,6 +66,7 @@ extern_wasm! {
 /// The first version of `wasip3_task` which implies the existence of the
 /// fields `ptr`, `waitable_register`, and `waitable_unregister`.
 pub const WASIP3_TASK_V1: u32 = 1;
+pub const WASIP3_TASK_V2: u32 = 2;
 
 /// Indirect "vtable" used to connect imported functions and exported tasks.
 /// Executors (e.g. exported functions) define and manage this while imports
@@ -104,4 +105,58 @@ pub struct wasip3_task {
     /// Returns the `callback_ptr` passed to `waitable_register` if present, or
     /// `NULL` if it's not present.
     pub waitable_unregister: unsafe extern "C" fn(ptr: *mut c_void, waitable: u32) -> *mut c_void,
+}
+
+unsafe impl Send for wasip3_task {}
+unsafe impl Sync for wasip3_task {}
+
+/// Indirect "vtable" used to connect imported functions and exported tasks.
+/// Executors (e.g. exported functions) define and manage this while imports
+/// use it.
+#[repr(C)]
+pub struct wasip3_task_v2 {
+    /// TODO
+    pub v1: wasip3_task,
+    /// TODO
+    pub vtable: &'static wasip3_task_vtable,
+}
+
+/// Indirect "vtable" used to connect imported functions and exported tasks.
+/// Executors (e.g. exported functions) define and manage this while imports
+/// use it.
+#[repr(C)]
+pub struct wasip3_task_vtable {
+    /// Currently `WASIP3_TASK_V1`. Indicates what fields are present next
+    /// depending on the version here.
+    pub version: u32,
+
+    /// Register a new `waitable` for this exported task.
+    ///
+    /// This exported task will add `waitable` to its `waitable-set`. When it
+    /// becomes ready then `callback` will be invoked with the ready code as
+    /// well as the `callback_ptr` provided.
+    ///
+    /// If `waitable` was previously registered with this task then the
+    /// previous `callback_ptr` is returned. Otherwise `NULL` is returned.
+    ///
+    /// It's the caller's responsibility to ensure that `callback_ptr` is valid
+    /// until `callback` is invoked, `waitable_unregister` is invoked, or
+    /// `waitable_register` is called again to overwrite the value.
+    pub waitable_register: unsafe extern "C" fn(
+        ptr: *mut c_void,
+        waitable: u32,
+        callback: unsafe extern "C" fn(callback_ptr: *mut c_void, code: u32),
+        callback_ptr: *mut c_void,
+    ) -> *mut c_void,
+
+    /// Removes the `waitable` from this task's `waitable-set`.
+    ///
+    /// Returns the `callback_ptr` passed to `waitable_register` if present, or
+    /// `NULL` if it's not present.
+    pub waitable_unregister: unsafe extern "C" fn(ptr: *mut c_void, waitable: u32) -> *mut c_void,
+
+    /// TODO
+    pub clone: unsafe extern "C" fn(ptr: *mut c_void) -> *mut c_void,
+    /// TODO
+    pub drop: unsafe extern "C" fn(ptr: *mut c_void),
 }
--- a/crates/guest-rust/src/rt/async_support/cabi.rs
+++ b/crates/guest-rust/src/rt/async_support/cabi.rs
@@ -135,10 +135,6 @@ pub struct wasip3_task_v2 {
 /// This was introduced in the "v2" ABI and is a member of `wasip3_task_v2`.
 #[repr(C)]
 pub struct wasip3_task_vtable {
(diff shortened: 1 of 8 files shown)


crates/guest-rust/src/rt/async_support/cabi.rs
```diff
@@ -80,6 +104,41 @@ pub struct wasip3_task {
...
+}
+
+/// Function pointer operations that can operate on `wasip3_task::ptr`.
+///
+/// This was introduced in the "v2" ABI and is a member of `wasip3_task_v2`.
+#[repr(C)]
+pub struct wasip3_task_vtable {
+    /// Currently `WASIP3_TASK_V2` as that was the first version that specified
+    /// vtables.
+    pub version: u32,
```
dicej: Why do we need a `version` field here and also in `wasip3_task`?

dicej (approved):

crates/guest-rust/src/rt/async_support/cabi.rs
alexcrichton: My thinking was that this vtable would grow over time independent of the original `wasip3_task` but that might be a bit too overeager thinking on my part. Does that sound reasonable?

dicej: I don't have strong feelings, but I would imagine that the one in `wasip3_task` could cover everything, including the vtable.

badeend:
Awesome! thanks

PR #1640, opened 2026-06-30
test(csharp): cover map runtime exports

yordis:
- Prevents C# Map support from relying on generated bindings alone while the rest of the SDK matrix exercises Map through runtime composition.
- Catches ABI regressions across nested maps, optional values, variants, results, tuples, and byte-list map values.
- Reduces the risk of C# silently diverging from the shared Map ABI as component-model Map support evolves.

### DIFF
--- a/crates/csharp/src/csproj.rs
+++ b/crates/csharp/src/csproj.rs
@@ -12,6 +12,7 @@ pub struct CSProjectLLVMBuilder {
     clean_targets: bool,
     world_name: String,
     binary: bool,
+    skip_wit_component: bool,
 }
 
 pub struct CSProjectMonoBuilder {
@@ -31,6 +32,7 @@ impl CSProject {
             clean_targets: false,
             world_name: world_name.to_string(),
             binary: false,
+            skip_wit_component: false,
         }
     }
 
@@ -69,6 +71,14 @@ impl CSProjectLLVMBuilder {
             "<OutputType>Library</OutputType>"
         };
 
+        let component_type_linker_arg = if self.skip_wit_component {
+            "<CustomLinkerArg Include=\"-Wl,--skip-wit-component\" />".to_string()
+        } else {
+            format!(
+                "<CustomLinkerArg Include=\"-Wl,--component-type,{camel}_component_type.wit\" />"
+            )
+        };
+
         let mut csproj = format!(
             "<Project Sdk=\"Microsoft.NET.Sdk\">
 
@@ -94,7 +104,7 @@ impl CSProjectLLVMBuilder {
         </ItemGroup>
 
         <ItemGroup>
-            <CustomLinkerArg Include=\"-Wl,--component-type,{camel}_component_type.wit\" />
+            {component_type_linker_arg}
         </ItemGroup>
         "
         );
@@ -176,6 +186,10 @@ impl CSProjectLLVMBuilder {
 
     pub fn binary(&mut self) {
         self.binary = true;
+    }
+
+    pub fn skip_wit_component(&mut self) {
+        self.skip_wit_component = true;
     }
 
     pub fn clean(&mut self) -> &mut Self {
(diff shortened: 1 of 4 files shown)


yordis:
Context from debugging this PR:

- C# generated Map bindings compile correctly; `map<K, V>` lowers to `Dictionary<K, V>` as expected.
- The failure was in the NativeAOT component metadata path, not in the generated C# Map code.
- The default NativeAOT linker path tries to consume `*_component_type.wit` through `-Wl,--component-type,...`, but that parser does not currently handle WIT `map<...>` syntax.
- The working path in this PR is to let NativeAOT compile the core Wasm, skip its own component metadata step for this fixture, and let the wit-bindgen test harness append Map-aware metadata with `wit-component`.

I’m planning to follow up separately in `dotnet/runtime` to investigate the NativeAOT / ILCompiler LLVM component-type parser so `-Wl,--component-type,<world>_component_type.wit` can handle WIT `map<K, V>` directly. That would remove the need for this harness workaround long term.

yordis:
Follow-up notes / order of operations for the native `-Wl,--component-type` Map path:

- This PR proves C# Map runtime behavior in wit-bindgen by avoiding NativeAOT's current text-WIT componentization path for this fixture and letting the test harness append Map-aware metadata with `wit-component`.
- The native `-Wl,--component-type,<world>_component_type.wit` path still depends on the WASI SDK linker toolchain understanding `map<K, V>` directly.
- The relevant source chain is:
  1. `bytecodealliance/wasm-tools` / `wit-parser` / `wit-component` need Map support.
  2. `bytecodealliance/wasm-component-ld` must depend on that Map-capable wasm-tools stack.
  3. `WebAssembly/wasi-sdk` must package that newer `wasm-component-ld`.
  4. `dotnet/runtimelab` must bump its WASI SDK pin for `Microsoft.DotNet.ILCompiler.LLVM`.
- Current investigation found `dotnet/runtimelab`'s `feature/NativeAOT-LLVM` branch pins WASI SDK 29, and WASI SDK 29 packages `wasm-component-ld` 0.5.19. The local `wasm-component-ld` checkout is already at 0.5.25 with `wit-parser` / `wit-component` 0.252.0, which is the direction needed for native Map parsing.

My next follow-up outside this PR is to validate the newest `wasm-component-ld` against a minimal `map<K, V>` component-type WIT, then work forward through the WASI SDK / runtimelab pin chain if that succeeds.

yordis:
Follow-up correction on the external toolchain thread:

- I opened `bytecodealliance/wasm-component-ld` PR #107 as an exploratory regression test for `--component-type` with WIT `map<K, V>`: https://github.com/bytecodealliance/wasm-component-ld/pull/107
- After talking to @alexcrichton, we decided to close that PR for now: https://github.com/bytecodealliance/wasm-component-ld/pull/107#issuecomment-4922050498
- So the wasm-component-ld PR should not be treated as an active follow-up path from this PR.
- The useful takeaway remains that newer `wasm-component-ld` / `wasm-tools` already has the Map-capable parser stack; the longer-term NativeAOT `-Wl,--component-type` path is still a packaging/toolchain adoption question across WASI SDK and runtimelab, not something this wit-bindgen PR needs to solve.

For this PR, the intended scope remains the working wit-bindgen-side bridge: NativeAOT compiles the C# Wasm, and the test harness appends Map-aware component metadata with `wit-component`.

yordis:
Additional follow-up status on the WASI SDK side:

- WASI SDK `main` already pins `wasm-component-ld` 0.5.26.
- The latest release candidate I checked, `wasi-sdk-34-rc.1`, still pins `wasm-component-ld` 0.5.24.
- Older releases lag further behind: `wasi-sdk-33` uses 0.5.22, `wasi-sdk-30` uses 0.5.20, and `wasi-sdk-29` uses 0.5.19.

So the next external step does not look like a WASI SDK code PR right now; `main` already has the newer linker. The remaining question is release/consumption: either wait for a WASI SDK release that includes the 0.5.26 bump, or decide whether `dotnet/runtimelab` should consume a specific unreleased WASI SDK commit/build.

PR #1649, opened 2026-07-06
Fix exception safety violation in AbiBuffer::advance

### LINKED ISSUE #1646: `AbiBuffer::advance` is not exception safe, leading to use after free
Hello, we are security researchers targeting Rust security. By running our tool in this repository, we found a soundness issue. **This report is written by 100% human**. We promise that all you read will never be generated by LLM. Moreover, we are responsible, so if you confirm these issues do exist, we are happy to file PRs to fix it.

The problem happens here:

https://github.com/bytecodealliance/wit-bindgen/blob/f1e55577645c5e0caad85e9e31ff069272359171/crates/guest-rust/src/rt/async_support/abi_buffer.rs#L122-L145

In line 140, `self.ops.dealloc_lists` is called to free the pointers, and in line 144, `self.cursor` is updated accordingly to avoid use of those freed pointers. However, `self.ops.dealloc_lists` will call a user-provided callback function `dealloc_lists` registered via `StreamVtable` (line 93):

https://github.com/bytecodealliance/wit-bindgen/blob/f1e55577645c5e0caad85e9e31ff069272359171/crates/guest-rust/src/rt/async_support/stream_support.rs#L73-L113

So if user deliberately designs a `dealloc_lists`, which will panic after freeing certain lists, then `self.cursor` will not be updated, while some of the lists have been freed. If user recovers from the panic with `std::panic::catch_unwind`, then `self.cursor` can be used to access freed memory.

To fix it, I think we can move the `self.cursor` updating before the for-loop, or update it repeatedly in the for-loop.

Evian-Zhang:
Fix #1646

### DIFF
--- a/crates/guest-rust/src/rt/async_support/abi_buffer.rs
+++ b/crates/guest-rust/src/rt/async_support/abi_buffer.rs
@@ -133,7 +133,11 @@ impl<O: StreamOps> AbiBuffer<O> {
         let (mut ptr, len) = self.abi_ptr_and_len();
         assert!(amt <= len);
         for _ in 0..amt {
-            // Update self.cursor incrementally for exception safety
+            // Update self.cursor incrementally for exception safety.
+            // When `self.ops.dealloc_lists` panics (which is a user-provided
+            // callback), we can make sure any item before (including the
+            // panic one) will not be dealloced again, and the remaining items
+            // can still get advanced properly.
             self.cursor += 1;
             // SAFETY: we're managing the pointer passed to `dealloc_lists` and
             // it was initialized with a `lower`, and then the pointer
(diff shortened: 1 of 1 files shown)


crates/guest-rust/src/rt/async_support/abi_buffer.rs
```diff
@@ -133,6 +133,8 @@ impl<O: StreamOps> AbiBuffer<O> {
         let (mut ptr, len) = self.abi_ptr_and_len();
         assert!(amt <= len);
         for _ in 0..amt {
+            // Update self.cursor incrementally for exception safety
```
alexcrichton: Could this expand on what it means by "exception safety"? For example something along the lines of handling the case where `dealloc_lists` panics we've at least flagged the rest of the buffer as not being advanced but this particular element won't be handled again.

Evian-Zhang: Added :)

alexcrichton (approved):

PR #1651, opened 2026-07-06
Move all async tests to general folder

alexcrichton:
WASIp3 and component-model-async have shipped and are stable, so these are no longer expected to fail.


.github/workflows/main.yml
```diff
@@ -130,17 +141,17 @@ jobs:
...
+    # this language only implements either the runner or test component
     - run: |
-        cargo run test --languages rust,${{ matrix.lang }} tests/runtime \
+        cargo run test --languages rust,c,${{ matrix.lang }} tests/runtime \
           --artifacts target/artifacts \
           --rust-wit-bindgen-path ./crates/guest-rust
 
   # While async is off-by-default and toolchains are percolating this is a
   # separate job to get configured slightly differently.
   async:
```
dicej: Should probably rename this job and the comment above it since it's now about threading rather than async.

dicej (approved):

Issue #1467, opened 2025-12-15
[Rust] custom derive for struct/enum with resource types


chenyan2002:
Currently custom derive applies to all struct and enum definitions. Understandably, the resource struct doesn't use the custom derive. The problem is that when the struct/enum type contains a resource type, the derive trait will not work. We can use `additional-derive-ignore` to opt-out such types individually, but that's not very convenient.

I can see several ways to fix it, but not sure which one makes more sense:
1) Only apply custom derive to struct and enum that doesn't contain resource types. This is equivalent to manually opt-out those types with `additional-derive-ignore`.
2) Apply the custom derive to resource struct as well. But many traits don't make sense for resource types.
3) Have a flag to define custom derive for individual types, similar to `with`, we can have a `with-derive`.

Maybe in the short term, we can apply the first option, which is also easy to implement?

alexcrichton:
There's some additional discussion related to derive at https://github.com/bytecodealliance/wit-bindgen/issues/1386 and https://github.com/bytecodealliance/wit-bindgen/issues/812

Personally I'd prefer to avoid (1) here as it has a high risk of creating confusing situations where an end users enables derives but is confused when they're not present, not understanding the clause where it's' omitted for resource-containing-things. I understand though how other solutions are not easy and I don't have a silver bullet myself.

Issue #1386, opened 2025-09-11
Rust: Support path-specific `additional_derives`


tomasol:
Currently `additional_derives` can only be set globally, which feels too broad.
For use cases like adding `serde` annotations it would be better if the derives could be scoped per type or interface. Would it make sense to extend the `generate!` macro to accept a map of `path -> derives`?

dman-os:
Not just derives but heavily in need of type targeted attrs. Use-cases like serde customization are critical in many cases. Also, would be great if it was hierarchical so that it can target whole packages/modules/types with ease.

For the basic need of segregating derives, I'm currently using multiple `generate!` sections:

```rust
mod wit {
   pub mod serde {
        wit_bindgen::generate!({
            //  #1395 unfortunately, the linker doesn't like `generate!` on the same world twice
            world: "serde-world",
            generate_all,
            additional_derives: [serde::Serialize, serde::Deserialize],
        });
    }
    wit_bindgen::generate!({
        world: "true-world",
        with: {
            "path:to/world": crate::wit::serde::path::to::world,
            // wasi io doesn't like serde
            "wasi:io/poll@0.2.6": generate,
            "wasi:logging/logging@0.1.0-draft": wasi::wit,
        }
    });
}
```

With how [popular](https://lib.rs/development-tools/procedural-macro-helpers) proc-macros are in the rust ecosystem, I suspect I'll always wanting more customization in `generate!`. Though maybe there's some elegant solution lying in wait.

Right now, the solution I've settled on is to write a simple code generator that spits out wit types, corresponding rust types with customization and their binding `generate!` from a single source of truth.

wilfreddenton:
#1656  does this. additional_type_attributes / additional_member_attributes take per-selector attributes keyed by the same names `with` uses (bare, package, interface, or fully-qualified), so you can scope a derive to a package, an interface, or a single type instead of the whole world. Curious whether the selector grammar covers what you had in mind.

dman-os:
@wilfreddenton that looks very good, thanks.

This is not a response to your PR but the general generate! macro itself but, I wonder if a more Rust like approach would be more appropriate. Macro guidelines suggest using a rusty syntax in macros where possible. Think something like

```rust
generate! {
    world: "xxx",
    {
         #[from = "path:to:world", generate_all]
         mod path::to::world {
              #[from = "record-name"]
              #[derive(Serde,Serialize)
               struct RecordName {
                   #[attrib_specifying_lift_lowering]
[...]

Issue #812, opened 2024-01-10
Unable to use `serde` macros in `additional_derives` when importing `wasi::http` types


landonxjames:
I have a simple library testing out the `wasi:http` types. When I try to add an `additional_derives: [serde::Serialize, serde::Deserialize]` to the `wit_bindgen::generate!` macro definition I get the below error. Not sure if thise repo or the `wasi-http` repo would be better for this issue, but trying here first. 

Reproduction can be found in [this repo](https://github.com/landonxjames/wasi-http-test). Uncommenting [this line](https://github.com/landonxjames/wasi-http-test/blob/5c1c393e2f432dc3c6ca4effbd36879ca2924488/runtime/src/lib.rs#L10) will cause the error.

```sh
error[E0277]: the trait bound `wasi::io::error::Error: Deserialize<'_>` is not satisfied
    --> src/lib.rs:5:1
     |
5    | / wit_bindgen::generate!({
6    | | world: "wasi-http-tests",
7    | | exports: {
8    | |     "wasi-http-tests:component/client/wasi-http-tests-client": WasiHttpTestsClient,
9    | | },
10   | | additional_derives: [serde::Serialize, serde::Deserialize]
11   | | });
     | |__^ the trait `Deserialize<'_>` is not implemented for `wasi::io::error::Error`
     |
     = help: the following other types implement trait `Deserialize<'de>`:
               bool
               char
               isize
               i8
               i16
               i32
               i64
               i128
             and 137 others
note: required by a bound in `newtype_variant`
    --> /Users/lnj/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.193/src/de/mod.rs:2119:12
     |
2117 |     fn newtype_variant<T>(self) -> Result<T, Self::Error>
     |        --------------- required by a bound in this associated function
2118 |     where
2119 |         T: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `VariantAccess::newtype_variant`
     = note: this error originates in the macro `wit_bindgen::generate` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `wasi::io::error::Error: Serialize` is not satisfied
   --> src/lib.rs:5:1
    |
5   | / wit_bindgen::generate!({
6   | | world: "wasi-http-tests",
7   | | exports: {
8   | |     "wasi-http-tests:component/client/wasi-http-tests-client": WasiHttpTestsClient,
[...]

PR #1656, opened 2026-07-15
Targeted Attribute Injection

wilfreddenton:
Targeted version of `additional_derives`. Instead of hitting every generated type,
you attach attributes to specific types and members. This is an attempt at the path-specific /
`with`-style derive idea from #1386, #1467, and #812 (also #291, #554).

Two options: `additional_type_attributes` for records/variants/enums, and
`additional_member_attributes` for fields and enum/variant cases. Each maps a selector
to a list of attributes, and the selectors are the same keys `with` uses (bare name,
package, interface, or fully-qualified, with `@version` when versioned):

    wit_bindgen::generate!({
        // ...
        additional_type_attributes: {
            "my:pkg/types/record-name": [r#"#[derive(serde::Serialize)]"#],
            "my:pkg/types": [r#"#[derive(Clone)]"#],          // every type in the interface
        },
        additional_member_attributes: {
            "record-name.field": [r#"#[serde(rename = "f")]"#],
        },
    });

Any attribute works, not just derives. An injected `#[derive(...)]` folds into the
type's generated derive so it dedups against the built-ins and `additional_derives`
instead of colliding.

Went with two options rather than one nested map to match prost's
`type_attribute`/`field_attribute`, keep the CLI flags simple, and avoid a bare name
being ambiguous between a type and a member. A selector that matches nothing is an
error, same as `with` with unused keys.

Some limits: only records/variants/enums (broad selectors just skip
flags/resources/aliases), attributes land on every ownership form so an owned-only
derive like `Deserialize` won't compile on a borrowed form, and it's Rust-only for
now following #678/#1199.

Not sure on the naming. `additional_*` sits next to `additional_derives`, but it acts
more like `with`. Happy to switch to `type_attributes`/`member_attributes` or whatever
you prefer.

Tests in `crates/rust/tests/codegen.rs`.

### DIFF
--- a/crates/rust/src/interface.rs
+++ b/crates/rust/src/interface.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 use anyhow::Result;
 use heck::*;
+use indexmap::IndexSet;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Write as _;
 use std::mem;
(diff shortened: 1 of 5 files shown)


alexcrichton:
I know I'm taking a bit of time to get to this, sorry about that. This hasn't fallen off my radar, just haven't gotten the time to review it yet

crates/rust/src/interface.rs
```diff
@@ -2086,6 +2087,119 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
...
+        let id = dealias(self.resolve, id);
+        match self.resolve.types[id].name.as_deref() {
+            Some(name) => vec![name.to_string(), full_wit_type_name(self.resolve, id)],
+            None => Vec::new(),
+        }
+    }
+
+    /// Every selector key that matches type `id`: its identity keys plus its
+    /// owning interface and package.
+    fn type_selector_keys(&self, id: TypeId) -> Vec<String> {
```
alexcrichton: I'd like to ideally have a relatively central method of specifying strings which identify/map to WIT types. To that extent this is at least somewhat present as `full_wit_type_name` at the root of the crate and how it affects `with`. Can that be integrated and/or updated to handle this new use case as well? I woudl prefer to not add another method of identifying types/etc with strings.

crates/rust/src/interface.rs
```diff
@@ -2086,6 +2087,119 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
...
+            });
+        self.r#gen.used_member_attr_selectors.extend(used);
+        attrs
+    }
+
+    /// Split injected attributes into derive paths and everything else. Derive
+    /// paths are merged into the generated `#[derive(...)]` so they dedup against
+    /// the built-in and `additional_derives` derives instead of colliding (E0119);
+    /// other attributes are emitted verbatim.
+    fn split_injected_derives(attrs: &[String]) -> (Vec<String>, Vec<String>) {
```
alexcrichton: I think it should be ok to drop this, it's ok to have two `#[derive]` attributes

crates/rust/src/interface.rs
```diff
@@ -2182,10 +2307,12 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
     {
         self.print_rust_enum(
             id,
+            // Raw wit case names; `print_rust_enum` upper-camel-cases them at the
+            // emit site so member selectors still match the wit (kebab) name.
```
alexcrichton: No need for historical docs here, these docs can be dropped.

crates/rust/src/lib.rs
```diff
@@ -234,6 +251,42 @@ pub struct Opts {
...
+    /// type, including the borrowed (`<'a>`) form produced under `ownership:
+    /// Borrowing`; there is no owned-only targeting, so an owned-only derive such as
+    /// `#[derive(Default)]` or `serde::Deserialize` fails to compile on the borrowed
+    /// form (as it would via `additional_derive_attributes`). Only records, variants,
+    /// and enums are covered; flags, resources, type aliases, and the built-in
+    /// option/result/list types are not. A package or interface selector silently
+    /// skips them (they do not count toward the selector being used), but a selector
+    /// that resolves to nothing covered is an error. For a derive on *all* types use
+    /// `additional_derive_attributes`. In a CLI, this flag can be specified multiple
+    /// times as `selector=attribute`.
```
alexcrichton: Could this be formatted a bit? right now it's just a giant wall of text describing the option

crates/rust/tests/codegen.rs
```diff
@@ -301,3 +301,170 @@ mod merge_structurally_equal_types {
...
+        ownership: Borrowing { duplicate_if_necessary: true },
+        // wit-bindgen already derives Clone/Debug on records+variants and every
+        // standard trait but Hash on enums; injected derives fold in (see `label`).
+        additional_type_attributes: {
+            // cumulative list; the trailing `Hash` also comes from the package
+            // selector below and folds into one derive rather than colliding.
+            "point": [
+                "#[derive(PartialEq, Eq)]",
+                "#[derive(PartialOrd, Ord)]",
+                "#[derive(Hash)]",
```
alexcrichton: With Rust syntax here, could the string literals be removed in favor of just the attributes themselves?

crates/rust/src/lib.rs
```diff
@@ -2007,3 +2085,28 @@ fn classify_constructor_return_type(
...
+                "point.x".to_string(),
+                r#"#[serde(rename = "x")]"#.to_string()
+            )
+        );
+        assert!(parse_attribute("no-equals-sign").is_err());
+        // an empty attribute is rejected rather than silently escaping the check
+        assert!(parse_attribute("point=").is_err());
+        assert!(parse_attribute("point=   ").is_err());
+    }
+}
```
alexcrichton: It's ok to drop tests like this, no need to test every single piece of functionality added (that's just too verbose and brittle). I know LLMs like to do this but it's not necessary.

alexcrichton (commented):
Thanks for the PR again, and apologies for the delay. This all seems reasonable to me but I'd primarily like to integrate the string-selector style syntax with the rest of the macro (as opposed to building something brand new). Could it also be integrated with the unused-attributes warning/error that wit-bindgen emits to catch typos?

crates/rust/src/interface.rs
wilfreddenton: Changed this to a thin wit_type_selectors wrapper around full_wit_type_name, which it calls rather than reimplements, so with is untouched. Pushing the coarser forms into with doesn't seem right though: a remapping is 1:1, so a package key would have to map every type in the package to the same Rust path, whereas attributes fan out by design.

wilfreddenton:
@alexcrichton for the unused-attributes warning I believe I've added that here: https://github.com/bytecodealliance/wit-bindgen/pull/1656/changes#diff-e256bf47e7fd893d97fe738d2d6b4a47ed74ad649bc94a403398a4ea509ca38bR1620

but I can fold it into a single diagnostic here: https://github.com/bytecodealliance/wit-bindgen/pull/1656/changes#diff-e256bf47e7fd893d97fe738d2d6b4a47ed74ad649bc94a403398a4ea509ca38bR1600

crates/rust/src/lib.rs
```diff
@@ -1948,6 +2037,24 @@ fn full_wit_type_name(resolve: &Resolve, id: TypeId) -> String {
...
+    names.extend(type_def.name.clone());
+    if let TypeOwner::Interface(iface) = type_def.owner {
+        // `Resolve::id_of` unwraps the package, so only ask once there is one.
+        if let Some(pkg) = resolve.interfaces[iface].package {
+            names.extend(resolve.id_of(iface));
+            names.push(resolve.packages[pkg].name.to_string());
+        }
+    }
+    names
+}
```
alexcrichton: I personally think that the set here is to coarse given the matching involved. This I feel is too easy to have clashes between package names, type names, and same-named-types across interfaces. I realize that this is different than the design you're writing up here, but would it be ok to require that selecting a type requires the full syntax? Something like `a:b/c.d` or something like that

crates/rust/src/interface.rs
```diff
@@ -2086,6 +2087,51 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
...
+    /// last one separates the two halves.
+    fn additional_member_attrs(&mut self, selectors: &[String], member: &str) -> Vec<String> {
+        let (attrs, used) =
+            Self::matching_attrs(&self.r#gen.opts.additional_member_attributes, |sel| {
+                sel == member
+                    || sel.rsplit_once('.').is_some_and(|(ty, m)| {
+                        m == member && selectors.iter().any(|name| name == ty)
+                    })
+            });
+        self.r#gen.used_member_attr_selectors.extend(used);
```
alexcrichton: Instead of returning two sets here, could a mutable references to the `used_*` set be passed into `matching_attrs`?

crates/rust/src/lib.rs
wilfreddenton: Ok wit_type_selectors has been fully removed and full syntax is required for both types and members: `my:pkg/types/point` and `my:pkg/types/point.x`.

alexcrichton (approved):

PR #1657, opened 2026-07-17
Add support for named implements in Go

### LINKED ISSUE #1642: Go: resource failed to be generated for implements feature
# Description

For the given wit with 2 keyvalue stores:
```wit
world handler {
  import primary: wasi:keyvalue/store@0.2.0-draft;
  import secondary: wasi:keyvalue/store@0.2.0-draft;
}
```

Build fails due to the `Bucket` and `Error` types not being generated:
```sh
$ componentize-go \
          --wit-path ./wit \
          --world handler \
          --world sdk \
          build
Error: 'go build' command failed: # implements-test/gen/secondary
gen/secondary/wit_bindings.go:19:13: undefined: Bucket
gen/secondary/wit_bindings.go:19:79: undefined: Error
gen/secondary/wit_bindings.go:76:13: undefined: Bucket
gen/secondary/wit_bindings.go:76:83: undefined: Error
gen/secondary/wit_bindings.go:123:13: undefined: Bucket
gen/secondary/wit_bindings.go:123:71: undefined: Error
gen/secondary/wit_bindings.go:168:13: undefined: Bucket
gen/secondary/wit_bindings.go:168:62: undefined: Error
gen/secondary/wit_bindings.go:213:13: undefined: Bucket
gen/secondary/wit_bindings.go:213:78: undefined: KeyResponse
gen/secondary/wit_bindings.go:213:78: too many errors
```

These types are generated in the primary interface `gen/primary/wit_bindings.go` but not in the secondary:

Image

Source code to reproduce can be found below:

https://github.com/jamesstocktonj1/implements-test

jfleitz:
Fixes #1642 

## Overview

In the scenario where you are importing the same interface under multiple names (implements feature in wasm), the Go generator only emitted the interface types from the first package, since there is a check for not adding a duplicate interface. 

This change adds the Go package name to the Hash key instead of just `TypeId`, allowing for all interfaces to be exported in all packages. There is no concern on duplicate interfaces in the same package, since that would be a Go compile error.

C++ bin gen was also changed to support the implements feature as well, with using the derived namespace in the WorldKey (`primary::Bucket`, `secondary::Bucket`).

## Testing

- Added `tests/codegen/issue1642.wit` importing the same interface with two names (to test implements) as well as non-named (to test prior functionality)
- Verified with local `wit-bindgen go` and re-generated the bindings for the [implements-test](https://github.com/jamesstocktonj1/implements-test) project.

Signed-off-by: Jeremy Fleitz <jeremy@cosmonic.com>

### DIFF
--- a/crates/go/src/lib.rs
+++ b/crates/go/src/lib.rs
@@ -739,9 +739,7 @@ impl WorldGenerator for Go {
         id: InterfaceId,
         _files: &mut Files,
     ) -> Result<()> {
-        if let WorldKey::Name(_) = name {
-            self.interface_names.insert(id, name.clone());
-        }
+        self.interface_names.insert(id, name.clone());
 
         let package = self.interface_name(resolve, Some(name));
         let mut data = {
@@ -786,9 +784,7 @@ impl WorldGenerator for Go {
         id: InterfaceId,
         _files: &mut Files,
     ) -> Result<()> {
-        if let WorldKey::Name(_) = name {
-            self.interface_names.insert(id, name.clone());
-        }
+        self.interface_names.insert(id, name.clone());
 
         let package = self.interface_name(resolve, Some(name));
         for (type_name, ty) in &resolve.interfaces[id].types {
--- a/crates/go/src/lib.rs
+++ b/crates/go/src/lib.rs
@@ -252,7 +252,7 @@ impl Go {
         if local == owner && (exported ^ in_import) {
             String::new()
         } else {
-            let package = self.interface_name(resolve, owner);
+            let package = self.go_package_name(resolve, owner);
             let package = if exported {
                 format!("export_{package}")
             } else {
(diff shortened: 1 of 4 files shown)


crates/go/src/lib.rs
```diff
@@ -738,15 +739,13 @@ impl WorldGenerator for Go {
         id: InterfaceId,
         _files: &mut Files,
     ) -> Result<()> {
-        if let WorldKey::Name(_) = name {
-            self.interface_names.insert(id, name.clone());
-        }
+        self.interface_names.insert(id, name.clone());
 
+        let package = self.interface_name(resolve, Some(name));
```
dicej: Maybe rename this to `go_package_name` just to avoid confusion with the WIT meaning of `package`?

crates/go/src/lib.rs
```diff
@@ -788,30 +784,27 @@ impl WorldGenerator for Go {
...
```
dicej: As above, renaming this to `go_package_name` would help avoid confusion.  We could also consider renaming the `interface_name` function to `go_package_name` for consistency.

dicej (approved):
Thanks for doing this, @jfleitz!  LGTM; just a couple of suggestions inline.

crates/go/src/lib.rs
jfleitz: Good call @dicej ! Just made the change.

PR #1667, opened 2026-07-26
test(cpp): cover map runtime roundtrips

yordis:
- Restores the C++ map runtime coverage deferred in #1590 and adds the import direction, which never existed, so C++ Map support is exercised through runtime composition like the rest of the matrix instead of relying on codegen tests alone.
- Catches ABI regressions across nested maps, optional values, variants, results, tuples, and byte-list map values, in every pairing with the C and Rust runners/tests.
- Fixes a latent use-after-free in the C++ generator that the new runner exposed: `ResultLower` moved the payload into a block-scoped local, freeing its heap data before the lowered import call consumed the flat buffer. Payloads are now bound by rvalue reference, matching how `VariantLower` binds case payloads. This affected any heap-carrying `result` passed as an argument into an import, not just maps; no existing C++ fixture covered that path.

### How the toolchain blocker is avoided

#1590 dropped its runtime test because the `wasm-component-ld` bundled with wasi-sdk doesn't understand the component-model `map` encoding. The C++ test harness now links a core module with `-Wl,--skip-wit-component` and componentizes with the workspace's `wit-component`, mirroring what the C harness already does for every C test. The wasi-sdk linker never parses the map metadata, so no external toolchain update is needed.

### Testing

With wasi-sdk 30 and wasmtime 46.0.1 (the CI versions):

- `tests/runtime` with `--languages rust,c,cpp`: 269 component builds, 236 runtime tests, all passing, including the full 3x3 map runner/test matrix.
- `tests/codegen` with `--languages cpp`: 101 tests passing, validating the `ResultLower` change across the codegen fixtures.

### DIFF
--- a/crates/cpp/src/lib.rs
+++ b/crates/cpp/src/lib.rs
@@ -3300,13 +3300,17 @@ impl<'a, 'b> Bindgen for FunctionBindgen<'a, 'b> {
                     &self.namespace,
                     Flavor::InStruct,
                 );
+                // Bind payloads by reference: the payload's heap data must
+                // stay owned by `op0` (which outlives the lowered call). A
+                // by-value binding would free that data at the end of the
+                // `if`/`else` block below, before the flat ABI consumes it.
                 let bind_ok = if let Some(_ok) = result.ok.as_ref() {
-                    format!("{ok_ty} {ok_payload} = std::move({op0}).value();")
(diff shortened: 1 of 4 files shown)


yordis:
@cpetig any opportunity to get some code review on this one?

PR #1677, opened 2026-08-05
fix(rust): restore `cabi_realloc` for `wasm32-unknown-unknown`

scothis:
`cabi_realloc` was switched to only be added for `p1` targets instead of not being added for `p2` targets in #1663. This causes modules built with `wasm32-unknown-unknown` and then enriched into a component with `wasm-tools component new` to fail.


scothis:
I plan to switch to `wasm32-component` once it becomes available, but if there is a better way to create a component that doesn't need wasi, please let me know.

alexcrichton (approved):

PR #1685, opened 2026-08-17
feat(rust) add native link option

BjornTheProgrammer:
Addresses [bytecodealliance/wit-bindgen#1062](https://github.com/bytecodealliance/wit-bindgen/issues/1062). Currently the Rust bindings stub with `unreachable!()` when compiling to wasm. This PR makes bindings usable on native targets, so people can test and run component code in ordinary native environments.

This is a new approach to what was initially done in https://github.com/bytecodealliance/wit-bindgen/pull/1565.

The reason this PR needs to exist is described by a comment in the original PR

> This PR despite making the possible stubs very modular, suffers from a lot of issues, I'm going to close this in favor of another PR that I've been making. The idea is a bit different, and slightly more complex then just calling a host function, since that would require the host binary to be compiled with "-rdynamic" on linux, and it wouldn't even work on Windows and MacOS.

To solve this issue I've added a register function which can add a pointer to the host function. This fixes the above issue.

## Testing

- Codegen tests have been added.
- End-to-end validation via [native-wit](https://github.com/BjornTheProgrammer/native-wit) hosting native plugins in [Pumpkin](https://github.com/Pumpkin-MC/Pumpkin/pull/2828).

The Pumpkin PR is interesting because the approach shows how one can make very minimal changes and get wasmtime and native runtimes working at the same time.

### DIFF
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -891,26 +891,41 @@ extern crate std;
 ///     // structurally equal, which is useful when import and export the same
 ///     // interface.
 ///     merge_structurally_equal_types: true,
-///
-///     // Make the same generated bindings usable on a native (non-wasm)
-///     // target as well as on wasm32.
-///     //
-///     // Imports normally compile to `unreachable!()` off wasm32. With this
-///     // enabled each one instead calls through a function pointer that a host
-///     // installs at load time via a generated
-///     // `__wit_bindgen_register_*` symbol, and exports additionally get a
-///     // native symbol whose name encodes the characters a linker cannot
-///     // accept. Both targets still build from one source.
-///     //
-///     // The registration symbols are prefixed with a hex-encoded
-///     // `<package>/<world>` so that two `generate!` invocations in one crate
-///     // don't collide. Binding the *same* world twice in one linkage unit
-///     // still does; use `type_section_suffix` to tell them apart. See
-///     // `wit_bindgen_rust::Opts::link_native_symbols` for the full list of
-///     // symbols a host can expect.
-///     link_native_symbols: true,
 /// });
 /// ```
+///
+/// ## Native (non-WebAssembly) targets
+///
+/// Generated bindings also compile for native targets, which is useful for
+/// testing component code without a wasm runtime or for building it as a
+/// `cdylib` plugin. Native linkers don't accept the `:`, `/`, `#`, `[` and
+/// `]` characters that canonical ABI symbol names use, so on native targets
+/// symbols are hex-encoded with the scheme in
+/// `wit_bindgen_core::symbol_name` (the same one the C++ generator uses).
+///
+/// Imports are not resolved by the native linker. Each import calls through
+/// a function pointer that starts out null, and a host provides an
+/// implementation at load time by calling the generated
+/// `__wit_bindgen_register_<world><import>` function with a function pointer
+/// of the import's core signature (`<import>` here is
+/// `make_external_symbol(module, name, GuestImport)`). This means everything
+/// links whether or not a host is present: a host only needs to register the
+/// imports it actually implements, and calling an import that was never
+/// registered aborts with a message naming the import and its registration
+/// function.
+///
+/// Exports, including post-return functions, async callbacks, and resource
+/// destructors, are exported under their hex-encoded core export names. A
+/// `__wit_bindgen_cabi_realloc_<world>` function is also exported so hosts
+/// can allocate guest-owned memory when lowering arguments, as the canonical
+/// ABI requires.
+///
+/// The `<world>` prefix above is a hex-encoded
+/// `<package>/<world><type_section_suffix>`, which keeps two `generate!`
+/// invocations in one binary from defining the same symbols. Note that
+/// binding the same world twice in one native binary will fail to link with
+/// duplicate symbols unless `type_section_suffix` is used to tell the two
+/// apart.
 ///
 /// [WIT package]: https://component-model.bytecodealliance.org/design/packages.html
 #[cfg(feature = "macros")]
--- a/crates/guest-rust/src/lib.rs
+++ b/crates/guest-rust/src/lib.rs
@@ -903,29 +903,39 @@ extern crate std;
 /// symbols are hex-encoded with the scheme in
 /// `wit_bindgen_core::symbol_name` (the same one the C++ generator uses).
 ///
-/// Imports are not resolved by the native linker. Each import calls through
-/// a function pointer that starts out null, and a host provides an
-/// implementation at load time by calling the generated
-/// `__wit_bindgen_register_<world><import>` function with a function pointer
-/// of the import's core signature (`<import>` here is
-/// `make_external_symbol(module, name, GuestImport)`). This means everything
-/// links whether or not a host is present: a host only needs to register the
-/// imports it actually implements, and calling an import that was never
-/// registered aborts with a message naming the import and its registration
-/// function.
+/// Imports are not resolved by the native linker. Each import instead calls
+/// through a function pointer looked up on first use from a host-installed
+/// resolver. After loading the library a host calls the exported
+///
+/// ```c
+/// void __wit_bindgen_set_import_resolver(
+///     void *(*resolver)(void *ctx, const char *module, const char *name),
+///     void *ctx);
+/// ```
+///
+/// with a callback mapping an import's core module and function name (e.g.
+/// `my:pkg/iface@1.0.0` and `[method]res.frob`) to a function pointer with
+/// the import's core signature, or null if the host doesn't implement it
+/// (see `wit_bindgen::rt::ImportResolver`). Everything links whether or not
+/// a host is present, and calling an import with no implementation aborts
+/// with a message naming it. Install the resolver once, before calling any
+/// export: each import caches the pointer it was given.
 ///
 /// Exports, including post-return functions, async callbacks, and resource
-/// destructors, are exported under their hex-encoded core export names. A
-/// `__wit_bindgen_cabi_realloc_<world>` function is also exported so hosts
-/// can allocate guest-owned memory when lowering arguments, as the canonical
-/// ABI requires.
+/// destructors, are exported under their hex-encoded core export names. The
+/// runtime crate also exports `__wit_bindgen_cabi_realloc` so hosts can
+/// allocate guest-owned memory when lowering arguments.
 ///
-/// The `<world>` prefix above is a hex-encoded
-/// `<package>/<world><type_section_suffix>`, which keeps two `generate!`
-/// invocations in one binary from defining the same symbols. Note that
-/// binding the same world twice in one native binary will fail to link with
-/// duplicate symbols unless `type_section_suffix` is used to tell the two
-/// apart.
+/// Each world additionally exports a marker
+/// `const char *__wit_bindgen_world_<world>(void)`, where `<world>` is the
+/// hex-encoded `<package>/<world><type_section_suffix>` (e.g.
+/// `my:pkg@1.0.0/my-world`), which returns that name. `dlsym` can't otherwise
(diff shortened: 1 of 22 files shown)


alexcrichton:
Could you address the [question I had in the other PR](https://github.com/bytecodealliance/wit-bindgen/pull/1565#issuecomment-4069639180)? This seems like a lot to take on in terms of maintainership here because only codegen is tested, not any runtime parts, and since that comment the component model has continued to add features like async which I'm not sure how would map to native counterparts.

BjornTheProgrammer:
Sure! I'll respond to the previous comment.

> Thanks for the PR, and I'm trying to page this all back in and understand the motivations/intentions here. Correct me if I'm wrong, but the high level idea is to be able to satisfy, on non-wasm platforms, wit-bindgen-generated bindings with a native implementation. Is that correct?

Yes. This has been demonstrated in the Pumpkin PR I shared. It allows for native and wasm targets to be compiled interchangeably with no code changes from the plugin developer.

> If so, I'm second-guessing this a bit in the sense of why go through ABI-marshalling at all? For example if you're calling a WIT function it seems like it'd be best to just turn around and immediately call a native function rather than serializing to some ABI-based representation just to deserialize and call a host function. Does that make sense to fit in here at all? Or am I perhaps missing something? (I could very well be)

The reason a direct native call doesn't work here is that the guest is compiled separately from the host and loaded with dlopen. Rust has no stable ABI, so passing native types like String or Vec across that boundary, potentially across different rustc versions, or from a non-Rust host won't work. The canonical ABI is already implemented by wit-bindgen and tapping into that is easy. We wouldn't really have to "serialize/deserialize." Nothing is encoded to a buffer. Scalars pass as-is and strings cross as pointer+length, so simple calls are effectively direct calls already.

Reusing the existing ABI also keeps the change small. The existing codegen already emits the full ABI representation on native targets (the unreachable!() stubs sit after all of that), so this PR just swaps the stub for a dispatch through a registered function pointer and hex-encodes the symbol names. Only around +130 lines of actual generator code. Most of that is the register hooks, which exist so the host doesn't have to re-export every import symbol into the dlopen'd library (the -rdynamic problem) and can implement only the subset of imports it needs. And because it's the same underlying C ABI emission, async, resources, and every other WIT construct should work without a parallel code path. Which is why I didn't really see the need to test the runtime part, since whatever ABI works on wasm should just work with native.

A direct-call mode (traits with idiomatic signatures) is imaginable, but I don't think it fits wit-bindgen's shape. There's no language-neutral contract for it to target, so it would be a separate parallel bindings mode per generator, none of which could interoperate across the dlopen or language boundary. The canonical ABI is the one representation every generator here already shares. Which is what lets, e.g., a C++ host load a Rust plugin under this scheme. Theoretically this could be added to the other generators as well.

tschneidereit:
@cpetig I know you've done similar work, so it'd be great if you could weigh in here as well.

cpetig:
Wow, good to see growing interest in applying the component model to native binaries.

I have a long living fork of wit-bindgen at https://github.com/cpetig/wit-bindgen/ which adds
 - hex encoding of the colorful wasm-link names (as native symbols are more constrained on ":/[]" characters) - should become standardized in something akin to https://github.com/WebAssembly/component-model/pull/378
 - removing the `unreachable!`
 - introducing a symmetric ABI mode which enables directly connecting exports to imports (needs one renamed when importing and exporting the same interface from the same module). 
    - a matching test runner for native with symmetric ABI covering a large fraction of the runtime tests
    - this also covers streams and futures with a matching runtime library for POSIX.
    - a dll import linking stub generator which removes the need to link component dlls in dependency graph order
 - some more unrelated niceties (including a now unmaintained mode to connect native canonical components by a "host" mesh, wamr host code)

The symmetric ABI is what makes the real difference in usability as the (also mental) overhead for (maintaining) a host mode (mesh) code generator is significant.

I guess we should have a discussion about standardizing native names in a `BuildTarget.md`, potentially about symmetric ABI and then merge our efforts. I simply dispersed myself in too many parallel activities over the past year to get it approved and merged upstream until now. Honestly the pain was too low to make this high priority.

crates/guest-rust/src/lib.rs
```diff
@@ -891,6 +891,24 @@ extern crate std;
...
+///     // native symbol whose name encodes the characters a linker cannot
+///     // accept. Both targets still build from one source.
+///     //
+///     // The registration symbols are prefixed with a hex-encoded
+///     // `<package>/<world>` so that two `generate!` invocations in one crate
+///     // don't collide. Binding the *same* world twice in one linkage unit
+///     // still does; use `type_section_suffix` to tell them apart. See
+///     // `wit_bindgen_rust::Opts::link_native_symbols` for the full list of
+///     // symbols a host can expect.
+///     link_native_symbols: true,
```
cpetig: (written before I saw that your code is likely the same when the native flag is turned on) 

I think my approach of using conditional compilation to choose between both naming schemes at compilation time of identical generated code is preferable. See e.g. https://github.com/cpetig/wit-bindgen/blob/work-in-progress/crates/cpp/tests/symmetric_future/future/src/future_world.rs#L17-L24 for an example.

crates/rust/src/interface.rs
```diff
@@ -1281,60 +1285,93 @@ unsafe fn call_import(&mut self, _params: Self::ParamsLower, _results: *mut u8)
...
+    /// emitted, as `(cfg, symbol)`: the `cfg` attribute to gate the copy with
+    /// and the symbol to export it as.
+    ///
+    /// Normally there's just one copy: the canonical ABI name with no `cfg`.
+    /// With `link_native_symbols` enabled a second, hex-encoded copy is emitted
+    /// for native targets as well, because native linkers reject the `:`, `/`,
+    /// `#`, `[` and `]` characters that canonical names contain. Names that
+    /// survive encoding unchanged (`$root` exports, for instance) are emitted
+    /// once with no `cfg` rather than twice.
+    fn core_export_symbols(&self, export_name: &str) -> Vec<(&'static str, String)> {
```
cpetig: ok, I see, this is additive, so the generated code is unchanged in the default case. 

I will leave this decision to Alex, but for my branch I decided that it should become the default, simplifying the generation logic and making the flag unnecessary. (It only adds visual clutter to the generated code, no runtime effects)

cpetig (commented):
I feel that the amount of code changes can be reduced by making the native symbol name generation and target-conditional linker attributes the default behavior. This is how I would solve this.

BjornTheProgrammer:
@cpetig thank you for the review!

I hadn't realized you had a fork tackling this issue as well, and I would love to combine our work. We're already aligned on naming (the hex encoding this PR moves into wit-bindgen-core is the same scheme your C++ generator uses).

I've also taken your suggestion. Native symbol name generation and target-conditional linker attributes are now the default behavior, and the link_native_symbols option is gone entirely. Exports are emitted once with paired cfg_attr(target_arch = "wasm32", ...) export names. The one place I deliberately differ from your fork is imports. Instead of undefined externs, the old unreachable!() slot is filled with a dispatch through a __wit_bindgen_register_<world><import>(func: unsafe extern "C" fn(...)) hook. That keeps the default free for existing users (bindings crates still build and link everywhere on host targets. No import stub libraries needed, and calling an unregistered import aborts with a message, same as the stub did), and it's what makes runtime-discovered plugins work: a dlopen'd cdylib can't resolve import symbols against the host executable without -rdynamic on Linux, and can't on Windows at all, which is why I abandoned #1565. The hooks are convention-agnostic. They just carry a core-signature function pointer, so a symmetric mode could register through the same mechanism later.

On the symmetric ABI, I had considered something similar for this PR, but I was worried about whether something requiring maintenance at that scale would be accepted, which is why I settled on the existing wasm32 ABI. It also helped that wasmtime's bindgen existed as a reference for making a fully working native host runtime ([native-wit](https://github.com/BjornTheProgrammer/native-wit)). A lot of our changes are shared regardless. Would it make sense to land this PR as the common substrate (standardized names, no stubs, a portable dispatch mechanism), and then adopt the symmetric ABI as an opt-in mode once the BuildTargets.md discussion stabilizes? Happy to help draft the naming section there. The encoding and hook naming in this PR could serve as input.

crates/rust/src/interface.rs
BjornTheProgrammer: I pushed some changes that now make it the default as well.

BjornTheProgrammer:
Hmm seems like a single unit test fails due to a naming conflict after my changes. I just had an idea how to fix this that might make naming resolution never a problem and also simplify the rest of the code. I'll probably push some changes tomorrow, but for now I would love to hear your input!

crates/guest-rust/src/rt/mod.rs
```diff
@@ -153,7 +153,7 @@ pub fn maybe_link_cabi_realloc() {
 /// `cabi_realloc` module above. It's otherwise never explicitly called.
 ///
 /// For more information about this see `./ci/rebuild-libwit-bindgen-cabi.sh`.
-#[cfg(any(target_env = "p1", target_env = ""))]
```
QuantumSegfault, line 156: You might need to revert https://github.com/bytecodealliance/wit-bindgen/commit/8cde5cec407cda04543a5d3e8cb00995e4ae60f5

Now that you actually need `cabi_realloc` for native platforms as well.

BjornTheProgrammer:
I've made some changes to how I register exports. Additionally I've added support for async, futures, and streams in native. I also made an end-to-end test to validate that nothing breaks between versions using libloading. Please let me know what you think about it.

yunuservices:
Um, is there any progress in this PR? I noticed the last commit was a week ago, I'm just wondering.

BjornTheProgrammer:
> Um, is there any progress in this PR? I noticed the last commit was a week ago, I'm just wondering.

I've made all of the changes that I felt were needed, and the CI just failed a single test because I think it timed out. Should be ready for review.

QuantumSegfault:
@cpetig @alexcrichton 

What's the verdict on this? Still interested in having this?

alexcrichton:
Apologies I've definitely let this languish and haven't been attentive. 

Overall on this I feel like I'm missing something pretty fundamental or not understanding something, and by missing that I'm wary to merge this in and try to maintain it. Having tests is definitely good, and I agree this isn't the biggest change ever, but despite these parts I'm still not entirely sure what to do about this. One thing I'm particulary wary of is what this test is doing internally which I'm assuming is roughly what external consumers would also be doing -- the `resolve`-style function is seemingly quite `unsafe` as there's no type-checks that the ABI requested is the one of the function that's given out. The test here has a handwritten `log` and `add` function for example and it seems like you just have to be super careful to get that correct. I suppose what I sort of expected was everything being 99% safe with some sort of `unsafe` thing in the middle where you'd assert "yes both halves are using the same WIT" or something like that, but this seems to not be shaped towards that. 

Another piece I haven't thought too carefully about is all the memory management/ownership/lifetimes w.r.t the canonical ABI and these functions. For example wasm exports receiving a string inherently expect an owned allocation to be received, but when calling a wasm import that takes a string no allocation is made. This is part of the implicit assumption that something else is what's translating between addresss spaces, but by putting everything in the same process that's breaking a pretty fundamental assumption. At a skim of this PR I'm not seeing where this is handled for example and where the export either no longer assumes it owns the memory or the caller allocates separate memory.

I guess put a bit more broadly everything about WIT and wit-bindgen and such has so far been developed without much heed to a use case like native plugins or native dynamic libraries. I'm wary of breaking this foundational assumption which can have rippling effects far beyond just some tests in this repo or similar. Now there's no really any way to quantify this one way or another, so I'm mostly just trying to explain my rationale of being wary -- I don't think it's possible for me to not be wary of a change like this but that doesn't mean it can't happen.

---

That's not really an answer for a verdict per-se, but I'm also not really equipped to say "definitely yes" or "definitely no" to a change like this. I can try to help out and explain some of the concerns/wariness I have and see how this takes shape over time, but I definitely don't have a picture in my head of "if you did this it'd be landable" for example.

PR #1690, opened 2026-08-19
Make the cli's `csharp` feature actually turn csharp off

### LINKED ISSUE #1510: Csharp dep is marked as optional even when it's a hard dep of test
As the title says csharp is marked as an optional dep in the cli however due to test having a hard requirement for csharp it cannot be disabled.

Lstarsky0:
`wit-bindgen-csharp` is `optional = true` in the cli, but the cli also depends on
`wit-bindgen-test` unconditionally, and that crate depended on `wit-bindgen-csharp`
outright. So the feature never actually removed anything:

```console
$ cargo tree --no-default-features -i wit-bindgen-csharp
wit-bindgen-csharp v0.60.0 (crates/csharp)
├── wit-bindgen-cli v0.60.0 (.)
└── wit-bindgen-test v0.60.0 (crates/test)
```

After this it's out of the graph for `--no-default-features` and for any single
non-csharp language, and still in for the default build and for
`--no-default-features --features csharp`.

`crates/test` is the only test driver that links a bindgen crate — the others shell
out — and it does so for `CSProject`, to write the .csproj. So `mod csharp` and the
`Language::Csharp` variant go behind the same feature. With csharp off the harness
no longer knows the `cs` extension and falls through to the `--custom` lookup, which
says the extension is unknown. Happy to keep the variant always present and error at
run time instead if you'd rather.

Closes #1510

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ wit-bindgen-markdown = { path = 'crates/markdown', version = '0.60.0' }
 wit-bindgen-moonbit = { path = 'crates/moonbit', version = '0.60.0' }
 wit-bindgen-go = { path = 'crates/go', version = '0.60.0' }
 wit-bindgen = { path = 'crates/guest-rust', version = '0.60.0', default-features = false }
-wit-bindgen-test = { path = 'crates/test', version = '0.60.0' }
+wit-bindgen-test = { path = 'crates/test', version = '0.60.0', default-features = false }
 
 [workspace.lints.clippy]
 # The default set of lints in Clippy is viewed as "too noisy" right now so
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! I'm a bit hesitant to do this though because it's a pretty minor gain (removing a crate at compile time), and this otherwise makes the internal implementation of the test framework inconsistent where C# gets a cargo feature but nothing else does.

One thing that I'm not certain of, however, is the original motivations of https://github.com/bytecodealliance/wit-bindgen/issues/1510. Do you have your own motivations for fixing this issue? Or are you mostly looking to contribute and that seemed a good bug to fix?

Lstarsky0:
No motivation of my own here — I'm not a C# user and I don't build with `--no-default-features`. I was looking for issues I could verify end to end without a language toolchain, and this one fit.

I'd also narrow what I wrote in the description: the feature does still gate the `csharp` subcommand, it's only the build graph that's unaffected. So the gain is smaller than I made it sound, and I agree it isn't worth being the one place in the test framework with a cargo feature.

Closing. Leaving #1510 open in case someone shows up who actually wants the smaller build.

PR #1697, opened 2026-08-25
Don't force link `cabi_realloc` on non-Wasm builds.

QuantumSegfault:
Fixes an issue introduced with #1663 where building libraries containing `wit-bindgen` produced bindings for native (non-Wasm) platforms would error out on link due to missing `cabi_realloc`.

This restores the original gate of the force-link behind `target_family = "wasm"`


QuantumSegfault:
@alexcrichton

alexcrichton (approved):

PR #1699, opened 2026-08-26
 rust: store local wrapper addresses in async stream/future vtables 

alexcrichton:
Under a position-independent dynamic library build, an intrinsic that is
only address-taken (stored in the StreamVtable/FutureVtable, never called
by name) is lowered by LLVM to a GOT.func global keyed by its mangled Rust
symbol, and the canonical (module, field) wasm import is dropped. The
component linker resolves GOT.func slots against library exports, but no
library exports that mangled symbol and the canonical name is gone, so
components using dynamic libraries fail to link.

Emit a local wrapper for each of the intrinsics on wasm and store the
wrapper's address in the vtable. The wrapper calls the canonical import
by name (keeping the import live) and the address-take becomes an
ordinary table relocation against a defined function, which the linker
resolves.

### DIFF
--- a/crates/test/src/lib.rs
+++ b/crates/test/src/lib.rs
@@ -1087,7 +1087,11 @@ status: {}",
         bail!("{error}")
     }
 
-    /// TODO
+    /// Converts the list of dynamic libraries in `dylibs` into a component and places it
+    /// in the destination specified by `compile`.
+    ///
+    /// This is similar to `convert_p1_to_component` except usese a
+    /// `wit_component::Linker` instead of a `wit_component::ComponentEncoder`.
     fn link_dylibs_to_component(&self, dylibs: &[PathBuf], compile: &Compile<'_>) -> Result<()> {
         let mut linker = wit_component::Linker::default();
         for dylib in dylibs {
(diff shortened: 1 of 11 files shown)


crates/test/src/rust.rs
```diff
@@ -40,10 +40,19 @@ struct RustConfig {
     /// Space-separated list or array of compiler flags to pass.
     #[serde(default)]
     rustflags: StringList,
+
     /// List of path to rust files to build as external crates and link to the
     /// main crate.
     #[serde(default)]
     externs: Vec<String>,
+
+    #[serde(default)]
```
dicej, line 52: Should probably have a doc comment since the other fields do.

crates/test/src/lib.rs
```diff
@@ -1087,6 +1087,25 @@ status: {}",
         bail!("{error}")
     }
 
+    /// TODO
```
dicej: TODO as in later or TODO as in for this PR?

dicej (approved):

crates/test/src/lib.rs
alexcrichton: Oops, no this is just "TODO alex fix this before making a PR" which I pretty consistently always forget to do...

PR #1702, opened 2026-09-07
feat(d): Improve export marking & naming

QuantumSegfault:
Reworks the marking of `@witExport` to greatly reduce verbosity. `@witInterface` is now a separate UDA from `@witExport`, allowing the same `@witInterface` to easily be bulk applied to symbols. Within a resource, `@witInterface` is not needed and is inferred from the interface of the resource in question.

`@witExport` also now allows omitting the name, instead inferring it by kebab-casing what's assumed to be the camelCase or PascalCase name of the symbol identifier. Furthermore, when specified, `@witExport` also only requires the name of the actual method in question. No need to to include `[method]`/`[static]`, or the name of the resource. e.g. `[method]x.foo` is now specified as just `foo`.

### DIFF
--- a/crates/d/src/lib.rs
+++ b/crates/d/src/lib.rs
@@ -650,9 +650,11 @@ impl WorldGenerator for D {
                     ));
 
                     if emit_exports_stubs {
+                        r#gen
+                            .stub_src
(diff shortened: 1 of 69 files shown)


QuantumSegfault:
@alexcrichton

alexcrichton (approved):

PR #1704, opened 2026-09-08
fix(d): Fix crashes with large component type embeddings

QuantumSegfault:
Currently, when trying to compile generated D bindings against sufficiently large components, there ends up being a stack overflow (or so it seems) in the compiler due to the potentially MASSIVE tree of concatenation expressions used to join the possibly thousands of lines of strings together to form the LLVM IR for the component type section.

Line breaks are needed to keep editors from choking trying to process (e.g. split, syntax highlight, etc.) what would an absurdly long line.

To fix this issue, the raw bytes of the encoded component type are instead embedded as a hex-string (which allows whitespace freely within without it affecting the final contents), and is transformed on the D side at compile time into the string that needs to be embedded into the IR.

### DIFF
--- a/crates/d/src/lib.rs
+++ b/crates/d/src/lib.rs
@@ -904,6 +904,38 @@ impl WorldGenerator for D {
             .unwrap();
 
             world_src.push_str(&format!(
+                "private enum componentTypeBytes = x\"{}\n\";\n",
+                &component_type
+                    .iter()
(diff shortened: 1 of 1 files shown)


QuantumSegfault:
@alexcrichton

alexcrichton (approved):

PR #1710, opened 2026-09-11
`guest-rust`: Implement a `Task` handle for spawned futures

adamrk:
The `Task` can be awaited to get the returned value of the future, or it can be dropped or explicitly canceled to cancel the future. It can also be detached to allow running in the background and match the previous behavior of `spawn_local`.

It's possible that the component-model task into which the future was spawned could be canceled while the `Task` referencing it is still alive. In that case, awaiting or canceling the task will return `None`.

Note: The naming and cancellation behavior here matches [`async_task::Task`](https://docs.rs/async-task/latest/async_task/struct.Task.html) because that's what we were already using in `wstd` and it would allow switching to `wit_bindgen::Task` without requiring a wrapper. But I could also see an argument for going with the naming and behavior of `tokio`, so happy to switch to that if other people think it's better.

### DIFF
--- a/crates/guest-rust/src/rt/async_support/spawn.rs
+++ b/crates/guest-rust/src/rt/async_support/spawn.rs
@@ -7,7 +7,10 @@ use crate::rt::async_support::BoxFuture;
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::future::Future;
+use core::pin::Pin;
 use core::task::{Context, Poll};
+use futures::channel::oneshot;
+use futures::future::{AbortHandle, Abortable, Aborted};
 use futures::stream::{FuturesUnordered, StreamExt};
 
 /// Any newly-deferred work queued by calls to the `spawn` function while
@@ -94,10 +97,10 @@ impl<'a> Tasks<'a> {
 ///   computations executing within a [`block_on`] call, however, the spawned
 ///   tasks will be executed within that scope. This notably means that for
 ///   [`block_on`] spawned tasks will prevent the [`block_on`] function from
-///   returning, even if a value is available to return.
-///
-/// * There is no handle returned to the spawned task meaning that it cannot be
-///   cancelled or monitored.
+///   returning, even if a value is available to return. If `spawn_local` is
+///   called within a component-model async task which is then terminated (e.g.
+///   by the host) before the future resolves, awating the `Task` will return
+///   `None`.
 ///
 /// * The task spawned here is executed *concurrently*, not in *parallel*. This
 ///   means that while one future is being polled no other future can be polled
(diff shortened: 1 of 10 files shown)


crates/guest-rust/src/rt/async_support/spawn.rs
```diff
@@ -108,8 +111,79 @@ impl<'a> Tasks<'a> {
...
+/// allow the future to continue running in the background and [`Task::cancel`] will
+/// explicitly wait for the cancelation to complete.
+///
 /// [`block_on`]: crate::block_on
 /// [#1305]: https://github.com/bytecodealliance/wit-bindgen/issues/1305
-pub fn spawn_local(future: impl Future<Output = ()> + 'static) {
-    unsafe { SPAWNED.push(Box::pin(future)) }
+pub fn spawn_local<T: 'static>(future: impl Future<Output = T> + 'static) -> Task<T> {
+    let (sender, receiver) = oneshot::channel();
+    let (abort, registration) = AbortHandle::new_pair();
```
alexcrichton, line 124: Instead of `AbortHandle` which I believe is a separate allocation from the oneshot, could the oneshot's `close` and `poll_cancelled` methods be used instead?

crates/guest-rust/src/rt/async_support/spawn.rs
```diff
@@ -108,8 +111,79 @@ impl<'a> Tasks<'a> {
...
+    unsafe {
+        SPAWNED.push(Box::pin(async move {
+            let _ = sender.send(Abortable::new(future, registration).await);
+        }));
+    }
+    Task {
+        receiver,
+        abort,
+        cancel_on_drop: true,
```
alexcrichton, line 133: Semantically I think this should preserve the preexisting behavior where if `Task<T>` isn't interacted with then it's not cancelled. That matches the semantics of `std::thread` and `tokio::task` IIRC

crates/guest-rust/src/rt/async_support/spawn.rs
```diff
@@ -108,8 +111,79 @@ impl<'a> Tasks<'a> {
...
+    abort: AbortHandle,
+    cancel_on_drop: bool,
+}
+
+impl<T> Task<T> {
+    /// Cancels the spawned task and waits for cancellation to complete.
+    ///
+    /// This returns the task's output if it completed before it could be
+    /// cancelled, or `None` if it was cancelled or otherwise terminated.
+    pub async fn cancel(mut self) -> Option<T> {
```
alexcrichton, line 155: I might recommend drawing inspiration from Tokio's [`JoinHandle`](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html) for this method signature, notably changing this to `fn cancel(&self)` (or maybe `&mut self`). That way this can be decoupled with the `Future for Task<T>` implementation below as well. (e.g. the protocol is cancel-then-await if users care about the race)

crates/guest-rust/src/rt/async_support/spawn.rs
adamrk: Yep, switched to closing the `Receiver` to indicate cancellation.

crates/guest-rust/src/rt/async_support/spawn.rs
adamrk: Changed it to only cancel through the explicit call.

crates/guest-rust/src/rt/async_support/spawn.rs
adamrk: Makes sense. It's `fn cancel(&mut self)` now because closing the `Sender` requires `&mut`.

crates/guest-rust/src/rt/async_support/spawn.rs
```diff
@@ -108,8 +110,93 @@ impl<'a> Tasks<'a> {
...
+                match fut.poll(cx) {
+                    Poll::Ready(t) => {
+                        let _ = sender.send(t);
+                        Poll::Ready(())
+                    }
+                    Poll::Pending => {
+                        inner.sender = Some(sender);
+                        Poll::Pending
+                    }
+                }
```
alexcrichton: Could the order of these be swapped to check-the-future then check-the-cancel? That way a race of the two has the value production always win (sort of a last-attempt)

alexcrichton (approved):
Looks good to me! One final minor comment, and I think tests are timing out and/or infinite looping in CI

adamrk:
Actually using `close` and `poll_canceled` on the `oneshot` can't give us the detach-on-drop semantics because dropping the receive end automatically closes the channel (current tests are failing/hanging because futures are unexpectedly cancelled). So I think we need the `AbortHandle` or a channel back the other way for cancellation.

alexcrichton:
Oh, right, of course! In that case `AbortHandle` seems fine and can always work on optimizing later if necessary

PR #1711, opened 2026-09-15
feat(d): Make use of `pragma(inline, true)`

QuantumSegfault:
Introduces use of force-inlining across the bindings where valuable to help reduce binary sizes.

When the output of `wit-bindgen` is compiled as a separate library from the uses (e.g. when using `dub` with `  "targetType": "staticLibrary"`, there ends up being a lot of bloat from tiny wrappers/helpers that LDC avoids inlining, even at `-O3`. To help fight this, `pragma(inline, true)` has been added all around where it has helped reduce code size.

In my testing, it takes a minimal component implementing https://github.com/Pumpkin-MC/pumpkin-plugin-wit from nearly 1 MB, down to ~ 575 KiB when compiling with `-Oz`. This is with the `wit-bindgen` output being emitted in one static library, and `exports` in the main compilation. For reference, compiling it all at once *without* this PR yields a size of 625 KiB. So it helps in both cases.


QuantumSegfault:
@alexcrichton

alexcrichton (approved):

QuantumSegfault:
Needs to be requeued. There was a merge conflict.

PR #1716, opened 2026-09-20
fix(d): Make sure the component type is linked in for export-only worlds.

QuantumSegfault:
Adds `__wit_bindgen_component_type_force_link` to the world `Exports` template to make sure exports-only worlds don't get their type stripped.


QuantumSegfault:
@alexcrichton

alexcrichton (approved):

Issue #1703, opened 2026-09-08
Go: Conflicts between identifiers


QuantumSegfault:
While trying to compile Go bindings generated by `wit-bindgen` `0.61.1` for https://github.com/Pumpkin-MC/pumpkin-plugin-wit, I've run into issues with symbol conflicts. I've noticed two major categories:

------

```wit
    variant my-variant {
         tag
    }
```

A variant with member `tag` will generate a getter `Tag` conflicting with the internal `Tag` (discriminator) of the variant itself.

------

```wit
  record my-variant-item {
    // ...
  }
  variant my-variant {
    item(my-variant-item),
  }
```

Having a record `XY` conflicts with the enum member produced for the tag for a variant `X` with case `Y`.

QuantumSegfault:
@asteurer 

You seem to be the maintainer of the Go bindgen?

asteurer:
Thanks for tagging me @QuantumSegfault! Here's what I found: 

### Scenario 1: Enum variant named `tag`
**__Confirmed bug__**. Generated getter method name collides with built-in `Tag` method. Here's what I did to reproduce:

This is the WIT I used: 
```wit
package foo:bar;

world example {
    import test;
}

interface test {
    variant my-variant {
        tag(string),
    }
}
```

These are the resulting bindings:
```go
package foo_bar_test

const (
MyVariantTag uint8 = 0
)

type MyVariant struct {
        tag uint8
        value any
}

func (self MyVariant) Tag() uint8 {
        return self.tag
}

// COMPILE ERROR: method MyVariant.Tag already declared
func (self MyVariant) Tag() string {
        if self.tag != MyVariantTag {
                panic("tag mismatch")
        }
        return self.value.(string)
}

func MakeMyVariantTag(value string) MyVariant {
        return MyVariant{MyVariantTag, value}
}

```

### Scenario 2: Conflicting records and enums

**__Confirmed bug__**. Generated struct conflicts with generated enum variant. See reproduction below for more details. I think this will be a pretty straightforward fix, just a matter of deciding the best way to prevent name collisions between types. I'm going to allocate some cycles this week to think about this and report back with some possible solutions.

Here's the WIT I used:
```wit
package foo:bar;

world example {
    import test;
}

interface test {
    record my-variant-item {}
    variant my-variant {
        item(my-variant-item),
    }
}
```

Resulting bindings:
```go
package foo_bar_test

// COMPILE ERROR: MyVariantItem redeclared in this block
[...]

PR #1717, opened 2026-09-22
Issue 1703

asteurer:
Addresses #1703.

@dicej Should I add extra codegen testcases? I assumed that since this was a set of bugs that would only apply to the Go bindings it wouldn't make sense, but happy to add some if you feel differently.


crates/go/src/lib.rs
```diff
@@ -65,6 +65,26 @@ fn escape_go_keyword(name: String) -> String {
     }
 }
 
+/// The bindings generated for variants require helper methods
+/// that have a risk of colliding with bindings generated for
+/// user-defined WIT types.
+fn matches_reserved_variant_function_name(f: &str) -> bool {
+    match f {
```
dicej: Nit: could be shortened to `matches!(f, "Tag")`

crates/go/src/lib.rs
```diff
@@ -65,6 +65,26 @@ fn escape_go_keyword(name: String) -> String {
...
+        "Tag" => true,
+        _ => false,
+    }
+}
+
+/// The bindings generated for resources require helper methods
+/// that have a risk of colliding with bindings generated for
+/// user-defined WIT types.
+fn matches_reserved_resource_function_name(f: &str) -> bool {
+    match f {
```
dicej: Could be shortened to `matches!(f, "TakeHandle" | "SetHandle" | "Handle" | "Drop" | "OnDrop")`

crates/go/src/lib.rs
```diff
@@ -1951,7 +1971,11 @@ for index := 0; index < int({length}); index++ {{
                         call
                     }
                     FunctionKind::Method(_) | FunctionKind::AsyncMethod(_) => {
-                        let target = &operands[0];
+                        let target = operands[0].clone();
```
dicej: Why was this `.clone()` added?

crates/go/src/lib.rs
```diff
@@ -3002,15 +3026,39 @@ const (
                 case.ty.map(|ty| {
                     let case = case.name.to_upper_camel_case();
                     let ty = self.type_name(self.resolve, ty);
-                    format!(
-                        r#"func (self {name}) {case}() {ty} {{
-        if self.tag != {name}{case} {{
+                    if matches_reserved_variant_function_name(&case) {
+                        eprintln!(
```
dicej: Not necessary for this PR, but we might want to add an option to disable warnings, e.g. a `-q` option.

crates/go/src/lib.rs
```diff
@@ -3002,15 +3026,39 @@ const (
...
                 panic("tag mismatch")
         }}
         return self.value.({ty})
 }}
 "#
-                    )
+                        )
+                    } else {
+                        format!(
+                            r#"func (self {name}) {case}() {ty} {{
```
dicej: Could we reduce the duplication between the `if` and `else` cases here, e.g. putting the comment into a variable that's non-empty in the `if` case and empty in the `else` case, followed by a single `format!` call?

dicej (commented):
Thanks, @asteurer!  Looks good overall; just a few suggestions inline.

Also, it would be good to add a simple codegen test which includes a variant case named `tag`, a resource method named `take-handle`, etc. to make sure the name conflicts are avoided.

crates/go/src/lib.rs
```diff
@@ -22,6 +23,18 @@ use wit_bindgen_core::{
     AsyncFilterSet, Direction, Files, InterfaceGenerator as _, Ns, WorldGenerator, uwriteln,
 };
 
+thread_local! {
+    static WARNINGS: RefCell<BTreeSet<String>> = const { RefCell::new(BTreeSet::new()) };
```
dicej: I tend to consider thread locals (and more generally, global variables) a tool of last resort, e.g. when dealing with a third-party API that inhibits passing values through as parameters for some reason.  Could we instead make this a field on the `Go` struct and make `warn` an associated function which takes a `&mut self`?  That will presumably require making `func_declaration` an associated function as well, since it needs to call `warn`.

asteurer: Ah, that's a much better way to handle this. I was way overthinking that haha

asteurer:
Appreciate your feedback and patience with me @dicej!

dicej (approved):
Looks great, thanks!

